PR #94, opened 2023-02-08
Strip comments touching Aurora copyright line

chiphogg:
Now that we're using the full Apache copyright header, the single-file
versions have exploded in size.  This PR fixes that by stripping all
comments that are contiguous with the Aurora copyright header line.
Since we never have non-copyright comments contiguous with the copyright
(and in fact it would be wrong to do so), this is a safe solution.

Test plan
---------

I ran this command before and after this change:

```sh
make-single-file --units meters seconds | gview -
```

I verified that the only changes were to the version identifier
(specifically, appending `-dirty`), and that all copyright headers
_except the first_ were removed.

### DIFF
--- a/tools/bin/make-single-file
+++ b/tools/bin/make-single-file
@@ -62,9 +62,7 @@ def filenames(main_files, units, include_io):
     The script will be sure to include all of these, and will also include any
     transitive dependencies from within the project.
     """
-    names = (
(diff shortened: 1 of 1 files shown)


geoffviola (approved):
Tests work. Code makes sense. The output file looks cleaner. Thanks!

PR #95, opened 2023-02-08
Add stub prefix reference, and new SI prefixes

chiphogg:
This should be enough to get people started.

I was writing the doc: "We support every SI prefix..." --- wait, no we
don't!  Four new ones got added in 2022.  So I added support for these
newcomers (including tests) in order to make the docs true.

### DIFF
--- a/au/prefix.hh
+++ b/au/prefix.hh
@@ -52,6 +52,22 @@ struct PrefixApplier {
 // SI Prefixes.
 
 template <typename U>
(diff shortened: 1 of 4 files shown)


geoffviola (approved):
The new units match this [doc](https://www.bipm.org/documents/20126/77765681/Resolutions-2022.pdf/281f3160-fc56-3e63-dbf7-77b76500990f#page=8). The tests look good. The doc renders properly, has useful information, and seems like a useful place to start. Thanks!

PR #96, opened 2023-02-08
Add stub refs for Quantity, QuantityPoint

chiphogg:
This should be an acceptable placeholder.

For `Quantity`, we link to the tutorials, which I think should be pretty
useful.  We also point people to the `QuantityPoint` page specifically
for temperatures.

For `QuantityPoint`, we don't have any tutorials to link to, but the
unit tests should be a pretty good stopgap.


geoffviola (approved):
Docs seem useful even if they are a stub. The renderings look good.

PR #97, opened 2023-02-09
Enable trunk in au repository

EliSchleifer:
WIP - still need to sort out getting bazel includes to map correctly with clang-tidy


chiphogg:
I'm going to close this for inactivity, but we love the work that Trunk is doing, and we're definitely open to integrating at some point in the future!

PR #99, opened 2023-02-09
Cast explicit-Rep inverse function output to Rep

### LINKED ISSUE #98: Explicit-Rep inverse functions don't have correct output Rep
This test fails:

```cpp
TEST(InverseAs, ProducesCorrectRep) {
    EXPECT_THAT(inverse_as<int64_t>(nano(seconds), hertz(50.0)),
                SameTypeAndValue(rep_cast<int64_t>(nano(seconds)(20'000'000))));
}
```

I can get a fix up soon.

chiphogg:
Fixes #98.

Test plan
---------

Added a new unit test, and made sure that it would fail without the
associated code change.

### DIFF
--- a/au/math.hh
+++ b/au/math.hh
@@ -194,8 +194,8 @@ constexpr auto int_pow(Quantity<U, R> q) {
 template <typename TargetRep, typename TargetUnits, typename U, typename R>
 constexpr auto inverse_in(TargetUnits target_units, Quantity<U, R> q) {
     using Rep = std::common_type_t<TargetRep, R>;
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
Makes sense. The `static_cast` changes the result of `inverse_as<int64_t>(nano(seconds), hertz(50.0))` from `au::Quantity<au::Nano<au::Seconds>, double>` to `au::Quantity<au::Nano<au::Seconds>, int64_t>`. Thanks!

Issue #69, opened 2023-01-10
Revisit inverse threshold of 1000


chiphogg:
Right now, we permit taking inverses of integers as long as the value we divide into is at least 1,000.  This works well for toy examples like the following:

```cpp
inverse_as(hertz, milli(seconds)(5)); // Yields hertz(200)
```

However, it gets pretty bad for "medium" and "large" values.

```cpp
inverse_as(hertz, milli(seconds)(73));  // Yields hertz(13), instead of ~13.699 Hz
inverse_as(hertz, milli(seconds)(251)); // Yields hertz( 3), instead of ~ 3.984 Hz (!)
```

In the overflow safety surface, we used the intuition that we want to be very safe for values up to 1,000, because users would tend to use the next prefixed unit for values much bigger than this.  (We've since raised the threshold to over 4,000 in that case.)

Perhaps a similar intuition here would yield a threshold of 1,000,000 (inclusive)?  In that case, none of the above examples would compile (with integer inputs).  However, `micro(seconds)` would work.  If we had these same underlying values for units of microseconds instead of milliseconds, we would obtain:

```cpp
                                         // Result          True answer
inverse_as(hertz, micro(seconds)(5));    // hertz(200'000)  200,000.0 Hz
inverse_as(hertz, micro(seconds)(73));   // hertz( 13,698)   13,698.6 Hz
inverse_as(hertz, micro(seconds)(251));  // hertz(  3,984)    3,984.1 Hz
```

(Of course, with floating point inputs all of the above would work and would give accurate answers --- this issue is purely about tuning the safety surface for calculations that never leave the integer domain.)

PR #101, opened 2023-02-11
Keep QuantityPoint conversion within its Rep

### LINKED ISSUE #100: QuantityPoint::InheritsOverflowSafetySurfaceFromUnderlyingQuantityTypes test broken
This is one of those "uncomment to test" cases, because it is supposed to produce a hard compiler error.  I uncommented it, and it didn't produce the error.

I think there are two contributing causes.

1. We respect the underlying integer promotion rules of C++ (inherited from C).  This means that the subtraction that takes place inside of the `QuantityPoint` conversion gets promoted to `int` when it takes place between two 16-bit types.

2. Our origin happens to be defined with Rep `int`, which means the subtraction with a 16-bit Rep would produce `int` even if there weren't any integer promotion.

Impact
------

I don't see any safety risks here.  It looks like just a small annoyance for people who are trying to use smaller integral Reps, _and_ converting temperatures from one scale to another.

Remedy
------

The minimal remedy is to fix up the test case.  I'll see if I can tackle the underlying behaviour as well.

chiphogg:
If we call `celsius_pt(x).in(kelvins_pt / mag<20>())`, the result should
have the same type as `x`.  Currently, that's not the case for two
reasons, both of which stem from the underlying operation of
_subtracting the origin displacement_.

1. The origin itself typically has `int` Rep.

2. Even if it didn't, C++'s integer promotion rules would take us
   outside the realm of shorter integral types.

Part of the solution is to cast the origin displacement Rep to the
desired Rep type explicitly.  There's an opportunity for a safety check
here, so we add it (including unit tests).

The rest of the solution is simply to `rep_cast` at the end to "undo"
any integer promotion that may have happened.

Along the way, we also changed the operation from _subtraction_ of one
origin displacement, to _addition_ of the inverse displacement.  Now
that we're being more careful about checking whether the origin
displacement fits in the Rep, this would have broken unsigned type
support if we hadn't made this change.  (What I think was happening
before with the unsigned types was probably promotion to signed integers
--- so, no "incorrect" results, but probably an unexpected Rep.)

We also needed to add a trivial `rep_cast` overload for `Zero`.  This
helps it "stand in" for arbitrary quantities in one more way.

Fixes #100.

Test plan
---------

- [x] Added new test cases for `OriginDisplacementFitsIn`.
- [x] Uncommented the test case broken in #100, and verified that it
      fails for the right reason.
- [x] Test that there is no unacceptable compile time penalty.

### DIFF
--- a/au/quantity_point_test.cc
+++ b/au/quantity_point_test.cc
@@ -33,7 +33,7 @@ constexpr QuantityMaker<Kelvins> kelvins{};
 constexpr QuantityPointMaker<Kelvins> kelvins_pt{};
 
 struct Celsius : Kelvins {
-    static constexpr auto origin() { return centi(kelvins)(273'15); }
+    static constexpr auto origin() { return (kelvins / mag<20>())(273'15 / 5); }
 };
 constexpr QuantityMaker<Celsius> celsius_qty{};
 constexpr QuantityPointMaker<Celsius> celsius_pt{};
@@ -270,7 +270,7 @@ TEST(QuantityPoint, CanCompareUnitsWithDifferentOrigins) {
 }
 
 TEST(QuantityPoint, CanSubtractIntegralInputsWithNonintegralOriginDifference) {
-    EXPECT_THAT(celsius_pt(0) - kelvins_pt(273), QuantityEquivalent(centi(kelvins)(15)));
+    EXPECT_EQ(celsius_pt(0) - kelvins_pt(273), centi(kelvins)(15));
 }
 
 TEST(QuantityPoint, InheritsOverflowSafetySurfaceFromUnderlyingQuantityTypes) {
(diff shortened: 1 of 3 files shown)


au/quantity_point_test.cc
```diff
@@ -33,7 +33,7 @@ constexpr QuantityMaker<Kelvins> kelvins{};
 constexpr QuantityPointMaker<Kelvins> kelvins_pt{};
 
 struct Celsius : Kelvins {
-    static constexpr auto origin() { return centi(kelvins)(273'15); }
+    static constexpr auto origin() { return (kelvins / mag<20>())(273'15 / 5); }
```
geoffviola, line 36: Can you comment on the `mag<20>()` and `/ 5` parts?

au/quantity_point_test.cc
```diff
@@ -304,4 +304,34 @@ TEST(QuantityPointMaker, CanScaleByMagnitude) {
     StaticAssertTypeEq<decltype(kelvins_pt / mag<5>()),
                        QuantityPointMaker<decltype(Kelvins{} / mag<5>())>>();
 }
+
+namespace detail {
+TEST(OriginDisplacementFitsIn, CanRetrieveValueInGivenRep) {
+    EXPECT_TRUE((OriginDisplacementFitsIn<uint64_t, Kelvins, Celsius>::value));
```
geoffviola, line 310: Does it make sense to test a case where the origin displacent results in a negative number and doesn't fit in the unsigned value?

au/quantity_point_test.cc
```diff
@@ -270,7 +270,7 @@ TEST(QuantityPoint, CanCompareUnitsWithDifferentOrigins) {
 }
 
 TEST(QuantityPoint, CanSubtractIntegralInputsWithNonintegralOriginDifference) {
-    EXPECT_THAT(celsius_pt(0) - kelvins_pt(273), QuantityEquivalent(centi(kelvins)(15)));
+    EXPECT_EQ(celsius_pt(0) - kelvins_pt(273), centi(kelvins)(15));
```
geoffviola, line 273: Commend: Less casting. Looks good!

geoffviola (approved):
LGTM

In the summary, it says

> If we call celsius_pt(x).in(kelvins), the result should have the same
type as x

I get that the intention on the output rep, but I couldn't get that code to work with `kelvins` or with integers and `kelvins_pt`.

```
celsius_pt(0).in(kelvins)
```
There's a compiler error about no matching member function for call to 'in'.

```
celsius_pt(0).in(kelvins_pt);
```
there's `static_assert Dangerous conversion`.

It seems like it runs the same on main so it doesn't seem too concerning.


Also, I ran this test
```
celsius_pt(static_cast<uint8_t>(0)) - kelvins_pt(static_cast<uint8_t>(0));
```
On main, it compiles. In this PR, there's `static_assert Dangerous conversion`.

chiphogg (commented):
> In the summary, it says
> 
> > If we call celsius_pt(x).in(kelvins), the result should have the same
> > type as x
> 
> I get that the intention on the output rep, but I couldn't get that code to work with `kelvins` or with integers and `kelvins_pt`.
> 
> ```
> celsius_pt(0).in(kelvins)
> ```
> 
> There's a compiler error about no matching member function for call to 'in'.
> 
> ```
> celsius_pt(0).in(kelvins_pt);
> ```
> 
> there's `static_assert Dangerous conversion`.
> 
> It seems like it runs the same on main so it doesn't seem too concerning.

Oops --- I was too hasty in writing that in the PR summary.  Clearly, I never tried compiling it!  There are two errors.

1. I should have written `kelvins_pt`, instead of `kelvins`.
2. I needed to divide by (at least) 20, because otherwise this will of course fail for every integral Rep.

I plan to add a test for this, but we'll need to tweak the overflow surface threshold first, which I'm planning to do as part of fixing #69 anyway.

> Also, I ran this test
> 
> ```
> celsius_pt(static_cast<uint8_t>(0)) - kelvins_pt(static_cast<uint8_t>(0));
> ```
> 
> On main, it compiles. In this PR, there's `static_assert Dangerous conversion`.

Perfect!

PR #102, opened 2023-02-12
Tweak overflow and inverse thresholds

### LINKED ISSUE #69: Revisit inverse threshold of 1000

chiphogg:
The primary motivation is to move to a more defensible policy for
inversions.  Formerly, our policy was "avoid truncating nonzero values
to 0" for any value <= 1000.  The new policy is "keep round-trip
inversion lossless" for any such value.  This amounts to changing the
threshold from 1,000 to 1,000,000.

Unfortunately, if we made this change alone, we'd have poor ergonomics
even for pretty reasonable conversions.  For example:

```cpp
constexpr auto period = inverse_as(micro(seconds), hertz(40));
```

This fails because converting `1` to the (dimensionless) product of the
units multiplies by 1,000,000.  Since we're using `int`, this would
overflow the old threshold of 4,294.  So, these thresholds are
connected.

I've actually been thinking for a while that it would be nice to loosen
that threshold ever so slightly.  It's a nuisance to treat `int32_t` and
`uint32_t` differently, despite a very small difference in their ranges.
So I also make that change in this PR.  Ultimately, I think this will be
a very durable "sweet spot".  It's pleasing that we prioritize values of
up to 1,000 generally, and you can fit 3 factors of 1,000 in an `int`,
with a little room to spare (which we put towards extending the safety
surface).

Fixes #69.

Test plan
---------

- [x] Existing unit tests (subject to appropriate tweaks).
- [x] Tested on Aurora's whole codebase.  No significant errors
      introduced.
- [x] Checked compile time impact: no significant effect.
- [x] Added a test, `QuantityPoint.PreservesRep`, which I had wanted to
      add in #101, but couldn't.

### DIFF
--- a/au/math_test.cc
+++ b/au/math_test.cc
@@ -713,8 +713,8 @@ TEST(CeilIn, SupportsDifferentOutputTypes) {
 }
 
 TEST(InverseAs, HandlesIntegerRepCorrectly) {
-    constexpr auto period = inverse_as(milli(seconds), hertz(40));
-    EXPECT_THAT(period, SameTypeAndValue(milli(seconds)(25)));
+    constexpr auto period = inverse_as(micro(seconds), hertz(40));
+    EXPECT_THAT(period, SameTypeAndValue(micro(seconds)(25'000)));
 }
 
 TEST(InverseAs, SupportsDividendLessThanOneThousandForFloatingPointRepOnly) {
@@ -732,8 +732,8 @@ TEST(InverseAs, SupportsDividendLessThanOneThousandForFloatingPointRepOnly) {
 }
 
 TEST(InverseIn, HasSameValueAsInverseAs) {
-    EXPECT_THAT(inverse_in(milli(seconds), hertz(3)),
-                SameTypeAndValue(inverse_as(milli(seconds), hertz(3)).in(milli(seconds))));
+    EXPECT_THAT(inverse_in(micro(seconds), hertz(3)),
+                SameTypeAndValue(inverse_as(micro(seconds), hertz(3)).in(micro(seconds))));
 
     EXPECT_THAT((inverse_in<double>(seconds, hertz(3))),
                 SameTypeAndValue(inverse_as<double>(seconds, hertz(3)).in(seconds)));
--- a/au/conversion_policy.hh
+++ b/au/conversion_policy.hh
@@ -37,7 +37,7 @@ constexpr bool can_scale_without_overflow(Magnitude<BPs...> m, Rep value) {
 
 namespace detail {
 // Chosen so as to allow populating a `QuantityU32<Hertz>` with an input in MHz.
-constexpr auto OVERFLOW_THRESHOLD = 4'294;
+constexpr auto OVERFLOW_THRESHOLD = 2'147;
 
 // This wrapper for `can_scale_without_overflow<...>(..., OVERFLOW_THRESHOLD)` can prevent an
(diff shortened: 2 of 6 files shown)


au/math_test.cc
```diff
@@ -713,8 +713,8 @@ TEST(CeilIn, SupportsDifferentOutputTypes) {
 }
 
 TEST(InverseAs, HandlesIntegerRepCorrectly) {
-    constexpr auto period = inverse_as(milli(seconds), hertz(40));
```
geoffviola, line 716: Comment: I guess we are using the integer division operator so there may be data loss for cases like `hertz(41)`. With `hertz(40)`, 1/40 is a multiple of 0.005, but that can only be determined at compile time.

au/math_test.cc
```diff
@@ -713,8 +713,8 @@ TEST(CeilIn, SupportsDifferentOutputTypes) {
...
-    EXPECT_THAT(period, SameTypeAndValue(milli(seconds)(25)));
+    constexpr auto period = inverse_as(micro(seconds), hertz(40));
```
geoffviola, line 716: Comment: This PR mentions round trip data loss, but operations like `inverse_as(hertz, micro(seconds(25'000)))` don't compile.

geoffviola (approved):
Round-trip lossless inversion sounds like a good direction. The tweaks to align `int32_t` to work better with micro seems reasonable. Thanks!

au/math_test.cc
chiphogg: Misplaced `)`.  This compiles:

```cpp
inverse_as(hertz, micro(seconds)(25'000))
```

It hadn't occurred to me before that people might try applying a prefix to a quantity, not a unit, but I can see how this syntax error would be an easy one to make.  :thinking:

au/math_test.cc
chiphogg: `hertz(41)` is technically still lossless in the sense that the round trip is guaranteed to recover the initial value for everything up to 1,000.  (It's true that the integer result itself won't be exact, of course.)

PR #103, opened 2023-02-12
Mark tutorial executables as `manual`

chiphogg:
When we run `bazel test //...:all`, we currently get this nuisance
warning:

```
tutorial/101_quantity_makers.cc:68:20: warning: unused variable 'track_length' [-Wunused-variable]
    constexpr auto track_length = meters(track_length_m);
                   ^
```

Tutorials should only ever be run manually.  Tagging them as such
prevents this distraction.

### DIFF
--- a/tutorial/BUILD
+++ b/tutorial/BUILD
@@ -23,6 +23,7 @@ cc_binary(
     name = "101_quantity_makers",
     testonly = True,
     srcs = ["101_quantity_makers.cc"],
+    tags = ["manual"],
     deps = [
         ":utils",
         "//au",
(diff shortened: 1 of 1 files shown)


tutorial/BUILD
```diff
@@ -23,6 +23,7 @@ cc_binary(
     name = "101_quantity_makers",
```
geoffviola, line 23: Q: Any reason this is a cc_binary and not a `cc_library` + `cc_test`?

geoffviola (approved):
I confirmed that the warning exists with `bazel test //...:all` on main. The change removes the warning from `tutorial/101_quantity_makers.cc:68`. The change makes sense, since the tutorials should probably be more of a free-form, literate programming style. Thanks!

tutorial/BUILD
chiphogg: I had to dig a bit to jog my memory.  I think I did it this way because of the `print_raw_number_and_quantity()` function, which simply gets called and isn't part of a test.  Not a super compelling reason, but we can revisit it later.

PR #104, opened 2023-02-14
Ensure that `Rep` is `std::is_arithmetic`

### LINKED ISSUE #38: Guard Rep behind trait with static_assert
What if a user accidentally tries to form a Quantity whose Rep isn't an arithmetic type?  Long-term, we'd like to support this by clearly defining the axioms a Rep has to fulfill.  But there are some Rep types we'll never support (e.g., `std::chrono::nanoseconds` intrinsically makes no sense).  It would be better to fail fast.

The fix here would be to add a static_assert on the Rep.  We could check for `is_arithmetic`.  We could also consider defining a trait which _defaults_ to `is_arithmetic`, although maybe it's best to save that for when we are ready to support this for real.

chiphogg:
This applies to both `Quantity` and `QuantityPoint`.  Most users tend
not to hit this use case, but when they do, it can be pretty confusing.
This `static_assert` will give a direct, readable error message.

It turns out we really only need/want to add this to `Quantity`, not
`QuantityPoint`.  If we add it to both, we get a double error message for
`QuantityPoint`.

Fixes #38.

Test plan
---------

I tried some silly test cases, like `meters(meters(3))`, and
`meters_pt(meters_pt(3))`.  They fail to compile both with and without
this PR in both cases, but the error message with this PR is _much_
shorter and more direct.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -113,6 +113,9 @@ class Quantity {
     using Unit = UnitT;
     static constexpr auto unit = Unit{};
 
+    static_assert(std::is_arithmetic<Rep>::value,
+                  "Rep must be built-in numeric type for now; see #52");
+
     // IMPLICIT constructor for another Quantity of the same Dimension.
     template <typename OtherUnit,
               typename OtherRep,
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
I tried it out with the given examples. Here's a comparison in numbers of lines
Quantity 239->10
QuantityPoint 46->16

That's a great improvement. Thanks!

PR #109, opened 2023-02-18
Upgrade bazel to 6.0.0

chiphogg:
This will give us the `add_prefix` parameter for our `http_archive`
repository rules.  We need this to stay clean once we start bringing in
other libraries to use them in our unit tests.

For example, we're about to provide a compatibility layer for the
nholthaus/units repo.  Without `add_prefix`, our include would look like
either `"include/units.h"` or `"units.h"`, which is generic and error
prone.  With `add_prefix = "nholthaus_units"`, we can make this file
`"nholthaus_units/units.h"`.

I found one breakage from this change.  `bazel` now zealously complains
when we run `clang-format`:

```
WARNING: /home/chogg/.cache/bazel/_bazel_chogg/f4320e9077711029fb47fb5166db48f7/external/llvm_14_toolchain_llvm/BUILD.bazel:18:14: @llvm_14_toolchain_llvm//:bin/clang-format is a source file, nothing will be built for it. If you want to build a target that consumes this file, try --compile_one_dependency
```

The suggested remedy, `--compile_one_dependency`, is a red herring.
Nothing depends on this target, so this won't work.  The solution is
twofold.

1. Distinguish between tools we need to `build`, and those we need to
   simply `fetch`.  This forces an extra argument on us for
   `make_command_from_bazel_run` so each tool can tell us which kind of
   tool it is.

2. Run the executable directly, instead of using `bazel run`.  We need
   to do a `bazel query` to translate it from "label" form to "filename"
   form.  We also need some post-processing, since the result of
   `location` output is of the form `<filename>:1:1 <more stuff>`.  This
   is what the `%%:*` is for: it removes everything after the first `:`.

You may be asking yourself whether we could simply do away with all of
this `lib/command_from_bazel` overhead for `clang-format`, since we now
realize it's just a simple file.  The answer, I think, is no: we need to
ensure fetching happens, and this could still take a while if the llvm
14 toolchain hasn't been fetched.

Test plan
---------

- [x] `bazel test //...:all`
- [x] `clang-format --version`
- [x] `buildifier --version`

The last of these still lists the versions as "redacted", but that's a
pre-existing issue.


tools/lib/command_from_bazel.sh
```diff
@@ -30,15 +30,24 @@ function wrap_bazel () {
...
   # Run _real_ command.
-  wrap_bazel run "$TARGET" "$@"
+  if [ "$BUILD_CMD" == "fetch" ]; then
+    # A target that is simply fetched is a pre-built target.  Bazel 6.0.0 will complain if we use
+    # `bazel run`, because `bazel run` uses `bazel build` under the hood, and there is nothing to
+    # build.  So in this case, we simply execute the target directly.
+    QUERY_RESULT="$(bazel --nohome_rc query --noshow_progress $TARGET --output=location)"
+    "${QUERY_RESULT%%:*}" "$@"
+  else
+    wrap_bazel run "$TARGET" "$@"
```
geoffviola, line 51: Nit: `bazel build` and then `bazel run` causes the build graph to be reloaded maybe unnecessarily.

geoffviola (approved):
The bazel upgrade looks good and the reasoning behind it makes sense. I wasn't able to reproduce the warning. Also, I'm not sure about the bash workaround for the bazel warning. Maybe, @philsc can provide some guidance.

tools/lib/command_from_bazel.sh
chiphogg: This is true.  And bazel's overhead isn't negligible.  But I think "have a slight but noticeable delay on every invocation" is a price worth paying to avoid "have an extremely long delay, with no explanation, on occasional invocations".  Maybe someday we'll find something better!

PR #111, opened 2023-02-20
Revert "Ensure that `Rep` is `std::is_arithmetic` (#104)"

### LINKED ISSUE #38: Guard Rep behind trait with static_assert

chiphogg:
This reverts commit db8819ca28ad138c8a0d841cde9d460bc117f142, which
breaks some code I hadn't been aware of in Aurora's repo.  It turns out
that a few people have started using non-arithmetic Rep already, and it
seems to be working out alright.

I also updated the documentation to reflect our "good" status here,
which I elaborated.  I think mp-units still gets "best" here, since they
have a well-defined concept for their Rep.

Fixes #38.


geoffviola (approved):
Too bad for now. We can always revisit this later.

PR #112, opened 2023-02-20
Add compatibility utility for nholthaus units

chiphogg:
This creates a file, `"compatibility/nholthaus_units.hh"`, which makes
it easy to set up a correspondence between each Au Quantity type, and
the corresponding nholthaus units type.  This "correspondence" includes
bidirectional implicit conversions: so, you can pass
`::au::QuantityD<Meters>` to an API expecting
`::units::length::meter_t`, and vice versa.

This correspondence is purely opt-in.  The file also needs to be
included _after_ both of the other two libraries are included.  Since
this is fragile in general, we give instructions and an example for how
to make it _not_ fragile, in `nholthaus_units_example_usage.hh`.

We include a variety of unit tests to show that the quantities are
indeed equivalent between the libraries.  (Actually, this isn't quite
always the case.  These tests uncovered a bug in the nholthaus library,
nholthaus/units#289.)

We reserve the documentation of the corresponding-quantity feature in
general, as well as its application to the nholthaus library, for a
future PR.

Test plan
---------

- [x] Add new unit tests for this file.
- [x] Add example file that shows how to use it.
- [x] Test that the file can replace Aurora's existing implementation.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -116,6 +116,15 @@ http_archive(
     url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.14.0.tar.gz",
 )
 
+http_archive(
+    name = "nholthaus_units",
+    add_prefix = "nholthaus_units",
+    build_file = "@//third_party/nholthaus_units:BUILD",
+    sha256 = "b1f3c1dd11afa2710a179563845ce79f13ebf0c8c090d6aa68465b18bd8bd5fc",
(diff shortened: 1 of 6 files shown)


geoffviola (approved):
I only did a quick review. The usage header gives some notes on how to `#include` it and the tests provide sample conversions. Seems useful for migration. Thanks!

PR #114, opened 2023-02-27
Compile with  `-Wextra`

philsc:
This patch adds `-Wextra` to the compiler options for the three
compilers we currently support. It's a bit uglier than I'd hoped
because we can't specify extra options. We can only override the list
of options. So we have to enumerate the full list manually in clang's
case. For GCC, we have access to the default list.
    
I validated manually with `bazel aquery` that the only difference
before and after this patch is the added `-Wextra` flag.


build/copts.bzl
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+# Extra flags we want to pass to the compilers.
+# -Wall is already set by aspect_gcc_toolchain.
+EXTRA_COPTS = [
+    "-Wextra",
+]
+
+# Since the clang toolchain we're using doesn't let us extract the default flags, we have to
+# manually specify the default flags here. These are copied from:
```
chiphogg, line 22: It's too bad we need to copy, but I understand why and I think it's the best solution.  I really appreciate how you've cleanly separated the defaults from the extras we add on.  That will make it easy to check later on if the defaults have diverged.

chiphogg (approved):

PR #115, opened 2023-02-27
Treat warnings as errors in CI

chiphogg:
This will keep Au compiling warning-free in perpetuity.

We do _not_ set this for _local_ builds because it unnecessarily hinders
experimentation (for example, it adds a bunch of hoops to jump through
for Test Driven Development).

Test plan
---------

I'm posting this after #114 but before #113, because this should turn the silent
warnings in the builds into hard errors on the GitHub checks.  After I confirm
this, I'll land #113 and then confirm that it fixes the checks on this PR.


philsc (approved):

nkiraly (approved):
:shipit:

PR #125, opened 2023-03-19
Replace copy initialization with direct initialization

chiphogg:
This is for the default member initializer of the `value_` member in the
`Quantity` template.

For reasons which are unclear to me, and which I cannot reproduce in a
standalone unit test, #120 has broken a couple of callsites for Quantity
instances with non-arithmetic Rep.  Using direct initialization fixes
the builds.

What I know about the errors is:

- They happen at _template class instantiation_ time.
- They are of the form, `could not convert '0' from 'int' to 'T'`, where
  `T` is the non-arithmetic Rep.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -380,7 +380,7 @@ class Quantity {
  private:
     constexpr Quantity(Rep value) : value_{value} {}
 
-    Rep value_ = 0;
+    Rep value_{0};
 };
 
 // Force integer division beteween two integer Quantities, in a callsite-obvious way.


chiphogg:
> Should the single value constructors like
> 
> ```
> constexpr Quantity(Zero) : value_{0} {}
> ```
> 
> be marked explicit?

I gave them a quick once-over.  I don't think we want to make them explicit --- keeping these implicit is the desired behaviour.

> Do we use cpplint.py?

Interestingly, one has a `NOLINT` comment, but the others don't.  Evidently we're not doing any kind of C++ linting yet.  We could do cpplint.py, but maybe it would be better to add trunk support and up our linting game more broadly.

Issue #131, opened 2023-05-20
Support versioned documentation


chiphogg:
Inspired by mpusz/units#435.

I have done some initial exploration.  I've understood how [mike](https://github.com/jimporter/mike) works, and I've got a proof of concept showing that I can get it running in our repo.

Here's what I think we'd need to accomplish to call this done.

- [x] The documentation has a version selector combo box at the top.
- [x] The release process has a new step to manually generate the documentation at that version.
- [x] `main` is the default label.
- [x] Every existing page redirects to the corresponding version on `main`.
- [x] Every commit to `main` automatically updates the docs stored there.
- [x] `au-docs-serve` still works, and shows the latest version of the content that would go to `main`.

So, this would change our URLs in the following way:

- [`https://aurora-opensource.github.io/au/develop/`](https://aurora-opensource.github.io/au/develop/) (old)
- [`https://aurora-opensource.github.io/au/main/develop/`](https://aurora-opensource.github.io/au/main/develop/) (new)
- [`https://aurora-opensource.github.io/au/0.3.2/develop/`](https://aurora-opensource.github.io/au/0.3.2/develop/) (new; stable content)

We could also consider stretch goals around warnings for old versions:

- [ ] Add [warning](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#version-warning) for concrete versions other than the latest.

chiphogg:
I decided not to include the version warning.  I played around with it for a bit, but couldn't seem get it to do exactly what I wanted.

This issue is now resolved!

PR #132, opened 2023-05-20
Add `//:mike` command for versioned docs

chiphogg:
`mike` (https://github.com/jimporter/mike) is a tool for managing versioned documentation when using both mkdocs and GitHub pages (as we do).  The idea is that you render a given version of the docs once, store it in your `gh-pages` branch, and never touch it again.  `mike` will store it in a subfolder with that version's name.  You can also set up a combo box on each page, so users can switch to see the current page at any version of the docs.

This PR is just the first step for #131.  We'll add a new target, `//:mike`, which exposes the tool: to run,

```sh
mike ...
```

you would instead run,

```sh
bazel run //:mike -- ...
```

To prove that it works, we'll cherry-pick this commit onto `0.3.1` and `0.3.2` and generate versions of the documentation website.  (We won't do `0.3.0` or earlier, because those versions predate public availability, and they had pretty embarrasing placeholder content.) We'll publish these doc site versions to `gh-pages`, so they'll be publicly available, but nothing will link to them yet.  We'll also publish a version for `main`.

We also update the instructions for cutting releases to take versioning into account.

Future steps to resolve the "core" of #131:

1. Make another PR so the GitHub action that updates the doc website on every commit to `main` uses `mike`.

2. Set `main` as the "default" in `mike`, which should set up a redirect for the root website.  (This won't be a PR; we'll commit directly to `gh-pages`.)

3. Replace every existing HTML page on the current website with a redirect to `main`, patterned off of the main page.  (This won't be a PR; we'll commit directly to `gh-pages`.)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -38,6 +38,23 @@ py_binary(
     ],
 )
 
+py_binary(
+    name = "mike",
+    srcs = ["mike_bin.py"],
+    main = "mike_bin.py",
+    data = [
+        "mkdocs.yml",
+        ":au_hh",
+        ":au_noio_hh",
+    ] + glob(["docs/**"]),
+    deps = [
+        ":update_docs",
+        requirement("mike"),
+        requirement("mkdocs"),
+        requirement("mkdocs-material"),
+    ],
+)
+
 BASE_UNITS = [
     "meters",
     "seconds",
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -120,8 +120,17 @@ On the tags page, click the three-dots menu, and select "Create release".
 
 ### Regenerate the doc website
 
-This will update the manifest for `au.hh` and `au_noio.hh`.
+First, create the version of the doc website corresponding to this release.
 
 ```sh
-bazel run //:update_docs -- gh-deploy
+bazel run //:mike -- deploy --push 0.3.1
+```
+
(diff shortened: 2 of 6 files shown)


chiphogg:
To see that this PR works, check out the "versioned" version of the website:

https://aurora-opensource.github.io/au/main/

Note that there's a combo box in the top, which lets users select `main`, `0.3.2`, or `0.3.1`.  Most of these pages aren't really different at all: there are currently very few changes among these versions.  However, the generated single-file packages give a good demonstration.  Check out these links:

- https://aurora-opensource.github.io/au/main/au.hh
- https://aurora-opensource.github.io/au/0.3.2/au.hh
- https://aurora-opensource.github.io/au/0.3.1/au.hh

Look at the manifest comment for each file.  It shows the correct version!  This also means we'll have permanent URLs where it's easy to grab the single-file package for any specific version, which is pretty neat.  :sunglasses: 

(The reason the `0.3.x` releases are called `0.3.x-1` is because I had to cherry-pick this PR's commit in order to generate them.  Later on, I can overwrite the files in a manual commit to the `gh-pages` branch.)

chiphogg:
I tested `au-docs-serve`.  It still works just as before.  There's no combo box to pick among releases.  That's normal and expected.

timhirsh (approved):
Nice addition, LGTM 👍

Issue #116, opened 2023-03-06
Support value conversion with rational exponents


chiphogg:
Au supports units with rational exponents (e.g., square-root meters).  You can get values in and out.  What you _can't_ do is to perform _conversions_ whose factor includes rational exponents.  The reason is simple: we compute our conversion factors at compile time, and we don't have a `constexpr` version of `std::powl`.  (We could perhaps get one from [kthohr/gcem](https://github.com/kthohr/gcem), but we **really** don't want to go from 0 to 1 external dependencies.)

Here's a simple "acceptance test" which gives an example of what this feature would enable.

```cpp
constexpr auto geo_mean = sqrt(inches(1.0) * centi(meters)(1.0)).in(centi(meters));
```

This issue is resolved when this line compiles and gives a correct answer.

It's not anticipated to be a high priority any time soon.  We've basically never needed this in our applications.  I'm filing it to keep track of it for completeness' sake.

PR #138, opened 2023-06-19
Fix traits for num, denom, and integer part

### LINKED ISSUE #83: Rename Magnitude numerator/denominator traits
`numerator(m)` gives the _integer part_ of the numerator.  This is counterintuitive.  It should instead correspond to "the parts that a human would put on top of the fraction".  Similar reasoning applies to `denominator(m)`.

To fill the role played by the current versions of `numerator` and `denominator`, we can either make "integer"-named versions, or a generic `integer_part(m)` function which can compose with them.

chiphogg:
The current definitions of numerator and denominator assume we're
talking about the integer part only.  This policy is error-prone and
unclear from the callsite.  It also holds us back from automatically
generating labels for `Magnitude<...>` types, which would improve basic
library usability significantly at the margins.

Of course, we need the current definitions, which is why we added them
in the first place.  To continue meeting our needs, we add a new trait
operation: `integer_part(m)`.  This is equivalent to the old behaviour
of `numerator(m)`: it pulls out the _integer part_ of the numerator of
`m`.  But it frees up `numerator(m)` for a more natural interpretation.

On the implementation side, I took the opportunity to replace a
recursive implementation with one which is expanded inline, which may be
slightly more efficient.  This new approach exposed a bug in our
parameter pack handling: apparently, the unary product only actually
worked for single-element packs.  I added a test to expose this and made
the test pass.

I also grepped for `numerator`, `NumeratorT`, `denominator`, and
`DenominatorT`, to make sure we updated all of our old uses.  This
grepping caught a use case which we can't currently test for: rational
exponents (#116).  There _is_ a commented-out test case here, in the
`sqrt.CanConvertIfConversionFactorRational` case from `//au:math_test`,
which is commented out because it currently fails to compile.  (Note
that even if we had missed this update, it still wouldn't have compiled
because of a safeguard in `magnitude.hh`.)  In any case, I did find and
update this use case, which will save a little work in case we ever
attempt #116.

Testing plan:
- [x] New unit tests
- [x] Grepped existing uses
- [x] Compile time impact measurements

Fixes #83.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -96,18 +96,26 @@ struct InOrderFor<Magnitude, A, B> : LexicographicTotalOrdering<A, B, detail::Or
 // Type trait based interface for Magnitude.
 
 template <typename MagT>
-struct Numerator : stdx::type_identity<Magnitude<>> {};  // Default, with specializations below.
+struct IntegerPartImpl;
 template <typename MagT>
-using NumeratorT = typename Numerator<MagT>::type;
+using IntegerPartT = typename IntegerPartImpl<MagT>::type;
+
+template <typename MagT>
+struct NumeratorImpl;
+template <typename MagT>
+using NumeratorT = typename NumeratorImpl<MagT>::type;
 
 template <typename MagT>
 using DenominatorT = NumeratorT<MagInverseT<MagT>>;
 
 template <typename MagT>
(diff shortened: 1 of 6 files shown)


geoffviola (approved):
The API looks more explicit. Compile times look unaffected. The test coverage seems better.

PR #141, opened 2023-06-22
Change the label for Fahrenheit from "F" to "degF"

### LINKED ISSUE #60: Change Fahrenheit label from "F" to "degF"
This would resolve a weird discrepancy with celsius, and an ambiguity with the label for farads when we add those.

chiphogg:
Fixes #60.


geoffviola (approved):
Looks clearer.

PR #142, opened 2023-06-23
Prevent integer division with raw numerics

### LINKED ISSUE #39: Tighten up integer division checks on Quantity
This test fails:

```cpp
EXPECT_EQ(hertz(100), 1 / milliseconds(10));
```

It _should_ fail to _compile_.  (i.e., the goal is to expand our integer division check to include the cases where only one argument is a Quantity.)

chiphogg:
This closes a loophole and makes the library more consistent.

Fixes #39.


geoffviola (approved):
Looks good.

I thought mixing quantities and raw values was a little weird. But we know they are `int` and they are being used as ratios. So it should be OK.

PR #146, opened 2023-07-07
Fix compiler errors on MSVC v142

### LINKED ISSUE #143: Windows MSVC C++17 CMake/Ninja Compile Errors on prebuilt header
Using prebuilt https://aurora-opensource.github.io/au/0.3.2/au_noio.hh, I included the file in my project and I get this block of errors. 
```
C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au\au.h(1247): error C2672: 'au::detail::are_consecutive_elements_in_order_for': no matching overloaded function found
  C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au/au.h(1247): note: see reference to class template instantiation 'au::AreElementsInOrder<Pack,Pack<Ts...>>' being compiled
C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au\au.h(1247): error C3207: 'au::detail::are_consecutive_elements_in_order_for': invalid template argument for 'Pack', class template expected
  C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au/au.h(1229): note: see declaration of 'au::detail::are_consecutive_elements_in_order_for'
C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au\au.h(1247): error C2977: 'au::detail::are_consecutive_elements_in_order_for': too many template arguments
  C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au/au.h(1219): note: see declaration of 'au::detail::are_consecutive_elements_in_order_for'
C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au\au.h(1261): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'std::array<bool,sizeof...(Ts)>'
  C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au/au.h(1261): note: The target type has no constructors
  C:\Users\conno\Documents\SC\Platform\Compute_Server\externals\Au/au.h(1261): note: see reference to class template instantiation 'au::AreAllPowersNonzero<Pack,Pack<Ts...>>' being compiled
```

I am unable to figure out what the cause might be. It seems this may be the only problem with Au on this compiler, though, so it's valuable to know a fix or workaround.

### Compile info:
- Windows 10
- RelWithDebInfo Configuration
- Toolchain: msvc_x64_64
- MSVC v142 - VS 2019
- cl version 14.29.30133
- CMake build through Visual Studio's CMake support (Ninja)
- CMake configuration:
```cmake
set(CMAKE_CXX_STANDARD 17) 
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```

chiphogg:
Changes include:

- Replacing function template overloads with struct template
  specializations
- Taking values directly in `all_true()`, instead of forcing end users
  to construct a `std::array`

I believe our previous code was valid standard C++14, but happily, these
changes look like strict improvements in their own right.

The second change enabled us to get rid of `<array>` in `:packs`.
Apparently, this exposed an IWYU violation in `:math`, so we needed to
add `<algorithm>`.  However, this also seems like a net win.

Fixes #143.

### DIFF
--- a/au/packs.hh
+++ b/au/packs.hh
@@ -14,7 +14,6 @@
 
 #pragma once
 
-#include <array>
 #include <cstdint>
 #include <ratio>
 #include <utility>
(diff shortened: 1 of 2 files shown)


au/packs.hh
```diff
@@ -519,39 +518,35 @@ struct IsValidPack : stdx::conjunction<detail::IsPackOf<Pack, T>,
...
-template <std::size_t N>
-constexpr bool all_true(const std::array<bool, N> &values) {
-    for (auto i = 0u; i < values.size(); ++i) {
-        if (!values[i]) {
+namespace detail {
+
+constexpr bool all_true() { return true; }
+
+template <typename... Predicates>
+constexpr bool all_true(Predicates &&...values) {
```
geoffviola, line 536: Comment: I guess we could use `std::all_of`, but that is only `constexpr` in C++20 and we want it in C++14.

geoffviola (approved):

au/packs.hh
chiphogg: Good suggestion, but yeah, I think we'll need to roll our own for that reason.

Issue #144, opened 2023-07-05
Test basic functionality for wide variety of platforms/compilers


chiphogg:
#143 has exposed a tooling deficiency.  So far, we've assumed that simply "write correct C++14" is good enough to support everyone who's using a standard at least as recent as C++14.  But that issue shows a compiler error in a product whose name includes "2019", which suggests this isn't true.  It shows we're currently blind to our _actual_ level of support for various platform/compiler combinations.

This issue has two components for a definition of "done":

- [ ] Document our two "tiers" of platform/compiler support
- [x] Add measurements for a wider variety of platforms/compilers

As to the first, a platform/compiler is "fully supported" whenever we fulfill two conditions:

1. We have automatic gating tests for that platform/compiler.
2. It is easy for almost all end users to run those tests **locally**.

The motivation for the second condition is that working on mpusz/units#300 was quite painful due to the gating MSVC jobs --- it ended up being delayed about 2 or 3 months because I had a bizarre error which I couldn't iterate on locally.  (I think it turned out to be an actual compiler bug!  :sweat_smile:)

The second tier of support would be "best effort", where we have automatic tests to _measure_ support for that platform, but we don't strictly require that those tests pass.  If they fail, we need to make an effort to get them passing, but we won't guarantee that we'll block PRs on keeping them passing.

Currently, our only real support is for platforms that can offer a hermetic bazel toolchain, which includes two versions of clang and one of gcc.  I believe we can make GitHub actions for a wider variety of platforms, including various versions of MSVC.  _If it's possible to run these GitHub actions locally_, then we could significantly broaden the set of platforms/compilers that will be fully supported.  That'll be the goal.  Otherwise, we will still try to add them, but as "best effort".

PR #151, opened 2023-07-24
Add CI jobs for Windows

chiphogg:
The job builds a single-file version of the Au library, and runs a
simple test.  This test is enough to reproduce the errors in #143:
https://github.com/aurora-opensource/au/actions/runs/5496101666/job/14882361539

It also catches another few compiler errors on other versions of MSVC,
which we fix here.  The issue is that function parameters can't be used
in a `constexpr` context in general.  These particular parameters are
monovalue types
(https://aurora-opensource.github.io/au/main/reference/detail/monovalue_types/),
so the compiler's being a little pedantic in enforcing this rule, but
this also means that it's easy to fix the error.

With this PR, we go from 0 to 1 in terms of CI jobs for "best effort"
platforms.  This gives us a pattern we can use for future new platforms.

Helps #144.

### DIFF
--- a/.github/workflows/single-file-build-and-test.yml
+++ b/.github/workflows/single-file-build-and-test.yml
@@ -1,0 +1,28 @@
+# Copyright 2023 Aurora Operations, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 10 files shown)


timhirsh (approved):
LGTM 👍

chiphogg:
Since these new jobs are the first in the "best effort" tier, technically they don't have to always be passing.  However, I think we want to make, well, our best effort to keep them passing, so I'll make them blocking unless and until something forces us to let them fail.

PR #153, opened 2023-07-26
Document adding new base dimensions

### LINKED ISSUE #19: Document adding new base dimensions
This is very rarely needed, but we should explain how to do this to complement the new unit definition how-to doc.

chiphogg:
We explain the context for this operation (that it is an implementation
detail), and then we show how to do it.

To make this operation friendlier for external users, I also made a
couple tweaks to our implementation.  First, I shifted our existing
indices down by 100 and into the negatives, so we can save all of the
positive indices for our users.  Second, I changed the index type from
`int` to `int64_t`, so that users taking the "Unix epoch" approach won't
be vulnerable to the "2038 problem" (admittedly unlikely to be relevant,
but also unlikely to cause a problem).

I added a couple related links: one from the main "how-to" page, and an
anchor for a section header.

To test this change, I added a unit test which follows the instructions.
I also followed every individual link in the rendered docs, and made
sure they went where I was expecting.

Fixes #19.


johnzoppina (approved):
This looks good to me

tobin (approved):

PR #154, opened 2023-07-26
Make separate workflow for each build config

chiphogg:
This will enable us to give each config its own badge.  We can display
these badges prominently on our README file.  We can also add them to
the new page we'll add which clarifies which configurations we support.


chiphogg:
Status badges:

[clang14-ubuntu](https://github.com/aurora-opensource/au/actions/workflows/clang14-ubuntu.yml)
[clang11-ubuntu](https://github.com/aurora-opensource/au/actions/workflows/clang11-ubuntu.yml)
[gcc10-ubuntu](https://github.com/aurora-opensource/au/actions/workflows/gcc10-ubuntu.yml)

[MSVC-x64-19.29.30151](https://github.com/aurora-opensource/au/actions/workflows/msvc-x64-19-29-30151.yml)
[MSVC-x64-19.35.32217.1](https://github.com/aurora-opensource/au/actions/workflows/msvc-x64-19-35-32217-1.yml)

I think these should all go from "grey" to "green" once we land this to main.

Also, my plan is to remove the required build-and-test checks in the repo settings once this lands, and then add the new versions of the jobs.

timhirsh (approved):
👍  I like the badges

> Also, my plan is to remove the required build-and-test checks in the repo settings once this lands, and then add the new versions of the jobs.

Small clarification: the old versions of the required checks are going to block you from merging this.  So you'll need 3 phases: 

- disable current required build checks
- merge
- enable new required builds checks

chiphogg:
Right... that was what I was picturing in my head.  Honest!  :smile:

chiphogg:
Click "View details" above to see the job names that motivated #155:

image

Also, just confirming that the badges did indeed turn green.  :slightly_smiling_face:

PR #155, opened 2023-07-26
Use more descriptive job names

chiphogg:
On a recent PR (#154), basically every job's status was reported as
either "build", or "build / build".  We should make these more
descriptive so it's easier to tell what's what.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -22,7 +22,7 @@ on:
         type: string
 
 jobs:
-  build:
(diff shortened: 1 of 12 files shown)


chiphogg:
Apparently, we also needed this change in order to be able to set the various jobs to "required" in the repo settings.  I've now done so --- including jobs like the linters and ensure-pinned-actions that we couldn't do before!

timhirsh (approved):
Makes sense to me 👍

PR #156, opened 2023-07-28
Clarify tiers of support for compilers

### LINKED ISSUE #144: Test basic functionality for wide variety of platforms/compilers

chiphogg:
We'll provide 3 tiers: full support, best effort, and assumed support.
The new doc explains what these 3 tiers mean.

The motivation behind this division is in part due to my experience
working on mpusz/mp-units#300.  That project got stalled for multiple
months because I couldn't iterate locally on an issue we encountered
with the MSVC compiler (which turned out to be a compiler bug!).
Therefore, the compilers which _do_ let us iterate locally get the
highest level of support.  Next up is "anything we can make a CI job
for", including several Windows/MSVC configurations.  We plan to keep
these passing always, although a time might come when something in this
tier gets dropped because it's too painful to maintain.  The final tier
is "assumed support", which is just what it sounds like.

I also added the new badgest to the README.md.  I thought about adding a
section for compiler support, but it didn't flow well, and besides,
everybody understands more or less what the badges mean --- we can
revisit this if and when any of them ever start failing.

Fixes #144.


docs/index.md
```diff
@@ -15,6 +15,10 @@ These pages will be most useful as you begin your Au journey:
   What else is out there?  This page gives a detailed comparison to some of the most prominent other
   C++ units libraries.
 
+- **[Supported Compilers](./supported-compilers.md).**  Au aims to work with _any_ C++ compiler with full
+  support for C++14 or later.  That said, some platform/compiler combinations have more detailed
```
johnzoppina, line 19: "with...with"

How about,

>  Au aims to work with _any_ C++ compiler that fully supports C++14 or later.

connorjak (approved):
Good docs changes! I agree that the badges are enough in the top readme.

docs/index.md
chiphogg: Done.

tobin (approved):

PR #157, opened 2023-07-28
Add almost all core SI units

chiphogg:
We get our units from the figure on this NIST page:
https://www.nist.gov/pml/owm/metric-si/si-units

Grammatically, we assume that `becquerel`, `lux`, and `tesla` have the
same plural form as the singular.

We avoid dealing with Sievert for now because I'm worried it would be
too error-prone.  A key goal is to avoid mixing it up with Gray.  A
quantity kind solution, such as the one which mp-units has, would be
ideal here.  Alternatively, we could experiment with defining a brand
new dimension for "absorbed dose equivalent", the same way that the SI
does for luminant quantities.  But in any case I think we should defer
this until we have actual use cases, and even then tread cautiously.

By contrast, I'm not especially worried about mixing up Becquerel,
Hertz, and inverse Seconds; I don't think it's likely to cause problems
in practice.

### DIFF
--- a/au/units/becquerel.hh
+++ b/au/units/becquerel.hh
@@ -1,0 +1,35 @@
+// Copyright 2023 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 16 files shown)


geoffviola (approved):
New units and tests seem reasonable.

PR #158, opened 2023-07-30
Clarify "dangerous conversion" error messages

### LINKED ISSUE #127: Improve error messages for dangerous conversion
Consider this code:

```cpp
(meters / second)(1).in(miles / hour);
```

It produces this error message:

```
In file included from au/au_test.cc:15:
In file included from ./au/au.hh:19:
In file included from ./au/math.hh:20:
./au/quantity.hh:159:9: error: static_assert failed due to requirement 'implicit_rep_permitted_from_source_to_target<int>(unit, u)' "Dangerous conversion: use .as<Rep>(NewUnit) instead"
        static_assert(implicit_rep_permitted_from_source_to_target<Rep>(unit, u),
        ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./au/quantity.hh:181:20: note: in instantiation of function template specialization 'au::Quantity<au::UnitProduct<au::Meters, au::Pow<au::Seconds, -1>>, int>::as<au::UnitProduct<au::Miles, au::Pow<au::Hours, -1>>, void>' requested here
            return as(u).in(u);
                   ^
./au/quantity.hh:203:16: note: in instantiation of function template specialization 'au::Quantity<au::UnitProduct<au::Meters, au::Pow<au::Seconds, -1>>, int>::in<au::UnitProduct<au::Miles, au::Pow<au::Hours, -1>>, void>' requested here
        return in(NewUnit{});
               ^
au/au_test.cc:100:26: note: in instantiation of function template specialization 'au::Quantity<au::UnitProduct<au::Meters, au::Pow<au::Seconds, -1>>, int>::in<au::UnitProduct<au::Miles, au::Pow<au::Hours, -1>>>' requested here
    (meters / second)(1).in(miles / hour);
                         ^
1 error generated.
```

This could be better.

1. Dangerous conversions basically always use integral rep, AFAIK.  Users might not notice that they're using integers, and might not realize that Au performs unit conversions in the same type.  It'd be nice if we could add hints about the integral rep being the part that makes the conversion dangerous.
2. We tell users to use `.as<Rep>(...)`, but we don't say instead of what.  In this case, it's because `.as` is an implementation detail for `.in`.  This means it looks to the end user as though we're saying to replace `.in` by `.as`!  We should probably just link directly to the troubleshooting guide, so that we're not unconditionally guiding users to the explicit-Rep version when alternative solutions (such as just using `1.0` in this case) would be better.

Once we update the error messages, we'll need to regenerate the error messages in the troubleshooting guide as well.

chiphogg:
The status quo has two major problems.  First, it gives out-of-context
advice to use the explicit-Rep version `.as<Rep>`, which (as the
troubleshooting guide explains) is usually not the _best_ remedy.
Second, it gives the user no hint that the true root cause is likely the
use of integer Rep, which is the only case where this message can occur.

The remedy has two parts.  First, we make that assumption --- that this
never occurs unless we have an integral Rep --- explicit, testing it
with a new `static_assert` (one that we _never_ expect end users to
see).  This motivated encapsulating the conditions into named variables,
which incidentally improves the compiler error messages.  Second,
instead of telling users to "just use `.as<Rep>()`", we link directly to
the relevant section of the troubleshooting guide, so they can see what
remedies are available and understand their tradeoffs.

This change, of course, makes the troubleshooting guide out-of-date.
Instead of updating only that section, I regenerated all of it by
re-running the compiler on the example commands.  There are several
categories of changes.

1. **Line numbers.**  By far the most numerous change.  Both the Au
   library files, and the `error_examples.cc` file, have changed (the
   latter, via inclusion of the full copyright header).

2. **Dangerous conversion.**  This, of course, stems from the main
   change in this PR.

3. **Integer division.**  We now encapsulate this logic inside a helper,
   `warn_if_integer_division`, which changes the error message.

4. **Dangerous inversion.**  Instead of a raw literal `1000`, we now
   have a variable called `threshold` (and the _value_ of that variable
   has also changed).

Fixes #127.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -158,8 +158,17 @@ class Quantity {
 
     template <typename NewUnit, typename = std::enable_if_t<IsUnit<NewUnit>::value>>
     constexpr auto as(NewUnit u) const {
-        static_assert(implicit_rep_permitted_from_source_to_target<Rep>(unit, NewUnit{}),
-                      "Dangerous conversion: use .as<Rep>(NewUnit) instead");
+        constexpr bool IMPLICIT_OK =
+            implicit_rep_permitted_from_source_to_target<Rep>(unit, NewUnit{});
+        constexpr bool INTEGRAL_REP = std::is_integral<Rep>::value;
+        static_assert(
+            IMPLICIT_OK || INTEGRAL_REP,
+            "Should never occur.  In the following static_assert, we assume that IMPLICIT_OK "
+            "can never fail unless INTEGRAL_REP is true.");
+        static_assert(
+            IMPLICIT_OK,
+            "Dangerous conversion for integer Rep!  See: "
+            "https://aurora-opensource.github.io/au/main/troubleshooting/#dangerous-conversion");
         return as<Rep>(u);
     }
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -79,32 +79,32 @@ It's the "classic" error the units library aims to prevent.
 
     **Compiler error (clang 14)**
     ```
-    au/error_examples.cc:19:17: error: calling a private constructor of class 'au::Quantity<au::Seconds, double>'
+    au/error_examples.cc:33:17: error: calling a private constructor of class 'au::Quantity<au::Seconds, double>'
         set_timeout(0.5);
                     ^
-    ./au/quantity.hh:369:15: note: declared private here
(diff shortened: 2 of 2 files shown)


geoffviola (approved):
The error messages make more sense. Thanks!

PR #159, opened 2023-07-30
Update python locked dependencies: remove e-Tugra

chiphogg:
This fixes dependabot security alert 4:
https://github.com/aurora-opensource/au/security/dependabot/4

As usual, we simply re-run the entire dependency lock script, since it
is good to update our packages routinely:

```sh
bazel run //:requirements.update -- --upgrade
```

To test this change, I ran `au-docs-serve`, and switched back and forth
between the local and production versions of the website, for the first
several pages, including the generated `au.hh` file.  I didn't see any
visible changes.


timhirsh (approved):
:shipit:

PR #161, opened 2023-07-31
Improve dimensionless nholthaus unit support

chiphogg:
Previously, we had a special implementation for `percent` only.  This PR
switches it to a more general solution that will handle every
dimensionless value.  (Note that the existing `percent` tests still
pass, unmodified.)

This change uncovered the fact that some nholthaus units (including
`ppb_t`) are defined in terms of other `unit<...>` specializations,
_not_ directly in terms of a `base_unit<...>`.  I added a new generic
pattern to deal with this case recursively.


geoffviola (approved):
Thanks for improving compatibility!

PR #164, opened 2023-08-05
Add style guide

### LINKED ISSUE #22: Document style guide
As pointed out in #20, our `.hh` extension differs from the Google style guide.  We probably also want to make explicit that `.clang-format` is authoritative in all matters of formatting, especially since we're likely to change it before release so that it differs significantly from Google's formatting guide.

chiphogg:
I skimmed through the Google C++ style guide, and looked for any
differences in how we do things.  I called out the ones I found.  We can
add more here as we encounter them.

The result can be previewed here:
https://github.com/aurora-opensource/au/blob/style-guide%2322/CONTRIBUTING.md#style-guide

To test, I clicked every link.  They all go to the right anchor,
although some of the later ones seemed to be a little "off" to me in
their location.  However, the link is definitely correct, so I think
this is a quirk we can't do anything about.

Fixes #22.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ Au follows the [Google C++ Style Guide](https://google.github.io/styleguide/cppg
 from the explicit exceptions we enumerate below.  Please make sure your PR conforms to this modified
 style guide before landing.
 
-Here are the modifications we've made to the Google style guide for Au:
+Here are the modifications we've made to the Google C++ Style Guide for Au:
 
 1. The `.clang-format` file in the repository supersedes the Google style guide in all matters of
    code formatting.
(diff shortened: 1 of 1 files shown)


johnzoppina (approved):
This looks good to me. Extra credit would be to make these list items parallel (for me, that means human subjects vs. non-human subjects, followed by parts of speech)

chiphogg:
> This looks good to me. Extra credit would be to make these list items parallel (for me, that means human subjects vs. non-human subjects, followed by parts of speech)

Sorry, I didn't quite understand the extra credit part.  :sweat_smile:

tobin (approved):

chiphogg:
> > This looks good to me. Extra credit would be to make these list items parallel (for me, that means human subjects vs. non-human subjects, followed by parts of speech)
> 
> Sorry, I didn't quite understand the extra credit part. 😅

Synced offline.  The extra credit was about resolving inconsistencies in style, but since this wasn't trivial I decided to pass.

PR #165, opened 2023-08-14
Implement `clamp(v, lo, hi)`

### LINKED ISSUE #42: Support clamp
This would naturally complement our support for `min` and `max`.

https://en.cppreference.com/w/cpp/algorithm/clamp

chiphogg:
This is similar to `std::clamp(v, lo, hi)`, introduced in C++17, but
with some interesting twists due to handling units.  Similar to `min`
and `max`, we express the return value in the common unit of all inputs.
Once again, this also means we return by value instead of reference.

One consequence, which may be counterintuitive at first, is that the
return type can be different from the type of the value we're clamping!
However, this makes sense after more careful consideration.  If we ask
to clamp a value in, say, _meters_, to a range whose bounds we express
in _millimeters_, then if the value falls outside this range, _of
course_ we'll need to express it in millimeters.  Since the return type
can't depend on whether the value is outside the range, this is
sufficient.

(Of course, I expect the usual case when units are different is that the
value is already more precise than the bounds.  And I expect the usual
_usual_ case is that the units are _not_ different.)

`QuantityPoint` is even more interesting.  It turns out we need to take
the origin offset into account whenever it exists.  Imagine clamping 0
degrees Celsius to within two values in Kelvins.  We can't compare the
value to the bounds without adding the relative offset, which can't be
expressed more coarsely than 1/20 Kelvins.  Fortunately, our "common
point unit" machinery handles this case with ease.

Includes unit tests for both `Quantity` and `QuantityPoint`, as well as
reference documentation.

Fixes #42.

### DIFF
--- a/docs/reference/math.md
+++ b/docs/reference/math.md
@@ -110,6 +110,76 @@ disambiguate our `min` or `max` implementations with respect to `std::min` and `
     support combining different units.  This means the return type will generally be different from
     the types of the inputs.
 
+#### `clamp`
+
+"Clamp" the first parameter to the range defined by the second and third.  This is a _unit-aware_
+analogue to [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp), which was introduced
+in C++17.  However, this version differs in several respects from `std::clamp`, in order to handle
+quantities more effectively.  We'll explain these differences at the end of the `clamp` section.
+
+**Signatures:**
+
+```cpp
+// 1. `Quantity` inputs
+template <typename UV , typename RV ,
+          typename ULo, typename RLo,
+          typename UHi, typename RHi>
+constexpr auto clamp(
+    Quantity<UV , RV > v,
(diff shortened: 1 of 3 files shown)


docs/reference/math.md
```diff
@@ -110,6 +110,76 @@ disambiguate our `min` or `max` implementations with respect to `std::min` and `
...
+template <typename UV , typename RV ,
+          typename ULo, typename RLo,
+          typename UHi, typename RHi>
+constexpr auto clamp(
+    QuantityPoint<UV , RV > v,
+    QuantityPoint<ULo, RLo> lo,
+    QuantityPoint<UHi, RHi> hi);
+```
+
+??? note "A note on custom comparators"
```
geoffviola, line 142: Thanks for documenting this API choice.

geoffviola (approved):
Implementation, tests, and documentation look good.

PR #166, opened 2023-08-16
Make Au compatible with `-Wconversion`

chiphogg:
The goal is for Au to not raise any warnings for users who build with
`-Wconversion`.

Note that this is not the same thing as simply enabling `-Wconversion`
by default on all of our builds.  Since Au simply "wraps" the underlying
C++ computations, it will always be easy to "make" Au violate
`-Wconversion`, simply by wrapping a non-complicant computation.
Indeed, we have a couple of tests that do this, and I think they are
important to retain.

The solution is to confine all of these tests to a single new test
target.  Then we add a separate step in the build-and-test jobs that
runs every _other_ test with `-Wconversion` enabled.  This will give us
confidence that we're not re-breaking `-Wconversion` later on.


timhirsh (approved):
GitHub action changes LGTM.  I'll defer the remainder of the review to @geoffviola

geoffviola (approved):

PR #168, opened 2023-08-22
Make most unit traits compatible with unit slots

### LINKED ISSUE #167: Some unit traits don't support unit slots
```cpp
constexpr auto cpu_ticks = inverse(hertz * mag<400'000'000>());
constexpr auto ratio = unit_ratio(cpu_ticks, nano(seconds));
```

This second line doesn't compile.  It should though!  Same for other related traits.

chiphogg:
The motivating use case was `unit_ratio`.  I was working with an ad hoc
type for CPU frequency which _only_ had a quantity maker.  See the
example from the issue:

```cpp
constexpr auto cpu_ticks = hertz * mag<400'000'000>();
```

Right now, there's no good way to get the unit ratio between this and
`Nano<Seconds>`.  Enabling `unit_ratio(cpu_ticks, nano(seconds))` seems
like a slam dunk.

There are of course many other unit traits.  I decided to update them
all while I was in there.  `is_unit` felt like it should be an
exception, so I made it one, and explained why in detail in the docs.

I also added test cases to make sure we have the desired behaviour.  In
doing so, I replaced our ad hoc definitions with the direct unit
definitions.  Making a length system based on `Feet` was kind of cute,
but we don't need it.  This also brings in the quantity makers such as
`feet`, which was the real motivation for this change.  I kept `Celsius`
as-is because several tests depend on its exact choice of type for
origin, and they're easier to understand if the definition is in the
same file.

Fixes #167.

### DIFF
--- a/au/BUILD
+++ b/au/BUILD
@@ -392,6 +392,7 @@ cc_test(
         ":prefix",
         ":testing",
         ":unit_of_measure",
+        ":units",
         "@com_google_googletest//:gtest_main",
     ],
 )
--- a/au/unit_of_measure.hh
+++ b/au/unit_of_measure.hh
@@ -195,47 +195,53 @@ constexpr bool is_unit(T) {
     return IsUnit<T>::value;
 }
 
-// Check whether two objects are Unit types of the same Dimension.
+// `fits_in_unit_slot(T)`: check whether this value is valid for a unit slot.
+template <typename T>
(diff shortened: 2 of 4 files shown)


geoffviola (approved):
Changes make sense. Thanks for updating the documentation. It's good to see improved integer support with inverse and ratio.

Issue #122, opened 2023-03-14
Consider separating "forcing" from explicit-Rep


chiphogg:
I have always conflated these ideas, probably because `static_cast` does both.  But it could be nice to be able to write, say, `inches(50).force_as(feet)`: we wouldn't need to repeat the `Rep`, and the "force" makes the potential lossiness more obvious.  This would free up `q.as<T>(unit)` to respect the safety checks.

chiphogg:
I woke up this morning with an exciting new idea for APIs.  Writing it down now so I can get to it later.

- Make an enum class, `RiskCheck`, with four values, `NONE` (00b), `OVERFLOW` (01b), `TRUNCATION` (10b), `ALL` (11b).  Note how it automatically acts as a bitmask.
- Make a constant type, templated on `RiskCheck`, and make four pre-made values:

```cpp
constexpr auto SKIP_ALL_CHECKS = RiskPolicy<RiskCheck::NONE>;
constexpr auto SKIP_TRUNCATION_RISK_CHECK = RiskPolicy<RiskCheck::OVERFLOW>;
constexpr auto SKIP_OVERFLOW_RISK_CHECK = RiskPolicy<RiskCheck::TRUNCATION>;
constexpr auto PERFORM_ALL_CHECKS = RiskPolicy<RiskCheck::ALL>;
```

I think expressing the callsite in terms of which check we are _explicitly skipping_ will better communicate intent.

- Add a second argument to `.in` and `.as`.
- Make the single-argument `.in` and `.as` delegate to the appropriate overload: `SKIP_ALL_CHECKS` for explicit-rep (for now!), and `PERFORM_ALL_CHECKS` for implicit-rep.
- Deprecate `coerce_in` and `coerce_as`: the naming was always controversial.
- Re-express all conversion policies in terms of _separate_ checks for overflow and truncation.
  - Make sure all overflow checks are simplified down to "smallest overflowing value" implementations (#349), or no-op implementations for conversions that can't overflow.
- The two-argument form of `.in` and `.as` will call out to the appropriate specializations for these separated conversion policies, _depending on the second argument_.

Benefits:

- This is a totally new signature, so people can adopt it without forcing upgrades (due to giving an old signature a new meaning, as we hope to do with explicit-rep).
- Far easier upgrade path for explicit-rep APIs
- Much better at expressing intent: some embedded use cases, for example, explicitly want to truncate, and these callsites would be much more self-documenting.

PR #171, opened 2023-08-29
Add coercing versions of conversion functions

chiphogg:
These are variants of `.as` and `.in` for both `Quantity` and
`QuantityPoint`.  When we precede these with the "coerce" vocabulary
word, they become forcing, ignoring the safety checks for truncation and
overflow.

This new syntax has two major advantages over the explicit-Rep version.
First, it makes the intent clearer.  Second, it stops forcing users to
repeat the rep when they want the same rep they started with, which I
think is the usual case.  (Thus, without these APIs, explicit-rep
obscures intent: one never knows whether the author wanted a cast, or
wanted to bypass the safety checks.)

It also unlocks another, later improvement: we will be able to extend
the safety checks to the explicit-rep versions!  But we won't attempt
that for a while because that's a breaking change.

There may be other APIs that would benefit from using the "coerce"
vocabulary word instead of explicit-rep.  However, we'll start with just
these, both because they're the overwhelmingly most common use case, and
because it gives us a chance to try out these ideas in practice for a
while.

To test this PR, I added new unit tests to cover all of the use cases.
I also rendered the docs locally and checked that the links worked as
intended.

Helps #122.

### DIFF
--- a/au/prefix_test.cc
+++ b/au/prefix_test.cc
@@ -46,7 +46,7 @@ TEST(PrefixApplier, ConvertsQuantityMakerToMakerOfCorrespondingPrefixedType) {
     constexpr auto d = inches(2);
     EXPECT_THAT(d.in(make_milli(inches)), SameTypeAndValue(2'000));
 
-    EXPECT_THAT(make_milli(inches)(5'777).in<int>(inches), SameTypeAndValue(5));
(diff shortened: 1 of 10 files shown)


geoffviola (approved):
The API changes look good. I like that the truncation/overflow at runtime is enabled with a longer, explicit name like `static_cast`. Also, it's nice not to have to repeat the rep in doing a conversion. The tests cover truncation and overflow for quantity and quantity point. The docs changes seem good.

PR #172, opened 2023-09-01
Update troubleshooting page for 0.3.3 release

chiphogg:
This will ensure that the released docs will match the released code.

It's mostly just line number updates, but I did find some misplaced
and/or extra lines.

It also looks like the windows 2022 image may have updated to a newer
version, which would be annoying if so, but their version numbering
scheme is so confusing that I don't really know what to do about it.


geoffviola (approved):
A spot check on the first line number looked good. The newline change seemed reasonable.

PR #176, opened 2023-10-02
Refine and document conversion factor policies

### LINKED ISSUE #173: Only separate num and denom for non-integer rational conversion factors
I was looking at some godbolt examples, and found that Au was less efficient at runtime in converting between revolutions and radians.  This is because when we apply a magnitude, we split it into `num`, `den`, and `irr`, and we multiply, divide, and multiply.

Keeping `num` and `den` separate makes good sense for rational numbers: it makes us more accurate when we can divide by 13, say, instead of multiplying by the floating point representation of 1/13.  But for irrational numbers, this just wastes computation.

We should see if we can apply the multiply-and-divide approach _only_ to non-integer rationals, and for everything else, simply multiply by a single number we compute at compile time.

Additionally, it'd be nice to make sure we have checks that we never use integer rep with irrational scale factors.

chiphogg:
So far, we've applied our conversion factors by first breaking them into
three parts:

- `num`, the integer part of the numerator
- `den`, the integer part of the denominator
- `irr`, the "leftover" irrational parts, if any

We've applied it as `value * num / den * irr`.

This breakdown isn't really uniquely defined if `irr != 1`, of course.
But the idea was that this was "good enough to get started", and we
could count on the compiler to optimize things away in individual cases.

We now replace this with a more thoughtful, principled approach.  It
takes into account both the numeric category of conversion factor
(integer, irrational, etc.), and the type `T` of the variable we're
scaling.

This was motivated by looking at godbolt while making the slides for
CppCon 2023, and noticing that sometimes our conversion factors used two
instructions, where the equivalent "no units library" code used only
one.

The specific instance was a "crystal clear" poor case converting between
revolutions and radians, where we applied the irrational "two pi" as a
separate `2` and `pi`.  I also investigated the "harder" case of a
_rational_ factor applied to floating point, and decided that a single
instruction made sense here too.

I also included a detailed discussion doc explaining these rules and
tradeoffs.  It's an "implementation" level doc, so it gets pretty into
the weeds, but I think certain users will find it really interesting.

Fixes #173.

### DIFF
--- a/docs/discussion/implementation/applying_magnitudes.md
+++ b/docs/discussion/implementation/applying_magnitudes.md
@@ -1,8 +1,9 @@
 # Applying Magnitudes
 
-Every unit conversion factor is a positive real number.  When we apply it to a value, conceptually,
-we're just multiplying the value by that number. However, that doesn't mean that multiplying by
-a number is the best _implementation_!  Consider these examples.
+Every unit conversion factor is a [_magnitude_](../../reference/magnitude.md) --- a positive real
+number.  When we apply it to a value, conceptually, we're just multiplying the value by that number.
+However, that doesn't mean that multiplying by a number is the best _implementation_!  Consider
+these examples.
 
 - **Factor: $\frac{5}{8}$; value: `12` (type: `int`).**
     - Computationally, we don't want to leave the integral domain.  But of course, $\frac{5}{8}$
--- a/docs/discussion/implementation/applying_magnitudes.md
+++ b/docs/discussion/implementation/applying_magnitudes.md
@@ -1,0 +1,136 @@
+# Applying Magnitudes
+
+Every unit conversion factor is a positive real number.  When we apply it to a value, conceptually,
+we're just multiplying the value by that number. However, that doesn't mean that multiplying by
+a number is the best _implementation_!  Consider these examples.
+
+- **Factor: $\frac{5}{8}$; value: `12` (type: `int`).**
+    - Computationally, we don't want to leave the integral domain.  But of course, $\frac{5}{8}$
+      can't be represented as an integer!  This suggests we should perform _two_ operations: first
+      multiply by `5`, and then divide by `8`, yielding `7`.
+- **Factor: $\frac{1}{13}$; value: `91.0f` (type: `float`).**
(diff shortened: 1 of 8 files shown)


docs/discussion/implementation/applying_magnitudes.md
```diff
@@ -0,0 +1,136 @@
+# Applying Magnitudes
+
```
johnzoppina, line 2: This page is all about **Magnitudes**, but the word "magnitude" itself isn't explicitly stated or mentioned in the introduction -- we don't see the word again until the **Magnitude Categories** section.

Is it possible that this introduction should include the content in the **Magnitude Categories** section?

johnzoppina (approved):
Just one note from me re: the intro on **Applying Magnitudes** — I don't necessarily think it's a blocking issue, but it's something we should revisit when convenient.

docs/discussion/implementation/applying_magnitudes.md
chiphogg: Nice catch --- WDYT about 21db19d?

johnzoppina: That likely contextualizes the rest of the page well. Thank you!

geoffviola (approved):
Makese sense. Tests look good.

There are tests with pi. Not sure if we need explicit performance tests around revolutions and radians. It might be nice to characterize precision and performance of the various types. Since we don't expect these conversions to be in a hot loop, it might not be as important.

chiphogg:
> Makese sense. Tests look good.
> 
> There are tests with pi. Not sure if we need explicit performance tests around revolutions and radians. It might be nice to characterize precision and performance of the various types. Since we don't expect these conversions to be in a hot loop, it might not be as important.

I'm happy with just double checking godbolt after it lands. :slightly_smiling_face:

chiphogg:
> > Makese sense. Tests look good.
> > There are tests with pi. Not sure if we need explicit performance tests around revolutions and radians. It might be nice to characterize precision and performance of the various types. Since we don't expect these conversions to be in a hot loop, it might not be as important.
> 
> I'm happy with just double checking godbolt after it lands. 🙂

Looks like we're good: https://godbolt.org/z/jjdf7v8dq

PR #178, opened 2023-10-07
Add pregenerated files with all units

### LINKED ISSUE #174: Make pre-generated "include ALL the units!!" single-file versions
The main thing that makes me want this is messing around on godbolt.  I keep needing to paste in unit definitions for things like `hertz` and `revolutions`.

It could also be a good option for people who just want to play with the library and have Everything Work, even if it's slower to compile.

We'll want significant disclaimers, though, as the compiler slowness will just get worse and worse as our unit variety gets better and better.

chiphogg:
The motivating example is to be able to use the library on godbolt.  For
this use case, we typically care far more about being comprehensive than
we do about compile times.

To enable this, we first add an `--all-units` flag to the single file
script.  This flag will cause the script to act as if the `--units` flag
had been populated with every unit that is known to the library.

(Under the hood, the script has a safety mechanism to make sure that we
actually use every command line parameter in the script.  This new
parameter would run afoul of this mechanism, because it's mutually
exclusive with the `--units` argument.  Therefore, we use `--all-units`
to populate `--units`, and then delete `--all-units`.)

Next, we create build rules that use this new flag to generate
single-file versions with all units, with and without the `<iostream>`
dependency.  Each new rule gets its own unit test.

Finally, we update the installation docs to provide this option.  We
provide an appropriate degree of discouragement for general use, and
clarity around suitable use cases.  The godbolt link is currently
"broken" (it doesn't compile), but we expect it to get fixed
automatically once this PR lands.

Test plan:

- [x] Run script with combinations of arguments:
    - [x] Neither `--units` nor `--all-units`: builds with no units.
    - [x] Only `--units`: includes specified units.
    - [x] Only `--all-units`: includes all units.
    - [x] Both `--units` and `--all-units`: exits with error.
- [x] Add unit tests for both single file generated versions. 
- [x] View rendered docs and follow links

Fixes #174.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -29,6 +29,8 @@ py_binary(
     srcs = ["update_docs.py"],
     data = [
         "mkdocs.yml",
+        ":au_all_units_hh",
+        ":au_all_units_noio_hh",
         ":au_hh",
         ":au_noio_hh",
     ] + glob(["docs/**"]),
@@ -43,6 +45,8 @@ py_binary(
     srcs = ["mike_bin.py"],
     data = [
         "mkdocs.yml",
+        ":au_all_units_hh",
+        ":au_all_units_noio_hh",
         ":au_hh",
         ":au_noio_hh",
     ] + glob(["docs/**"]),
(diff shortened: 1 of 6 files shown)


geoffviola (approved):
Thanks for adding this. This PR should make it easier to share code.

PR #180, opened 2023-10-15
Update release instructions

chiphogg:
We now include running the commented-out test cases.  This closes a gap
in our testing.  (It's not as good as running them automatically on
every commit, but we don't have an easy way to do that.)

We also make sure to check the cumulative compile time impact, to make
sure no significant regressions have slipped in without our notice.


RELEASE.md
```diff
@@ -4,6 +4,12 @@
 
 These are the steps to take before you actually cut the release.
 
+### Check all commented-out test cases
+
+These are test cases that we can't test automatically, usually because the intended behavior is
+a compiler error.  Grep the codebase for `uncomment`, and test every such case individually to make
```
timhirsh, line 10: I thought there might be a cleaner way to handle this (eg. using `manual` tags and other ways to isolate these failure cases), but I didn't really see anything aside from running bazel inside of bazel as an integration test.  If this project begins to accumulate a lot of these we may want to look into it further, but for now this seems ok as a release step.

timhirsh (approved):

RELEASE.md
chiphogg: Yeah... it's really tricky, because you do need to uncomment only one at a time.  I wish there were a good general solution for this.

Issue #90, opened 2023-02-06
Support constants


chiphogg:
In some sense, we can already do this.

```cpp
constexpr auto RAD = radians(1);
```

However, this brings `int` into the equation, via the `1`.  This could affect the arithmetic of the underlying Reps in some equations where it gets used.  It could also run afoul of our guards against integer division.

It would be nice if we could express constants _solely in the realm of dimensions and magnitudes_.  We might need to make a new type template, such as `Constant`.  It would probably be a "monovalue type" (in the nomenclature introduced in #88).  Perhaps something like this:

```cpp
constexpr auto RAD = make_constant(Radians{});
```

Alternatively: maybe a "constant" is just an instance of a _unit_?  We would need to enable multiplying and dividing a quantity by a unit, if so.  Then we would write:

```cpp
constexpr auto RAD = Radians{};
```

It's interesting to imagine how this would look with other constants.  Consider something like $E = mc^2$.  We might write `kilo(grams)(10.0) * squared(C)`... and the result might be labeled as `10 kg * c^2`.  Of course we could also apply `.as(joules)` to the result to get a more familiar unit.

This is an intriguing idea worth exploring more later.

chiphogg:
Having thought more about this: no, a constant should _not_ just be an instance of a unit.  A "unit" could be basically any type, so we'd need SFINAE, which would be expensive.  This argues in favor of using a `Constant` template.

It's probably still good to keep this template as a [monovalue type](https://aurora-opensource.github.io/au/reference/detail/monovalue_types/), and use a labeled unit.

chiphogg:
Here's a running checklist for what it would take to call this feature "done".  (I'll check things off as they're implemented locally in my client, rather than on remote.)

- [ ] Multiplication and division
   - [ ] Regular numbers
   - [ ] `Quantity`
   - [ ] `Magnitude`
   - [ ] Other `Constant`
   - [ ] `QuantityMaker`
   - [ ] `SingularNameFor`
   - [ ] `QuantityPoint` (`= delete`)
   - [ ] `QuantityPointMaker` (`= delete`)
- [ ] Explicit conversions to `Quantity`
   - [ ] `.as<T>()`: initialize with `T{1}` in units of the constant
   - [ ] `.as<T>(unit)`: choose unit-and-rep
      - [ ] Exact representability checks (no "safety surface" heuristics)
   - [ ] `.coerce_as<T>(unit)`: choose unit-and-rep, ignore safety checks
[...]

PR #183, opened 2023-10-21
Refactor `get_value<T>(m)` to enable success query

chiphogg:
Right now, `get_value<T>(m)` for a magnitude `m` either produces the
requested value, or invokes a hard compiler error via `static_assert`.
It would be nice if we had a way to ask whether this operation would
succeed.  For this purpose, we now provide `representable_in<T>(m)`.

It turns out that we can't actually answer this question without
effectively doing all of the work to compute the value.  We won't want
to do this twice, because that would uselessly slow down compile times.
Therefore, we extract all of our logic into a common implementation
function.  The logic is basically equivalent to the pre-existing version
of `get_value<T>(m)`, except that instead of each `static_cast`, we
return a unique value of a new `enum` we created for this purpose.

This means that each function can get its result efficiently.  For
`get_value`, we change the trigger for each `static_assert` to the
corresponding enum value.  For `representable_in`, we simply check
whether that enum has the `OK` value.

This PR helps pave the way for #90.  When we make our `Constant` class
implicitly convertible to `Quantity` types, it turns out that we can do
better than the "overflow safety surface"... because we know the exact
value!  We can achieve perfect compile-time checking.  This PR will give
us a single function that we can call in our implementation for that.

Includes docs and tests.

### DIFF
--- a/docs/reference/magnitude.md
+++ b/docs/reference/magnitude.md
@@ -127,6 +127,38 @@ To extract the value of a `Magnitude` instance `m` into a given numeric type `T`
     Thus, `get_value<float>(pow<3>(PI))` will be much more accurate than storing $\pi$ in a `float`,
     and cubing it --- yet, there will be no loss in performance.
 
+### Checking for representability
+
+If you need to check whether your magnitude `m` can be represented in a type `T`, you can call
+`representable_in<T>(m)`.  This function is `constexpr` compatible.
+
+??? example "Example: integer and non-integer values"
+
+    Here are some example test cases which will pass.
+
+    ```cpp
+    EXPECT_TRUE(representable_in<int>(mag<1>()));
+
+    // (1 / 2) is not an integer.
+    EXPECT_FALSE(representable_in<int>(mag<1>() / mag<2>()));
+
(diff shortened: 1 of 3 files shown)


docs/reference/magnitude.md
```diff
@@ -127,6 +127,38 @@ To extract the value of a `Magnitude` instance `m` into a given numeric type `T`
...
+
+    ```cpp
+    EXPECT_TRUE(representable_in<uint32_t>(mag<4'000'000'000>()));
+
+    // 4 billion is larger than the max value representable in `int32_t`.
+    EXPECT_FALSE(representable_in<int32_t>(mag<4'000'000'000>()));
+    ```
+
+Note that this function's return value also depends on _whether we can compute_ the value, not just
+whether it is repreesntable.   For example, `representable_in<double>(sqrt(mag<2>()))` is currently
```
geoffviola, line 159:
```suggestion
whether it is representable.   For example, `representable_in<double>(sqrt(mag<2>()))` is currently
```

geoffviola (approved):

PR #184, opened 2023-10-23
Update urllib to fix dependabot alert 6

chiphogg:
This didn't affect any of our primary dependencies.  Here are the
release notes for our affected secondary dependencies:

- `charset-normalizer`: 3.3.0 to 3.3.1
  (https://github.com/Ousret/charset_normalizer/blob/master/CHANGELOG.md#changelog)
- `mkdocs-material-extensions`: 1.2 to 1.3
  (https://github.com/facelessuser/mkdocs-material-extensions/blob/master/changelog.md#13)
- `pymdown-extensions`: 10.3 to 10.3.1
  (https://facelessuser.github.io/pymdown-extensions/about/changelog/#1031)
- `urllib3`: 2.0.6 to 2.0.7
  (https://github.com/urllib3/urllib3/releases/tag/2.0.7)

The last one is the motivation for this PR.

The only other significant one is the `mkdocs-material-extensions`:
version 1.3 deprecates the project, because its contents have been
folded in to the parent project `mkdocs-material`.  I followed the
instructions in the release notes, resulting in the changes to
`mkdocs.yml`.  I found this file by grepping for `materialx`, so this
should be the only change we need to make.

To test this change, I ran `au-docs-serve` and did a spot check blink
comparison on several pages.  In particular, I included the magnitude
reference page, because I know it has some emoji (checks and x's,
mainly).  I didn't see any visible change on any page.


timhirsh (approved):
👍

PR #188, opened 2023-11-08
Add unit-aware overload for `std::remainder`

chiphogg:
This is basically like `std::fmod`, but it's zero-centered.  It's really
handy for things like rectifying angles when we'd rather express
`degrees(270)` as `degrees(-90)`.

To make this change, I grepped the codebase for `fmod`, and simply made
whatever change was reasonable.  One of these changes was to factor out
a helper to make it more generic; I did this in a separate "checkpoint
commit" on the PR to make it easier to review.

### DIFF
--- a/au/math_test.cc
+++ b/au/math_test.cc
@@ -213,10 +213,20 @@ TEST(cos, GivesCorrectAnswersForInputsInDegrees) {
 // If its inputs have the same unit, then there is no conversion, only computation.  In that case,
 // we want to make sure we're doing exactly what std::fmod does w.r.t. input and output types (see:
 // https://en.cppreference.com/w/cpp/numeric/math/fmod).
-template <typename U, typename R1, typename R2>
-void expect_consistent_with_std_fmod(Quantity<U, R1> q1, Quantity<U, R2> q2) {
-    EXPECT_THAT(fmod(q1, q2),
-                QuantityEquivalent(make_quantity<U>(std::fmod(q1.in(U{}), q2.in(U{})))));
+template <typename AuFunc, typename StdFunc>
+struct ExpectConsistentWith {
+    template <typename U, typename R1, typename R2>
+    void operator()(Quantity<U, R1> q1, Quantity<U, R2> q2) const {
+        EXPECT_THAT(au_func(q1, q2),
+                    QuantityEquivalent(make_quantity<U>(std_func(q1.in(U{}), q2.in(U{})))));
+    }
+
+    AuFunc au_func;
+    StdFunc std_func;
+};
+template <typename AuFunc, typename StdFunc>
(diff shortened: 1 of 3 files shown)


au/math_test.cc
```diff
@@ -248,6 +262,43 @@ TEST(fmod, HandlesIrrationalCommonUnit) {
...
+TEST(remainder, ReturnsSameTypesAsStdRemainderForSameUnitInputs) {
+    const auto expect_consistent_with_std_remainder =
+        expect_consistent_with([](auto x, auto y) { return remainder(x, y); },
+                               [](auto x, auto y) { return std::remainder(x, y); });
+
+    expect_consistent_with_std_remainder(meters(4), meters(3));
+    expect_consistent_with_std_remainder(meters(4.f), meters(3.f));
+    expect_consistent_with_std_remainder(meters(4.), meters(3.));
+    expect_consistent_with_std_remainder(meters(4.l), meters(3.l));
+    expect_consistent_with_std_remainder(meters(4), meters(3.f));
```
geoffviola, line 279: Q: Why choose these mismatched types?

au/math_test.cc
```diff
@@ -248,6 +262,43 @@ TEST(fmod, HandlesIrrationalCommonUnit) {
...
+    EXPECT_THAT(remainder(a, b), IsNear(expected_result, make_quantity<Nano<Meters>>(1)));
+}
+
+TEST(remainder, HandlesIrrationalCommonUnit) {
+    EXPECT_THAT(remainder(radians(1), degrees(57)), IsNear(degrees(+0.2958), degrees(0.0001)));
+    EXPECT_THAT(remainder(radians(1), degrees(58)), IsNear(degrees(-0.7042), degrees(0.0001)));
+}
+
+TEST(remainder, CenteredAroundZero) {
+    EXPECT_THAT(remainder(degrees(90), revolutions(1)), IsNear(degrees(90), degrees(1e-9)));
```
geoffviola, line 298: Are the underlying types integers? Can this be exactly equals?

geoffviola (approved):

au/math_test.cc
chiphogg: The goal of this test case is to make sure that we handle combinations of _reps_ the same way that `std::remainder` handles combinations of _types_.  Adding a few mismatched-rep examples gives us better coverage for that property.

au/math_test.cc
chiphogg: The inputs are integers, but the outputs aren't.

I expect we _could_ do exactly-equals here, in _practice_, but because the underlying STL function is using floating point math, I don't feel confident that I could guarantee it.

chiphogg (commented):
Good questions!

PR #193, opened 2023-11-16
Make godbolt link more presentable

chiphogg:
A few changes:
- Add `-O1`, the simplest level of optimization, which gets rid of some
  weirdness in the old link (try writing `meters(1)` for example).
- Use a newer version of gcc (13.2 instead of 7.x).
- Stack the side panes vertically to be a little less claustrophobic.

We also present the godbolt link more prominently on our main README.


geoffviola (approved):
Links look good.

Issue #52, opened 2022-12-22
Support non-arithmetic Rep


chiphogg:
As mentioned in #38, it's not our intention to restrict Rep to built-in arithmetic types forever.  This issue tracks the idea of defining a separate trait which specifies _exactly_ which properties are needed for a valid Rep.

Note that as part of this effort, we would need to check the impact on compile times.

chiphogg:
To be clear: we already have experimental support for non-`is_arithmetic` Rep.  Apparently, people have been using this, and it's been working fine so far.  However, it would be nice to add constraints that specify exactly what a type needs in order to be a valid Rep.  This could give more direct error messages for end users.

chiphogg:
Here's a good "kick the tires" playground using the bbi library: https://godbolt.org/z/sqv1dv9ja

Issue #43, opened 2022-12-16
Support abbreviated symbols


chiphogg:
We never migrated these upstream!  It'd be nice to have them in place before we announce the library.

We could include the UDL inside each `"au/units"` file, but that makes me wary.

- Having split out units into individual files makes the cost of adding a new unit virtually zero.  Literals could change that, because we would have to worry about literal collision (e.g., `farads` and `fahrenheit` might each be tempted to use `_F`).
- Some literals correspond to units that don't _have_ an `"au/units"` file, e.g., `_MPH`, `_nm`, `_kg`.
- We've been doing really well on the "don't pay for what you don't use" front, and I'd like to keep that up.

-----

So here's a proposal.

1.  Define the UDLs inside an `"au/units/literals"` file (or, perhaps, simply `"au/literals"`), exactly corresponding to the `"au/units"` file.  So, either `"au/units/literals/meters.hh"`, or `"au/literals/meters.hh"`.
2. The literals file would automatically include the units file.  So, `"au/units/literals/miles_per_hour.hh"` would give `"au/units/miles.hh"` and `"au/units/hours.hh"`.
3. The single-file script would gain a new `--literals` argument which would act just like `--units`, but for the literals folder.
4. The pre-built single-file packages _would_ include literals.

chiphogg:
We should interpret this issue more broadly, and not wed ourselves to UDLs as the solution.  What we're really trying to do is provide a nice way for people to express themselves concisely using unit symbols.  @mpusz has expounded elsewhere on the downsides of UDLs, and they are indeed significant.  Perhaps with constants (#90), there might be a more flexible and ergonomic way to include the units concisely.

avrahamshukron:
Hi @chiphogg !
I'm really looking forward for this feature.

not sure if this is the correct place to discuss this, but I was trying to implement UDL for Au-based unit types, and I'm struggling to enforce safety at compile-time.

Lets assume I want `_mV` to return milli-volts represented as an `int32_t` (because the code is targeting embedded device) , i.e
```cpp
using Voltage = au::QuantityI32<au::Milli<au::Volts>>;
```

The trouble is that UDLs for integral types are forced to accept `unsigned long long` as an argument:
```cpp
constexpr Voltage operator""_mV(unsigned long long literal)
{
    if (literal > static_cast<unsigned long long>(std::numeric_limits<Voltage::Rep>::max()))
    {
        // Fail compilation somehow?
    }
[...]

PR #194, opened 2023-11-19
Add mixins to help with constants and symbols

chiphogg:
Constants (#90) and symbols (#43) have a lot of similarities.  For
example:

- Multiplying by a raw number produces a quantity.
- Multiplying by a quantity changes the units of the quantity.
- They can compose with other instances of the same family (e.g., the
  product of two constants is a constant).

Let's use the term "wrapper" to refer to constants and symbols here. To
make each wrapper as easy as possible to implement, I've created some
"mixin" classes, each of which adds the full set of multiplication and
division operations for a single combination of the wrapper and some
other family of types (raw numbers, quantities, other wrappers, etc.).

These wrappers use a "CRTP-ish" syntax.  But instead of the _type
itself_ being the first template parameter (e.g., `Wrapper<Unit>`), we
provide the _wrapper_ and _unit_ as two separate parameters (i.e.,
`Wrapper` and `Unit`).  The reason is that we need to know the `Unit`,
and providing it explicitly makes it easy to get.  These first two
parameters are the same for all mixins, which makes the list-of-mixins
at the class definition easier to read.

The unit tests are based on an example, `UnitWrapper`, which aggregates
a variety of properties.  The forthcoming "true" classes, `Constant` and
`SymbolFor`, will be defined very similarly to this.

Finally, a word about the plan for "numeric" inputs.  For now, we are
restricting to `std::is_arithmetic`.  This doesn't mean we don't support
other reps; it just means we won't be able to create quantities from
them via constants or symbols for a while.  The evolution plan is to
create a well-defined concept that defines what is a valid rep, and then
replace `std::arithmetic` with that concept (see also #52).

Test plan:

- [x] Add extensive new unit tests
- [x] Manually uncomment each individual "uncomment to test" case

### DIFF
--- a/au/wrapper_operations_test.cc
+++ b/au/wrapper_operations_test.cc
@@ -1,0 +1,163 @@
+// Copyright 2023 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "au/wrapper_operations.hh"
+
+#include "au/testing.hh"
+#include "au/units/liters.hh"
+#include "au/units/meters.hh"
+#include "au/units/moles.hh"
+#include "gtest/gtest.h"
+
+using ::testing::StaticAssertTypeEq;
+
+namespace au {
+namespace detail {
+namespace {
+
+template <typename Unit>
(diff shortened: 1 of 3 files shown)


au/wrapper_operations_test.cc
```diff
@@ -0,0 +1,163 @@
...
+namespace {
+
+template <typename Unit>
+struct UnitWrapper : MakesQuantityFromNumber<UnitWrapper, Unit>,
+                     ScalesQuantity<UnitWrapper, Unit>,
+                     ComposesWith<UnitWrapper, Unit, UnitWrapper, UnitWrapper>,
+                     ComposesWith<UnitWrapper, Unit, QuantityMaker, QuantityMaker>,
+                     ForbidsComposingWith<UnitWrapper, Unit, QuantityPointMaker>,
+                     ForbidsComposingWith<UnitWrapper, Unit, QuantityPoint>,
+                     CanScaleByMagnitude<UnitWrapper, Unit> {};
```
chiphogg, line 36: Note to reviewers: this is probably the best starting point.  The goal of this PR is to enable us to write something like this.  The test cases below demonstrate what capabilities this kind of definition provides.  Then, with that motivation, you can check out the implementations in the previous file.

Issue #139, opened 2023-06-21
Delete old-style, deprecated .in and .as APIs


chiphogg:
This is a follow-on to #32.  These have been deprecated for long enough that it's definitely safe to delete them.  Since this should force a minor version bump, we'll save this for 0.4.0.

chiphogg:
I'm making this issue more general, to just "delete all the `[[deprecated]]` stuff", because we're also going to deprecate `PI` in 0.3.5 as part of #246, and I don't want to forget to remove it.

PR #195, opened 2023-11-20
Unify `in`/`as` implementations using unit slots

chiphogg:
With a little more care in our definitions, we no longer need separate
overloads for `QuantityMaker`!  We can just use a unified definition.

Someday, this will also let us replace SFINAE with more readable
`static_assert`, which is good on the assumption that nobody will ever
call `in` or `as` with something that is not compatible with a unit
slot.  Unfortunately, this is blocked on #139, because sometimes
function calls can get resolved to the deprecated legacy APIs... and
this manifests in really weird ways, like the compiler thinking we want
"units" of `int` or `double`.

This paves the way for #43, because `SymbolFor` will also be compatible
with unit slots.  So, not only are we removing one set of overloads;
we're also avoiding the need to add _another_ set in the very near
future.

Test plan:

- [x] All Au tests pass
- [x] Make sure this would enable upcoming `.in(m)` test to pass
- [x] Test AV repo
- [x] Measure compile time impact

### DIFF
--- a/au/conversion_policy.hh
+++ b/au/conversion_policy.hh
@@ -77,8 +77,10 @@ struct PermitAsCarveOutForIntegerPromotion
 template <typename Rep, typename ScaleFactor>
 struct ImplicitRepPermitted : detail::CoreImplicitConversionPolicy<Rep, ScaleFactor, Rep> {};
 
-template <typename Rep, typename SourceUnit, typename TargetUnit>
-constexpr bool implicit_rep_permitted_from_source_to_target(SourceUnit, TargetUnit) {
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
Changes make sense. Deleting overloads seems like an improvement.

chiphogg:
Incidentally, there's a likely future change where we do this for `QuantityPoint` as well.  The blocker there is to create a new trait, something like `AssociatedPointUnitT`, that maps `QuantityPointMaker` but not `QuantityMaker` or the upcoming `SymbolFor`.  This is also just intrinsically less motivated than the present PR, because there's nothing it really unblocks.  A likely time to do this would be when we tackle #139.

PR #199, opened 2023-11-25
Add `as_chrono_duration()` utility

### LINKED ISSUE #187: Add utility to convert Au time quantity to chrono duration
We could call this `to_chrono()` or `as_chrono()`.

chiphogg:
This makes it easy to convert any Au duration quantity to an
exactly-equivalent `std::chrono::duration`.

Fixes #187.

### DIFF
--- a/au/au.hh
+++ b/au/au.hh
@@ -34,4 +34,14 @@ struct CorrespondingQuantity<std::chrono::duration<RepT, Period>> {
     static constexpr ChronoDuration construct_from_value(Rep x) { return ChronoDuration{x}; }
 };
 
+// Convert any Au duration quantity to an equivalent `std::chrono::duration`.
+template <typename U, typename R>
+constexpr auto as_chrono_duration(Quantity<U, R> dt) {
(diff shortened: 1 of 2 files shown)


au/au.hh
```diff
@@ -34,4 +34,14 @@ struct CorrespondingQuantity<std::chrono::duration<RepT, Period>> {
     static constexpr ChronoDuration construct_from_value(Rep x) { return ChronoDuration{x}; }
 };
 
+// Convert any Au duration quantity to an equivalent `std::chrono::duration`.
```
geoffviola, line 37: Q: au.hh's two functions seem to be `std::chrono` related. Is it worth breaking that out into its own files?

geoffviola (approved):
Implementation and test look good.

au/au.hh
chiphogg: That's a good point.  I'll do that in a follow-up PR.

chiphogg: #201.

PR #202, opened 2023-11-28
Use perfect conversion policy in inverse functions

### LINKED ISSUE #190: Overflow safety surface makes inverse functions too strict
This should compile:

```cpp
inverse_as(nano(seconds), hertz(10));
```

It doesn't, because we're using `int` (generally 32 bits).  When we convert `1` to units of `(hertz * nano(seconds))`, we multiply by `1'000'000'000`, which is too-close-for-comfort to the maximum `int` value.

This is a great use case for constants!  Once we get them with #90, they'll enable exact checking.  So we can replace a `Quantity` with a value of `1` by a constant, `ONE`, and simply call `.in<Rep>(unit_product)`.  I've already confirmed in my local branch that this fix will work!  Just need to get constants ready for review.

chiphogg:
Now that we have constants, we can replace the overflow safety surface
heuristic with exact checks.  This PR uses constants to do this.  It
adds a new unit test that would have failed before these implementation
changes.  It also includes an uncomment-to-check test case which still
would fail, as it must.

Fixes #190.

### DIFF
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -240,6 +240,7 @@ cc_library(
     name = "math",
     hdrs = ["math.hh"],
     deps = [
+        ":constant",
         ":quantity",
         ":quantity_point",
         ":units",
--- a/au/math.hh
+++ b/au/math.hh
@@ -19,6 +19,7 @@
 #include <limits>
 #include <type_traits>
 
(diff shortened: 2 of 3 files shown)


geoffviola (approved):
Looks useful. The test looks good.

PR #203, opened 2023-11-28
Add discussion of overflow and how to mitigate it

chiphogg:
This finally --- _finally!_ --- gives us an authoritative place to link for
explaining the "overflow safety surface".  I was going to make that the only
topic of the page, but as I wrote I realized that there's a lot more value in
discussing the overflow problem more generally!  I expect this link will be a
useful reference for other units libraries as well.

I adapted some contents that were hidden away a couple layers deep in the 103
tutorial, and replaced those contents with a link to the new page.


docs/discussion/concepts/overflow.md
```diff
@@ -0,0 +1,208 @@
...
+conditions.  However, it doesn't translate well to a _multi-dimensional_ units library: since there
+are many dimensions, and new ones can be created on the fly, it's infeasible to try to define
+a "practical range" for _all_ of them.  Besides, users can still form arbitrary
+`std::chrono::duration` types, and they may not realize the safety they have given up in doing so.
+
+### Adapt to risk
+
+Fundamentally, there are two contributions to the level of overflow risk:
+
+1. The _size of the conversion factor_: **bigger factors** mean **more risk**.
```
tobin, line 90: In the example you gave earlier, it's not the magnitude of the conversion factor between meters and yards that is the problem (indeed that conversion factor doesn't even exist in the integral domain) but the conversion factor to the implied common unit, or the magnitudes of the numerator and denominator of the conversion factor. Not sure whether this needs to be called out explicitly, but I think you have segued somewhat from the original motivating example.

docs/discussion/concepts/overflow.md
```diff
@@ -0,0 +1,208 @@
...
+
+Therefore, we should be able to create an _adaptive policy_ that takes these factors into account.
+The key concept is the "smallest overflowing value".  For every combination of "conversion factor"
+and "type," there is some smallest starting-value that will overflow.  The simplest adaptive policy
+is to forbid conversions when that smallest value is "small enough to be scary".
+
+How small is "scary"?  Here are some considerations.
+
+- Once our values get over 1,000, we can consider switching to a larger SI-prefixed version of the
+  unit.  (For example, lengths over $1000\,\text{m}$ can be approximated in $\text{km}$.)  This
```
tobin, line 103: This mention of "approximation" sort of triggers my anxiety. "Approximation?!! I thought our unit library was exact!" etc

docs/discussion/concepts/overflow.md
```diff
@@ -0,0 +1,208 @@
...
+While the overflow safety surface is a leap forward in safety and flexibility, it's still only
+a heuristic.  There will always be valid conversions which it forbids, and invalid ones which it
+permits.  On the latter point, note that adding an intermediate conversion can defeat the safety
+check: the overflow in `meters(10u).as(nano(meters))` would be caught, but the overflow in
+`meters(10u).as(milli(meters)).as(nano(meters))` would not.
+
+One way to _guarantee_ doing better is to check every conversion at runtime.  Some users may recoil
+at the idea of doing _runtime_ work in a units library, but it's easy to show that _this_ use case
+is innocuous. Consider: it's very hard to imagine a valid use case for performing unit conversions
+in a "hot loop".  Therefore, the extra runtime cost --- merely a few cycles at most --- won't
```
tobin, line 139: Care must be taken to avoid "implicit" unit conversions inside a hot loop, no? For example your motivating example of evaluating `a>b`.

docs/assets/overflow-safety-surface.png
tobin, line 1: Is it obvious that the points on the line are "permitted"?

tobin (approved):
As always, a pleasure to read, and I always learn something new! (I always assumed that feet and meters must have some irrational or at least many-significant-figures conversion factor, for example. Thus it was a nice surprise to see how the conversion can be done with integral types.)

docs/assets/overflow-safety-surface.png
chiphogg: It certainly wasn't, so I added a legend.

I also tweaked the text to use the same phrase as the legend.  (I guess I could have just used the original text in the legend, but I didn't see it until I was done updating the figure.)

docs/discussion/concepts/overflow.md
chiphogg: Good point.  I ended up going for a footnote here.  I considered discussing the kind of overflow that can happen with rational conversion factors applied to integral-rep quantities.  However, in the context of _this section_, we're talking about conversions with safety checks applied... and rational conversion factors would already fail the safety check for truncation, so they're not really in scope here.

docs/discussion/concepts/overflow.md
chiphogg: Hmm... I switched it from "approximated" to "more concisely expressed".  I hope it makes for a smoother read.  :slightly_smiling_face:

docs/discussion/concepts/overflow.md
chiphogg: True.  Tweaked "performing" to "needing to perform".  I didn't see an easy way to bring up the _implicit_ conversions _explicitly_ here without hurting the flow.

(But yeah, unintended implicit conversions could hurt performance.  I think the approach here would be something like what I used to hear at Google: "if you care about performance, and you don't have a benchmark, then you don't care about performance".  :grin:  I also got a question at CppCon 2023 about the possibility of making **all** unit conversions explicit --- which, although we're not planning to do it, _would_ solve this kind of problem.)

PR #205, opened 2023-12-02
Stop shadowing `ONE`

chiphogg:
Our loosening of the overflow safety surface introduced a warning, which
is prominently visible on our [canonical godbolt link]:

image

Oops!  We shadowed the _magnitude_ `ONE` (globally available in the
library) with the _constant_ `ONE`.

The quick fix is to rename the newcomer (which is not publicly visible
anyway) to `UNITY`.

To prevent this problem from recurring, we also add `-Wshadow` to our
regular compiler options.

[canonical godbolt link]: https://godbolt.org/z/KrvfhP4M3


geoffviola (approved):
The name seems fine. The compiler warning works as expected. Thanks for removing the warning!

PR #206, opened 2023-12-09
Let `u` in `unit_label(u)` be a unit slot

chiphogg:
I just noticed this wasn't already the case, and this seemed silly and"
annoying.

I also added a missing test case for the mechanism that makes unit slots
work, namely, `AssociatedUnitT`.

And I updated the docs.


geoffviola (approved):
The new API looks more readable.

PR #207, opened 2023-12-09
Convert `Constant` to other libraries' types

chiphogg:
If users have set up an exact equivalence with another library's types,
we'll now be able to convert any appropriate `Constant` directly to
those types.  We even take advantage of our exact conversion policy.


geoffviola (approved):
Test and use case look good.

PR #208, opened 2023-12-09
Add most basic runtime conversion checkers

chiphogg:
To start out with, this is only for `Quantity` (not `QuantityPoint`),
and only for same-rep conversions (so no explicit-rep variants).  The
plan is to start getting some experience with these, check them out on
godbolt, make some runtime-checking converters that use them, etc.  We
can follow up later with a more complete set of APIs.  Oh, and docs, too
--- this is starting out as an undocumented feature.

Every utility's signature looks something like:

```cpp
constexpr bool is_foo(Quantity q, TargetUnit target);
```

The instances of `is_foo` that we provide are specifically:

- `will_conversion_overflow`
- `will_conversion_truncate`
- `is_conversion_lossy`

The third is the disjunction of the first two.  So we would use them
something like this:

```cpp
constexpr bool ok = is_conversion_lossy(inches(36), feet);
```

The above would return `false`, but would return `true` if we replace
`inches(36)` with `inches(37)`.

On the implementation side, each magnitude-applying type gets its own
dedicated utility.  This is nice, because then the function that checks
each condition appears directly next to the function that applies the
magnitude, so it's easy to check that they're consistent!  (For example,
when applying a rational magnitude to an integral type, we check whether
the _numerator alone_ would cause overflow, because we know we'd be
multiplying by that numerator _before_ dividing by the denominator.)

Along the way, we also make a new type trait to make it easy to get the
appropriate type of magnitude applicator.  This made it a lot easier to
write the tests.

And speaking of tests, we concentrate the vast majority of them in the
`detail`-namespaced helpers, which separately check for overflow and
truncation.  For the `:quantity_test` tests, we just pick one
representative test case that is just complicated enough to give us
confidence that we're correctly using the utilities.

The most exciting test case is the one for `is_conversion_lossy()`.  We
take one type (`uint8_t`), and check for **every representable value**
that a round-trip unit coversion is the identity, _if and only if_ we
have identified that conversion as "not lossy".  We got 100% correct
results for both inches-to-feet (truncation), and feet-to-inches
(overflow).  Nice!

### DIFF
--- a/au/apply_magnitude_test.cc
+++ b/au/apply_magnitude_test.cc
@@ -128,5 +128,151 @@ TEST(ApplyMagnitude, MultipliesSingleNumberForIrrationalMagnitudeOnFloatingPoint
     EXPECT_THAT(apply_magnitude(2.0f, PI), SameTypeAndValue(2.0f * static_cast<float>(M_PI)));
 }
 
+TEST(ApplyMagnitude, DetectsOverflowForIntegerMultiply) {
+    auto ONE_BILLION = pow<9>(mag<10>());
+
+    {
+        using ApplyOneBillionToI32 = ApplyMagnitudeT<int32_t, decltype(ONE_BILLION)>;
+
+        EXPECT_TRUE(ApplyOneBillionToI32::would_overflow(3));
+
+        EXPECT_FALSE(ApplyOneBillionToI32::would_overflow(2));
+        EXPECT_FALSE(ApplyOneBillionToI32::would_overflow(0));
+        EXPECT_FALSE(ApplyOneBillionToI32::would_overflow(-2));
+
+        EXPECT_TRUE(ApplyOneBillionToI32::would_overflow(-3));
+    }
+
+    {
+        using ApplyOneBillionToU8 = ApplyMagnitudeT<uint8_t, decltype(ONE_BILLION)>;
+
+        EXPECT_TRUE(ApplyOneBillionToU8::would_overflow(1));
+
+        EXPECT_FALSE(ApplyOneBillionToU8::would_overflow(0));
+    }
+
+    {
+        using ApplyOneBillionToF = ApplyMagnitudeT<float, decltype(ONE_BILLION)>;
+
+        EXPECT_TRUE(ApplyOneBillionToF::would_overflow(3.403e29f));
+
+        EXPECT_FALSE(ApplyOneBillionToF::would_overflow(3.402e29f));
+        EXPECT_FALSE(ApplyOneBillionToF::would_overflow(-3.402e29f));
+
+        EXPECT_TRUE(ApplyOneBillionToF::would_overflow(-3.403e29f));
+    }
+}
+
+TEST(ApplyMagnitude, AlwaysDetectsNoOverflowForIntegerDivide) {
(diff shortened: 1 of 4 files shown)


chiphogg:
It turns out, this PR's implementation isn't quite right.  Due to integer promotion, the type of `T * T` can be different than the type of `T`, for small types like `uint8_t` or `int16_t`.  This results in us being a little too conservative, and labeling some conversions as lossy that actually work fine.  We can expose this by testing the conversions for a 16-bit type between meters and yards.

This only occurs for (non-integer, non-inverse-integer) rational factors applied to integral types, and only when the integral types are small enough for integer promotion to come into play.

I think the best approach is to solve this in a follow-up PR at some point.  Otherwise, it would probably add too much complexity to this PR.

au/quantity_test.cc
```diff
@@ -659,6 +659,106 @@ TEST(QuantityMaker, ProvidesAssociatedUnit) {
...
+
+    EXPECT_TRUE(will_in_to_ft_truncate_i32(-119));
+    EXPECT_FALSE(will_in_to_ft_truncate_i32(-120));
+    EXPECT_TRUE(will_in_to_ft_truncate_i32(-121));
+}
+
+TEST(IsConversionLossy, CorrectlyDiscriminatesBetweenLossyAndLosslessConversions) {
+    // We will check literally every representable value in the type, and make sure that the result
+    // of `is_conversion_lossy()` matches perfectly with the inability to recover the initial value.
+    auto test_round_trip_for_every_uint8_value = [](auto source_units, auto target_units) {
```
geoffviola, line 728: Did you mean for this to be named uint16?

au/quantity_test.cc
```diff
@@ -659,6 +659,106 @@ TEST(QuantityMaker, ProvidesAssociatedUnit) {
...
+        for (int i = 0; i <= 65535; ++i) {
```
geoffviola, line 729: Does it make sense to do the follow?
```c++
for (int i = numeric_limits<uint16_t>::min(); i <= numeric_limits<uint16_t>::max() ; ++i) {
```

au/apply_magnitude_test.cc
```diff
@@ -128,5 +128,306 @@ TEST(ApplyMagnitude, MultipliesSingleNumberForIrrationalMagnitudeOnFloatingPoint
...
+
+        EXPECT_TRUE(ApplyOneBillionToU8::would_overflow(1));
+
+        EXPECT_FALSE(ApplyOneBillionToU8::would_overflow(0));
+    }
+
+    {
+        using ApplyOneBillionToF = ApplyMagnitudeT<float, decltype(ONE_BILLION)>;
+
+        EXPECT_TRUE(ApplyOneBillionToF::would_overflow(3.403e29f));
```
geoffviola, line 157: Is there any significance to this value or was it found emperically?

geoffviola (approved):
Looks very useful. The tests cover a lot of cases. The feature seems ready for a soft release.

au/apply_magnitude_test.cc
chiphogg: It is one one-billionth of the maximum `float` value, which I looked up.

(Well, slightly _above_ one one-billionth of that value.)

PR #210, opened 2023-12-12
Add `-pedantic` to `EXTRA_COPTS`

chiphogg:
Our godbolt link is broken!  https://godbolt.org/z/KrvfhP4M3

To fix this robustly, we add the `-pedantic` option to our builds.  We
put it in `EXTRA_COPTS` instead of `BASE_CLANG_COPTS` because otherwise,
the warning doesn't show up when we run this repro command:

```sh
bazel clean && bazel build --config=gcc //au:constant_test
```

This PR fixes all of the other warnings that started showing up when I
added this compiler option, too.  Interestingly, the godbolt-breaking
problem (extra `;`) seems like it shows up only for gcc's `-pedantic`
implementation, not clang's.

### DIFF
--- a/au/constant.hh
+++ b/au/constant.hh
@@ -96,7 +96,7 @@ struct Constant : detail::MakesQuantityFromNumber<Constant, Unit>,
     constexpr operator T() const {
         return as<typename CorrespondingQuantity<T>::Rep>(
             typename CorrespondingQuantity<T>::Unit{});
-    };
+    }
 };
 
 // Make a constant from the given unit.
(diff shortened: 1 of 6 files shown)


chiphogg:
I intentionally delayed fixing the offending error just to double-check that CI would catch it.  It would!  Job link: https://github.com/aurora-opensource/au/actions/runs/7187543317/job/19575326547?pr=210

geoffviola (approved):
Thanks for fixing this. I like the addition of pedantic for better compiler support. It's a shame that designated initializers didn't make the standard until C++20.

PR #211, opened 2023-12-14
Check that every magnitude fits in its target type

chiphogg:
`get_value<T>(m)` converts a magnitude `m` to its representation in type
`T`.  It produces a hard compiler error if the value can't fit.  Its
relative `get_value_result<T>(m)` always returns a result, but produces
an error code if the value can't be represented.

Previously, there was a hole in this system.  We assumed that each
magnitude could be represented in `std::uintmax_t`.  Of course, if it
couldn't, `get_value` would fail to compile as it should.  But
`get_value_result` would _also_ fail to compile, instead of producing an
error condition.

To fix this, we need to move our checking utilities further up in the
file, so that all of our utilities can use them.  `product`, `int_pow`,
`base_power_value` --- they all need to _check their computations_ at
every step.  When we do this, we're able to gracefully detect the
non-representability of huge numbers even for the biggest types.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -267,6 +267,21 @@ struct NumeratorImpl<Magnitude<BPs...>>
 
 namespace detail {
 
+enum class MagRepresentationOutcome {
+    OK,
+    ERR_NON_INTEGER_IN_INTEGER_TYPE,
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
Checks look useful. Tests look good.

PR #212, opened 2023-12-15
Add overflow helpers for applying rational to int

chiphogg:
C++'s integer promotion rules for small types make this specific use
case --- applying a rational magnitude to an integral type --- really
interesting!  The product type can be bigger than the type of the two
inputs, so the overflow might not happen when you think it should.
_Moreover,_ the fact that we subsequently divide by a denominator means
that we can _re-enter_ the valid range and still produce a correct
result!

The only way I know how to deal with something this complicated is to
make a separate target that does only that one thing.  Happily, it's
still the case that there is some maximum and minimum value for each
type that will not overflow for a particular rational magnitude.  The
new helper target exists to figure out what those limits are.

I exposed this limitation by adding a conversion between meters and
yards in `uint16_t`.  This is a great test case because the unit ratio
between meters and yards is very close to 1 (it's 1143 to 1250), meaning
the denominator division "rescues" the great majority of cases.
Moreover, the multiplicative factor (either 1143 or 1250 depending on
direction) is far less than the ratio between `uint16_t` and its
promoted type, assuming the latter is equivalent to `int32_t`.  So,
despite the fact that we're multiplying by a number which is pretty
large relative to the type, we almost never actually overflow!

I fixed up the test case so that if it _does_ fail (say, we add some new
unit conversion that exposes a new logic error), it will be very easy to
understand why.

### DIFF
--- a/au/apply_magnitude.hh
+++ b/au/apply_magnitude.hh
@@ -164,7 +164,8 @@ struct ApplyMagnitudeImpl<Mag, ApplyAs::RATIONAL_MULTIPLY, T, true> {
                   "Mismatched instantiation (should never be done manually)");
 
     constexpr T operator()(const T &x) {
-        return x * get_value<T>(numerator(Mag{})) / get_value<T>(denominator(Mag{}));
+        return static_cast<T>(x * get_value<T>(numerator(Mag{})) /
+                              get_value<T>(denominator(Mag{})));
     }
 
     static constexpr bool would_overflow(const T &x) {
--- a/au/apply_magnitude.hh
+++ b/au/apply_magnitude.hh
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include "au/apply_rational_magnitude_to_integral.hh"
 #include "au/magnitude.hh"
 
 namespace au {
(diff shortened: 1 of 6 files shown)


chiphogg:
By the way: here's what we would get if we only applied the changes to `quantity_test.cc` onto main:

```
[ RUN      ] IsConversionLossy.CorrectlyDiscriminatesBetweenLossyAndLosslessConversions
au/quantity_test.cc:777: Failure
Expected equality of these values:
  is_lossy
    Which is: true
  did_value_change
    Which is: false
Conversion is lossy (overflows), but round-trip conversion did not change the value.  original: 1250 yd, converted: 1143 m, round_trip: 1250 yd
au/quantity_test.cc:777: Failure
Expected equality of these values:
  is_lossy
    Which is: true
  did_value_change
    Which is: false
Conversion is lossy (overflows), but round-trip conversion did not change the value.  original: 2500 yd, converted: 2286 m, round_trip: 2500 yd
au/quantity_test.cc:777: Failure
Expected equality of these values:
  is_lossy
    Which is: true
  did_value_change
    Which is: false

[... many, many rows omitted ...]

au/quantity_test.cc:777: Failure
Expected equality of these values:
  is_lossy
    Which is: true
  did_value_change
    Which is: false
Conversion is lossy (overflows), but round-trip conversion did not change the value.  original: 63750 yd, converted: 58293 m, round_trip: 63750 yd
au/quantity_test.cc:777: Failure
Expected equality of these values:
  is_lossy
    Which is: true
  did_value_change
    Which is: false
Conversion is lossy (overflows), but round-trip conversion did not change the value.  original: 65000 yd, converted: 59436 m, round_trip: 65000 yd
[  FAILED  ] IsConversionLossy.CorrectlyDiscriminatesBetweenLossyAndLosslessConversions (63 ms)
```

So, the current state of main calls certain yards-to-meters conversions lossy, even though the library produces a lossless result.  With the present PR's changes, these tests all pass!

au/apply_rational_magnitude_to_integral_test.cc
```diff
@@ -0,0 +1,376 @@
...
+    validate_spec<T, MagT>(spec);
+    result_out = MaxNonOverflowingValue<T, MagT>::value();
+}
+
+TEST(MaxNonOverflowingValue, AlwaysZeroIfNumCannotFitInPromotedType) {
+    // Case "1" above.
+    constexpr auto huge = pow<400>(mag<10>()) / mag<3>();
+
+    {
+        int max_int = 123;
```
geoffviola, line 128: Q: It looks like the value 123 is always written over. Is there any reason for this value vs 0?

au/apply_magnitude.hh
```diff
@@ -141,13 +164,12 @@ struct ApplyMagnitudeImpl<Mag, ApplyAs::RATIONAL_MULTIPLY, T, true> {
                   "Mismatched instantiation (should never be done manually)");
 
     constexpr T operator()(const T &x) {
-        return x * get_value<T>(numerator(Mag{})) / get_value<T>(denominator(Mag{}));
+        return static_cast<T>(x * get_value<T>(numerator(Mag{})) /
```
geoffviola, line 167: Commend: Nice catch!

geoffviola (approved):
The production code looks better. The test coverage has improved around many tricky cases. Thanks!

au/apply_rational_magnitude_to_integral_test.cc
chiphogg: Nope: I just needed something that is clearly not 0.  :sweat_smile:

au/apply_magnitude.hh
chiphogg: Actually, in chatting about the logic with you, I noticed a bug, and it was here!  Fortunately, it was pretty self-contained, and easy to write a test case for.  Check out 81c4a3bcd1f4da4b9f3d89019d77119228dc3b57 for the fix.

geoffviola: Great! I was thinking about `static_cast<T>`. I guess there's also the tricky bit about `get_value<P>`. This line is trickier than it looks 😅

PR #213, opened 2023-12-18
Support computing values for all rational powers

### LINKED ISSUE #116: Support value conversion with rational exponents

chiphogg:
I thought we would need to wait for a `constexpr` version of
`std::powl`, but I was wrong.  Instead, we can get away with a `root`
function.  Note that there's no guarantee that `std::powl` would be
better in all cases, even if we _could_ just use it.  After all, its
second argument is `1.0l / n`, which is a lossy representation of the
Nth root.

On the implementation side, we take advantage of the fact that this
function is only ever meant to be called at compile time, which lets us
prioritize accuracy over "speed" (because the runtime speed is always
essentially infinite).  We do a binary search between 1 and `x` after
checking that `x > 1` (and `n > 1`).  When we end up with two
neighboring floating point values, we pick whichever gives the closest
result when we run it through `checked_int_pow`.

This algorithm is guaranteed to produce the best representable result
for a "pure root".  For other rational powers, it's technically possible
that we could have some lossiness in the `checked_int_pow` computation,
if we enter a floating point realm that is high enough such that not all
integer values can be represented.  That said, I tried a bunch of test
cases to compare it to `powl`, and it's passed all the ones that I was
able to come up with.

Fixes #116.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -270,7 +270,7 @@ namespace detail {
 enum class MagRepresentationOutcome {
     OK,
     ERR_NON_INTEGER_IN_INTEGER_TYPE,
-    ERR_RATIONAL_POWERS,
+    ERR_INVALID_ROOT,
     ERR_CANNOT_FIT,
 };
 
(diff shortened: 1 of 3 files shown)


au/magnitude.hh
```diff
@@ -316,10 +316,101 @@ constexpr MagRepresentationOrError<T> checked_int_pow(T base, std::uintmax_t exp
...
+        return {MagRepresentationOutcome::OK, static_cast<T>(T{1} / inverse_result.value)};
+    }
+
+    //
+    // At this point, error conditions are finished, and we can proceed with the "core" algorithm.
+    //
+
+    // Always use `long double` for intermediate computations.  We don't ever expect people to be
+    // calling this at runtime, so we want maximum accuracy.
+    long double lo = 1.0;
```
geoffviola, line 369: FYI `long double` will be `double` with MSVC https://learn.microsoft.com/en-us/cpp/c-language/type-long-double?view=msvc-170

chiphogg: Nice catch!

I think that's fine.  By saying `long double`, what we're really saying is "give me the highest precision floating point type".  If MSVC produces slightly less precise results, then that's an acceptable platform level difference.

au/magnitude_test.cc
```diff
@@ -298,6 +314,106 @@ TEST(CheckedIntPow, FindsAppropriateLimits) {
...
+
+TEST(Root, RecoversExactValueWherePossible) {
+    {
+        const auto sqrt_4f = root(4.0f, 2);
+        EXPECT_THAT(sqrt_4f.outcome, Eq(MagRepresentationOutcome::OK));
+        EXPECT_THAT(sqrt_4f.value, SameTypeAndValue(2.0f));
+    }
+
+    {
+        const auto cbrt_125l = root(125.0l, 3);
```
geoffviola, line 361: Not sure if we subscribe to hicpp, but typically they recommend capital `L` over `l` to avoid `l` vs `1` reader error: https://clang.llvm.org/extra/clang-tidy/checks/readability/uppercase-literal-suffix.html

If we are really serious about adhering to that standard, we should put clang-tidy in CI.

au/magnitude_test.cc
```diff
@@ -298,6 +314,106 @@ TEST(CheckedIntPow, FindsAppropriateLimits) {
...
+    EXPECT_THAT(root(0.25, 2), FitsAndProducesValue(0.5));
+    EXPECT_THAT(root(0.0001, 4), FitsAndMatchesValue<double>(DoubleEq(0.1)));
+}
+
+TEST(Root, ResultIsVeryCloseToStdPowForPureRoots) {
+    for (const double x : {55.5, 123.456, 789.012, 3456.789, 12345.6789, 5.67e25}) {
+        for (const auto r : {2u, 3u, 4u, 5u, 6u, 7u, 8u, 9u}) {
+            const auto double_result = root(x, r);
+            EXPECT_THAT(double_result.outcome, Eq(MagRepresentationOutcome::OK));
+            EXPECT_THAT(double_result.value, DoubleEq(static_cast<double>(std::pow(x, 1.0l / r))));
```
geoffviola, line 377: Are the runtime compute times pretty similar?

geoffviola (approved):
Algorithm makes sense. New API looks useful. Tests look good.

au/magnitude.hh
```diff
@@ -316,10 +316,101 @@ constexpr MagRepresentationOrError<T> checked_int_pow(T base, std::uintmax_t exp
...
+        return {MagRepresentationOutcome::ERR_INVALID_ROOT};
+    }
+
+    // The "first root" is trivial.
+    if (n == 1) {
+        return {MagRepresentationOutcome::OK, x};
+    }
+
+    // We only support nontrivial roots of floating point types.
+    if (!std::is_floating_point<T>::value) {
```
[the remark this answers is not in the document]
chiphogg: I initially had it as one, but then I realized that there are some values of `n` and `x` that make sense for integral types.  Granted, they're all trivial.  :grin:  But since we're returning a value-or-error type instead of a simple value, I like being able to communicate this --- especially since it fits so nicely into a pre-existing error category.

I would also worry more about this kind of thing if we intended to expose `root` externally.

au/magnitude_test.cc
chiphogg: That's a great point.  I like this as a future work direction.  In any case, I'll make the changes for this present file right away.

au/magnitude_test.cc
chiphogg: I don't know, because I didn't check.  I assume not --- I assume ours is much slower!  I guess we could always write a benchmark, but I'm not too concerned.

What I did measure was compile time impacts.  I found no measurable slowdown at all.  This isn't surprising, as we only pay the cost if the denominator is nonzero, and we have no existing instances of this whatsoever.  Even once this feature gets out, we expect true fractional exponents to be very rare.  And even though our implementation is _relatively_ inefficient, I still don't expect it to be enough to have a noticeable impact on compile times.

PR #214, opened 2023-12-21
Rename `pip_deps` to `au_pip_deps`

philsc:
I was experimenting with importing au, but ran into a conflict with
rules_docker's `pip_deps` repository. This patch fixes the conflict by
switching to a more unique name.

Another solution is to switch to bzlmod completely, but that's more
work.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


chiphogg (approved):
Nice catch, thanks.

Also, you missed the 0.3.4 release by mere hours.  :sweat_smile:  Oh well!

PR #218, opened 2024-02-05
Correctly stream out quantities with a char Rep

hoffbrinkle:
Quantities that have a `char` `Rep` will have the resulting raw number
streamed as a `char`.  Given that a Quantity is a number, users would
expect that a the value would be streamed as the integral number, not
interpreted as a character code.

### DIFF
--- a/au/io.hh
+++ b/au/io.hh
@@ -25,7 +25,7 @@ namespace au {
 // Streaming output support for Quantity types.
 template <typename U, typename R>
 std::ostream &operator<<(std::ostream &out, const Quantity<U, R> &q) {
-    out << q.in(U{}) << " " << unit_label(U{});
+    out << +q.in(U{}) << " " << unit_label(U{});
     return out;
 }
 
--- a/au/io_test.cc
+++ b/au/io_test.cc
@@ -61,6 +61,8 @@ TEST(StreamingOutput, PrintsValueAndUnitLabel) {
     EXPECT_EQ(stream_to_string((feet / milli(second))(1.25)), "1.25 ft / ms");
 }
 
+TEST(StreamingOutput, PrintValueRepChar) { EXPECT_EQ(stream_to_string(feet(char{3})), "3 ft"); }
+
 TEST(StreamingOutput, DistinguishesPointFromQuantityByAtSign) {
     EXPECT_EQ(stream_to_string(celsius_qty(20)), "20 deg C");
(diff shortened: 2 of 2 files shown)


au/io_test.cc
```diff
@@ -61,6 +61,8 @@ TEST(StreamingOutput, PrintsValueAndUnitLabel) {
     EXPECT_EQ(stream_to_string((feet / milli(second))(1.25)), "1.25 ft / ms");
 }
 
+TEST(StreamingOutput, PrintValueRepChar) { EXPECT_EQ(stream_to_string(feet(char{3})), "3 ft"); }
```
chiphogg, line 64: What if we replace `char` with `int8_t`?

I think that would still be enough to reproduce the bug (https://godbolt.org/z/5cMoWGdWj), but it's probably better if we phrase it in terms of a quirk of the `int8_t` rep, so as not to suggest that people should be using `char` as a rep (which I think we don't want to encourage).

au/io.hh
```diff
@@ -25,7 +25,7 @@ namespace au {
 // Streaming output support for Quantity types.
 template <typename U, typename R>
 std::ostream &operator<<(std::ostream &out, const Quantity<U, R> &q) {
-    out << q.in(U{}) << " " << unit_label(U{});
+    out << +q.in(U{}) << " " << unit_label(U{});
```
chiphogg, line 28: Nice, elegant change!  I think this is a good use of integer promotion here (although I dislike integer promotion in general as a language feature of C and C++).

Since this is pretty non-obvious, how about a comment here?  Something to the effect of, "Unary plus causes integer promotion.  This helps when the rep is `int8_t`, which could otherwise be interpreted as `char` in a streaming context."

chiphogg (approved):
Thanks for finding and fixing the bug, and including a test!

I think you'll need to sign the CLA before I can merge, even though it's somewhat of a formality in your case.  (I know you work for Aurora, but unfortunately our tooling can't currently know that!)

au/io_test.cc
hoffbrinkle: That does introduce a dependency on `<cstdint>`, but probably not a bad idea.  I might be tempted to keep both cases (along with a comment about the `int8_t` resolving to a `char`)?

hoffbrinkle: Added in 71ed45e.

au/io.hh
hoffbrinkle: Done in 3689593.

chiphogg (approved):

PR #223, opened 2024-03-07
Perform QuantityPoint subtraction in target rep

### LINKED ISSUE #222: Incorrect conversion result from Kelvins to Celsius
We define our canonical representation of temperature as U32 Milli-Kelvins:
```cpp
/**
 * Our canonical Temperature unit is milli-kelvin stored in uint32.
 * uint32_t holds the relevant range of temperatures in mK while providing good resolution to our needs.
 */
using Temperature = au::QuantityPointU32<au::Milli<au::Kelvins>>;
```

Interestingly, the following static_assert fails:
```cpp
constexpr auto millikelvins = au::milli(au::kelvins_pt);
constexpr Temperature ABSOLUTE_ZERO = millikelvins(0);

// We expect to lose the `.15` here due to integer being used, but `ABSOLUTE_ZERO` should be `-273C`.
static_assert(ABSOLUTE_ZERO.in<int>(au::celsius_pt) == -273); // Fails to compile:  '4294694 == -273'
```

It doesn't fail if I ditch my predefined `QuantityPoint` and rather just use the maker directly:
```cpp
static_assert(millikelvins(0).in<int>(au::celsius_pt) == -273); // compiles, the world make sense again.
```

So clearly something in forcing the combination of mK and U32 does something here, and I'm not sure what.
It looks like integer underflow, buy would any subtraction take place in this conversion?
In order to go from mK to C all is needed is `(mK+273150) / 1000`

So I guess my question is what is going on, and how can I avoid it considering that I care about the actual representation of temperature since the project is an embedded device, so size matters.

Thanks.

chiphogg:
If we don't do this, then converting from unsigned to signed of the same
size will perform the subtraction in the unsigned type.

Included a unit test that failed before the change, but passes with it.

Fixes #222.

### DIFF
--- a/au/quantity_point.hh
+++ b/au/quantity_point.hh
@@ -132,7 +132,8 @@ class QuantityPoint {
               typename NewUnit,
               typename = std::enable_if_t<IsUnit<NewUnit>::value>>
     constexpr NewRep in(NewUnit u) const {
-        return (x_ - OriginDisplacement<Unit, NewUnit>::value()).template in<NewRep>(u);
+        return (rep_cast<NewRep>(x_) - rep_cast<NewRep>(OriginDisplacement<Unit, NewUnit>::value()))
+            .template in<NewRep>(u);
     }
 
     template <typename NewUnit, typename = std::enable_if_t<IsUnit<NewUnit>::value>>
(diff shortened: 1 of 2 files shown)


chiphogg:
FYI @geoffviola, the builds are failing due to #224.  I'm looking into how to get them to passing.

geoffviola (approved):
Implementation and test look good.

PR #225, opened 2024-03-07
Migrate to new home for LLVM toolchain

### LINKED ISSUE #224: Toolchain moved to new repo name, breaking hashes
See:

- bazel-contrib/toolchains_llvm#253
- bazel-contrib/toolchains_llvm#254

I think this will block us from landing any new code until we update this.  We might also be unable to build old commits forever; I'm not sure.

chiphogg:
The LLVM toolchain graduated from `grailbio/bazel-toolchain` to
`bazel-contrib/toolchains_llvm`.  Good for them!  Unfortunately, that
broke everyone's build who was using a `sha256` checksum, because GitHub
does not guarantee SHA stability for automatically generated artifacts.
(Remember when they [broke the world] in January 2023?)

We need to upgrade to the new repo location, which means changing the
SHA.  We would not like to be forced to do this again, so we'll want to
link to a concrete artifact, not an auto-generated one.  This means
moving from a particular commit hash to a named release, because the
`.tar.gz` file is attached to that release.  So we pick the latest one.

I followed and adapted the [0.10.3 instructions], using the "not
`bzlmod`" section.  I also grepped the repo for `grailbio`, and updated
a URL in some comments.  (I verified that the new URL gives the same
contents as the old one.)

But wait, there's more!  The `aspect-build/gcc-toolchain` repo _also_
migrated to a new home!  Now it lives with a particular individual
called `f0rmiga`, instead of the `aspect-build` bazel organization.  We
can't land code until we unbreak _all_ of the builds, so we need to fix
that in the same PR.  Here, I _don't_ want to migrate to a new release,
because (a) we have a patch on top of the specific commit that we had
been using, and (b) they don't even upload archive files for their
releases.

[broke the world]:
https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/
[0.10.3 instructions]:
https://github.com/bazel-contrib/toolchains_llvm/releases/tag/0.10.3

Fixes #224.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -89,11 +89,11 @@ http_archive(
     patches = [
         "@//:third_party/aspect_gcc_toolchain/0001-Expose-target_settings-and-set-std-c-14.patch",
     ],
-    sha256 = "e2e12202dd83f67d71101b24554044de25e1625d16b4b56bc453ecaa8f7c6bd0",
-    strip_prefix = "aspect-build-gcc-toolchain-ac745d4",
+    sha256 = "9c075a67d401d1aa8b4935cc520e9926b1926fba72c1ab609400b239c92f5639",
+    strip_prefix = "f0rmiga-gcc-toolchain-ac745d4",
     type = "tar.gz",
     urls = [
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
Using the named release and getting the build to pass looks good.

PR #227, opened 2024-03-09
Delete `ForbidsComposingWith` mixin

### LINKED ISSUE #226: Redefinition of 'operator*' and 'operator/' errors
I'm on a Mac using Xcode 11.7. In the build settings I have set Apple Clang C++ Language to C++17. I also tried GNU++17. I'm using the single header installation and have this issue with both a custom single file and the pre-built single file.

The errors I'm getting are `Redefinition of 'operator*'` and `Redefinition of 'operator/'`. It is happening on the four operators in `struct ForbidsComposingWith`. It actually happens twice on the first of the four operators for a total of five errors.

I don't know if I've done something wrong or if this is a compatibility issue.

chiphogg:
It causes problems with Apple Clang version 9, apparently.  I think this
compiler technically falls outside of our support window.  However, an
actual user is having issues with it, and I think `ForbidsComposingWith`
is adding only very marginal value anyway.  I tried deleting it and the
error messages became a couple of lines shorter, although perhaps less
direct (no more reference to "explicitly deleted").

Amusingly, I'm keeping the unit test for it around, because it's one of
those unfortunate "uncomment to test" deals, and the tests _will_ still
pass.  (We'll get compiler errors for doing unsupported operations.)

Fixes #226.

### DIFF
--- a/au/constant.hh
+++ b/au/constant.hh
@@ -40,8 +40,6 @@ struct Constant : detail::MakesQuantityFromNumber<Constant, Unit>,
                   detail::ComposesWith<Constant, Unit, Constant, Constant>,
                   detail::ComposesWith<Constant, Unit, QuantityMaker, QuantityMaker>,
                   detail::ComposesWith<Constant, Unit, SingularNameFor, SingularNameFor>,
-                  detail::ForbidsComposingWith<Constant, Unit, QuantityPointMaker>,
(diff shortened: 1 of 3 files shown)


geoffviola (approved):
This workaround seems reasonable.

It also looks like a compiler bug to me. I'm surprised to see this bug in a compiler released in 2019. If we are more confident that it's the compiler's fault and users can upgrade, we can add this code back in.

PR #229, opened 2024-03-30
Permit arithmetic operations with more types

### LINKED ISSUE #228: Add Multiplication Support for Complex Types
Thank you very much for sharing your code.

The multiplication operator in Quantity is only enabled if T suffices std::is_arithmetic.
However, my use case involves the use of complex numbers, e.g. std::complex.
What about allowing such types as well?

As a quick work-around I adapted the code as follows:

        template<typename T, typename = void>
	struct has_real_and_imag : std::false_type
	{
	};
	template<typename T>
	struct has_real_and_imag<T, std::void_t<decltype(std::declval<T>().real()), decltype(std::declval<T>().imag())>> : std::true_type
	{
	};
	template<typename T>
	struct is_complex : has_real_and_imag<T>
	{
	};

	// Scalar and Complex multiplication.
	template<typename T, typename = std::enable_if_t < std::is_arithmetic<T>::value || is_complex<T>::value >>
	friend constexpr auto operator*(Quantity a, T s)
	{
		return make_quantity<UnitT>(a.value_ * s);
	}

chiphogg:
The goal of these `std::is_arithmetic` SFINAE uses was to avoid
ambiguous overloads when we multiply a `Quantity` with another
`Quantity`.  It was only ever a shortcut.  It seems to work well in most
cases, but recently, a user pointed out that it doesn't help
`std::complex`.  (To be clear, we could _form_ a `Quantity` with
`std::complex` rep, but we couldn't multiply or divide a `Quantity` with
a `std::complex` scalar.)

This PR takes a different approach.

First, we define what constitutes a valid "rep".  We're very permissive
here, using a "deny-list" approach that filters out empty types, known
Au types, and other libraries' units types (which we detect via our
`CorrespondingQuantity` trait).  Incidentally, this should give us a
nice head start on #52.

Next, we permit an operation (multiplication or division) depending on
how the "candidate scalar" type would interact with the Quantity's rep:
simply put, the result must itself both exist, and be a valid rep.

The net result should be that we automatically consider a much wider
variety of types --- including complex number types from outside the
standard library --- to be valid scalars.  Crucially, this should also
be able to avoid breaking existing code: we are introducing potentially
a _lot_ of new overloads for these operators, and we can't allow any to
create an ambiguity with existing overloads.

Helps #228, but we need more Magnitude support.

### DIFF
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -395,6 +395,7 @@ cc_library(
         ":apply_magnitude",
         ":conversion_policy",
         ":operators",
+        ":rep",
         ":unit_of_measure",
         ":zero",
     ],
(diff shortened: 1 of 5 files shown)


geoffviola (approved):
Implementations and tests look good. Thanks!

PR #234, opened 2024-05-02
Explicitly specify `QuantityPoint` calc type

### LINKED ISSUE #233: QuantityPoint: Value is forced into target Rep before unit conversion
Consider the following:
```cpp
using Temperature = au::QuantityPointU32<au::Milli<au::Kelvins>>;
constexpr auto millikelvins = au::milli(au::kelvins_pt);
```
Given these definition, one would expect that:

```cpp
constexpr Temperature temp = millikelvins(313150); // 40C
constexpr auto temp_dK = temp.in<uint16_t>(au::deci(au::kelvins_pt));
static_assert(temp_dK == 3131); // FAIL. Actual value: 510.
```

So it looks like first `temp` is forced into `uint16_t`, truncating the value to `51006`, then the unit is converted from mK to dK, resulting in `510` which is equivalent to `-222.15C`

chiphogg:
We want to carry out our intermediate computations in the common type of
the old and new rep, generally.  However, we already know from #222 that
if the new rep is a signed integer, and the old rep is unsigned, we need
to be careful to use a signed type for intermediate computations.

To preserve this behaviour but restore correct behavoiur for other cases
(namely, #233), we add a new type trait for this intermediate type.  It
basically boils down to the common type, but with an explicit carve-out
to keep #222 fixed.  We explicitly cast both participants to this new
type-for-intermediate-computations, and then cast to the destination rep
at the end.

Compile time tests showed no discernible impact at all.  This change has
also passed all Aurora-internal builds.

Fixes #233.

### DIFF
--- a/au/quantity_test.cc
+++ b/au/quantity_test.cc
@@ -253,6 +253,12 @@ TEST(Quantity, CoerceInExplicitRepSetsOutputType) {
     EXPECT_THAT(feet(30).coerce_in<uint8_t>(inches), SameTypeAndValue(uint8_t{104}));
 }
 
+TEST(Quantity, CoerceAsPerformsConversionInWidestType) {
+    constexpr QuantityU32<Milli<Meters>> length = milli(meters)(313'150u);
+    EXPECT_THAT(length.coerce_as<uint16_t>(deci(meters)),
+                SameTypeAndValue(deci(meters)(uint16_t{3131})));
+}
+
 TEST(Quantity, CanImplicitlyConvertToDifferentUnitOfSameDimension) {
     constexpr QuantityI32<Inches> x = yards(2);
     EXPECT_EQ(x.in(inches), 72);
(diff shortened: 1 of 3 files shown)


au/quantity_test.cc
```diff
@@ -253,6 +253,12 @@ TEST(Quantity, CoerceInExplicitRepSetsOutputType) {
     EXPECT_THAT(feet(30).coerce_in<uint8_t>(inches), SameTypeAndValue(uint8_t{104}));
 }
 
+TEST(Quantity, CoerceAsPerformsConversionInWidestType) {
+    constexpr QuantityU32<Milli<Meters>> length = milli(meters)(313'150u);
+    EXPECT_THAT(length.coerce_as<uint16_t>(deci(meters)),
+                SameTypeAndValue(deci(meters)(uint16_t{3131})));
```
geoffviola, line 259: Is truncation or rounding expected?

geoffviola (approved):
This case is tricky. Changes and tests look good!

au/quantity_test.cc
chiphogg: `coerce_as` will truncate.

PR #238, opened 2024-05-22
Refresh python requirements and add instructions

chiphogg:
It's been a while since we updated `requirements_lock.txt`, and we've
accumulated four Dependabot alerts (7, 8, 9, and 10).  I took the
opportunity to write down all the steps we should take when doing this
update.  Naturally, I also performed all of those steps.

For the A/B website comparison, I found 3 differences:

- Menu alignment has been tweaked on a few pages to reduce word wrapping
- Tab headers are underlined
- Mermaid diagram arrowheads are thicker

For testing `mike`, the suggested workflow actually caught some breaking
changes for our monkey patch code.  I therefore made the first step to
simply directly check the patched APIs.

The new version of mkdocs now warns about anchors, and it caught an
incorrect anchor in a recent doc update.  I've fixed that as well.


geoffviola (approved):
I ran through the instructions. The docs still look good. Thanks for the updates!

Issue #246, opened 2024-06-19
naming conflict in Arduino causes compilation errors


pdaoust:
Hi, excited to use your library, but I'm hitting compilation errors when using Arduino. Currently using:

* `au_noio.hh` 0.3.4-11-g221d8f4
* esp-idf 6.5.0
* framework-arduinoespressif32 2.0.14
* GCC with `-std=gnu++2a` build flag

all within PlatformIO.

There are two conflicts between Arduino and AU:

* Arduino [defines a macro `B1`](https://forum.arduino.cc/t/b1-reserved/403083) (super useful 🙄 ) and AU uses `B1` as a template param in various places.
* Arduino defines a macro `PI` and AU defines a static constexpr `PI`.

```
In file included from /home/paul-ed/.platformio/packages/framework-arduinoespressif32/cores/esp32/Arduino.h:41,
                 from src/main.cpp:1:
/home/paul-ed/.platformio/packages/framework-arduinoespressif32/cores/esp32/binary.h:31:12: error: expected identifier before numeric constant
 #define B1 1
            ^
src/types/au_noio.hpp:345:17: note: in expansion of macro 'B1'
 template <class B1>
                 ^~
/home/paul-ed/.platformio/packages/framework-arduinoespressif32/cores/esp32/binary.h:31:12: error: expected '>' before numeric constant
 #define B1 1
            ^
src/types/au_noio.hpp:345:17: note: in expansion of macro 'B1'
 template <class B1>
                 ^~
In file included from src/sources/arduinoMillisClock.hpp:6,
                 from src/sources/everything.hpp:2,
                 from src/rheoscape_everything.hpp:4,
                 from src/main.cpp:2:
src/types/au_noio.hpp:346:22: error: type/value mismatch at argument 1 in template parameter list for 'template<class ...> struct au::stdx::conjunction'
 struct conjunction<B1> : B1 {};
                      ^
src/types/au_noio.hpp:346:22: note:   expected a type, got '1'
In file included from /home/paul-ed/.platformio/packages/framework-arduinoespressif32/cores/esp32/Arduino.h:41,
                 from src/main.cpp:1:
/home/paul-ed/.platformio/packages/framework-arduinoespressif32/cores/esp32/binary.h:31:12: error: expected class-name before numeric constant
 #define B1 1
            ^
src/types/au_noio.hpp:346:26: note: in expansion of macro 'B1'
 struct conjunction<B1> : B1 {};
                          ^~
[...]

PR #247, opened 2024-06-20
Avoid colliding with macro names `B1` and `PI`

chiphogg:
This helps Arduino compatibility.

Of course, this whole situation is an excellent case in point to
illustrate why, if you _must_ use macros, _never_ to give them such
easily collidable names.  But that is not the users' fault, so since we
can work around the problem, we do so.

The `B1` template parameter can be renamed to `B` without loss of
readability.

`PI` is a little more challenging.  We do use it in one place in our
code, in defining `Degrees`, so we must replace that with its definition
of `Magnitude<Pi>{}`.  We provide workaround instructions in the
comments.  We leave alone the many instances of `PI` in test files,
since that shouldn't affect these users.

Helps #246.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -131,7 +131,16 @@ using CommonMagnitudeT = typename CommonMagnitude<Ms...>::type;
 // Value based interface for Magnitude.
 
 static constexpr auto ONE = Magnitude<>{};
+
+#ifndef PI
+// Some users must work with frameworks that define `PI` as a macro.  Having a macro with this
+// easily collidable name is exceedingly unwise.  Nevertheless, that's not the users' fault, so we
+// accommodate those frameworks by omitting the definition of `PI` in this case.
+//
+// If you are stuck with such a framework, you can choose a different name that does not collide,
+// and reproduce the following line in your own system.
 static constexpr auto PI = Magnitude<Pi>{};
+#endif
 
 template <typename... BP1s, typename... BP2s>
 constexpr auto operator*(Magnitude<BP1s...>, Magnitude<BP2s...>) {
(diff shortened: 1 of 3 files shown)


au/magnitude.hh
```diff
@@ -131,7 +131,16 @@ using CommonMagnitudeT = typename CommonMagnitude<Ms...>::type;
...
 static constexpr auto ONE = Magnitude<>{};
+
+#ifndef PI
+// Some users must work with frameworks that define `PI` as a macro.  Having a macro with this
+// easily collidable name is exceedingly unwise.  Nevertheless, that's not the users' fault, so we
+// accommodate those frameworks by omitting the definition of `PI` in this case.
+//
+// If you are stuck with such a framework, you can choose a different name that does not collide,
+// and reproduce the following line in your own system.
 static constexpr auto PI = Magnitude<Pi>{};
```
geoffviola, line 142: I'm not sure if this is a perfect solution. It would depend on the order of includes. The order of headers is not great to reason about.

I don't have great ideas for alternative solutions 😓 
1. Use Google style guide for naming constants `kBlah`. It generally won't collide with macros, but it is ugly.
2. Rename `PI` to something like `AUPI`. It would be unique, but it's more verbose and farther from the simple Greek letter. Also, that is what namespaces are for 😄 
3. Define PI in another header. This would require more fine grained header management.

geoffviola (approved):

au/magnitude.hh
chiphogg: Having slept on this, I'm leaning towards Option 4: just don't define `PI`.  Using really short all-caps names is a recipe for collision with poorly designed frameworks.  That's not to say we'll _never_ do it, but it seems wise to minimize how much we do it.  And in this case, we have a clear collision with the Arduino framework, which we'd really like to support, so I think our `PI` has to go.

I'll deprecate it in a follow on PR for ease of review.  I think we'll still want to retain the `#ifndef` guards until we remove it entirely (probably in 0.4.0).

PR #251, opened 2024-06-23
Support `constexpr` in `min` and `max`

### LINKED ISSUE #244: min and max functions are not constexpr
C++'s [std::max](https://en.cppreference.com/w/cpp/algorithm/max) and [std::min](https://en.cppreference.com/w/cpp/algorithm/min) have been `constexpr` since C++14.  It would be fantastic to have the Au implementations of `max` and `min` similarly `constexpr` so they could be used similar to those in the standard library.

chiphogg:
These have been `constexpr` compatible since C++14.  Mechanically, the
way I did this was:

1. Find an overload that was not `constexpr`.
2. Make a test that requires it to be `constexpr` to pass (either
   tweaking an existing test, or making a new one).
3. Make sure the test fails.
4. Change the function until the test passes.

The biggest challenge was that our default compiler wouldn't accept our
lambda in a `constexpr` context.  That feels like a bug, but in any
case, we were able to work around it by making a manual function object
with an explicitly `constexpr` call operator.

Fixes #244.

### DIFF
--- a/au/math.hh
+++ b/au/math.hh
@@ -318,17 +318,27 @@ constexpr bool isnan(QuantityPoint<U, R> p) {
     return std::isnan(p.in(U{}));
 }
 
+namespace detail {
+// Some compilers do not support lambdas in constexpr contexts, so we make a manual function object.
+struct StdMaxByValue {
+    template <typename T>
+    constexpr auto operator()(T a, T b) const {
+        return std::max(a, b);
+    }
+};
+}  // namespace detail
+
 // The maximum of two values of the same dimension.
 //
 // Unlike std::max, returns by value rather than by reference, because the types might differ.
 template <typename U1, typename U2, typename R1, typename R2>
-auto max(Quantity<U1, R1> q1, Quantity<U2, R2> q2) {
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
The changes make sense. It looks like lambdas in a `constexpr` context have been added in C++17: https://stackoverflow.com/a/32697323

chiphogg:
Well... that would sure explain why it didn't work before!  TIL.

au/math.hh
```diff
@@ -318,31 +318,41 @@ constexpr bool isnan(QuantityPoint<U, R> p) {
     return std::isnan(p.in(U{}));
 }
 
+namespace detail {
+// Some compilers do not support lambdas in constexpr contexts, so we make a manual function object.
```
hoffbrinkle, line 322: Should we update the comment about this being a C++17 thing, or is it the case that even with C++17, some compilers still don't support it.

hoffbrinkle (commented):
Sweet, thanks!

au/math.hh
chiphogg: Good point!  Addressed in #252.

PR #254, opened 2024-07-02
Add foundations for CMake support

chiphogg:
We start by creating the top-level `CMakeLists.txt` file, following the
modern CMake guide (https://cliutils.gitlab.io/modern-cmake/).

We already have a `build/` directory, so we'll use `cmake/build` for our
build folder.  We add this to `.gitignore` (using a `*` so as to support
multiple build folders).

We update the release instructions so as to keep the version number in
sync, and the installation instructions so as to explain how to use
CMake.

Helps #215.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,30 @@
+# Copyright 2024 Aurora Operations, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# CMake should always be at least as new as your compiler[1].  Clang 14 was released on March 25,
+# 2022, and is our newest officially supported compiler.  The next CMake release was v3.23.0, on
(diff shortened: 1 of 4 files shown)


chiphogg:
@geoffviola, see [review comment from @Cazadorro](https://github.com/aurora-opensource/au/issues/215#issuecomment-2203543323).

CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+# CMake should always be at least as new as your compiler[1].  Clang 14 was released on March 25,
+# 2022, and is our newest officially supported compiler.  The next CMake release was v3.23.0, on
+# March 29, 2022.  Therefore, our minimum version must be at least CMake 3.23.
+#
+# The maximum version should be the latest version we've tested the project with.
+#
+# [1]: https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html
+cmake_minimum_required(VERSION 3.23...3.29)
```
geoffviola, line 22: Any reason that you chose cmake 3.23? Ubuntu 22.04.4 LTS packages cmake 3.22.1 😄

chiphogg: Yep, the comment above explains the reasoning.

CMake is in some ways counter-intuitive, at least to me.  There's a strong expectation that it will be very easy for users to get the latest version, and thus there are community norms around aggressively updating the minimum version.  I will say, I was pleased to see how easy it was to grab 3.29 on Ubuntu, simply by adding the official PPA.

geoffviola: Sounds good. I followed this [SO post](https://askubuntu.com/questions/355565/how-do-i-install-the-latest-version-of-cmake-from-the-command-line) and the code below worked.

geoffviola (approved):
Top-level code is documented and works with cmake 3.29.6.

PR #259, opened 2024-07-10
Add basic CI job that runs CMake builds

chiphogg:
For now, we simply run "the default CMake generator on the latest
Ubuntu".  Later on, if we tighten up our notion of "supported CMake
configurations", we can add more options here.

Helps #215.


chiphogg:
FYI @timhirsh, we'll need a security review on jwlawson/actions-setup-cmake@802fa1a2c4e212495c05bf94dba2704a92a472be.  Assuming it passes, we can add it to the allowlist in the repo settings, and after that, the job should run remotely.  (I already ran it locally using [nektos/act](https://github.com/nektos/act), so I know it works.)

chiphogg:
Alternatively, perhaps we should consider batching the security reviews (see #260).  Our version of `checkout`, for example, is pretty old.

.github/workflows/cmake-build-and-test.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  cmake-build-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@dc323e67f16fb5f7663d20ff7941f27f5809e9b6 #v2.6.0
```
timhirsh, line 28: See #261 for context:
```suggestion
      - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b #v4.1.5
```

timhirsh (approved):
This LGTM, pending completion of the security review 👍

PR #262, opened 2024-07-10
Make `header_only_library()` function

chiphogg:
Au is overwhelmingly built from header-only libraries.  We can make this
a lot easier to maintain with a CMake function, `header_only_library()`,
that presents an Au-centric interface.  Specifically, the function will
automatically:

- Make the target an `INTERFACE`
- Set `FILE_SET HEADERS`
- Set `BASE_DIRS` to the base Au directory

Users must provide the `HEADERS` option, and we will use those to set
the `FILES` option.  Users can optionall provide the `DEPS` option, and
we'll use it if present to populate a `target_link_libraries` call.

We also add an optional `GTEST_SRCS` argument.  If this is provided,
then we will automatically create a `_test`-suffixed target, link to the
library and to `gtest_main`, and run `gtest_discover_tests()`.  If the
test has any dependencies apart from the tested target, and GTest, they
can be supplied via an optional `GTEST_EXTRA_DEPS` argument.

Finally, we provide the option to mark a library as `INTERNAL_ONLY`. For
now, we borrow the `EXPORT_NAME` manipulation from the original CMake
branch, but we could choose to do other things later --- the important
part is marking the intent that the library is internal-only.

Helps #215.

### DIFF
--- a/cmake/HeaderOnlyLibrary.cmake
+++ b/cmake/HeaderOnlyLibrary.cmake
@@ -1,0 +1,75 @@
+# Copyright 2024 Aurora Operations, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+function(header_only_library)
+  #
+  # Handle argument parsing
+  #
+
+  set(prefix ARG)
+  set(noValueVars INTERNAL_ONLY)
+  set(singleValueVars NAME)
+  set(multiValueVars
+    HEADERS
+    DEPS
+    GTEST_SRCS
+    GTEST_EXTRA_DEPS
+  )
+
+  cmake_parse_arguments(
+    PARSE_ARGV 0
+    "${prefix}"
+    "${noValueVars}"
+    "${singleValueVars}"
+    "${multiValueVars}"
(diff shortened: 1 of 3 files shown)


cmake/HeaderOnlyLibrary.cmake
```diff
@@ -0,0 +1,75 @@
...
+  if (DEFINED ARG_GTEST_SRCS)
+    add_executable("${ARG_NAME}_test")
+    target_sources("${ARG_NAME}_test" PRIVATE ${ARG_GTEST_SRCS})
+    target_link_libraries(
+      "${ARG_NAME}_test"
+      PRIVATE
+      ${ARG_NAME}
+      ${ARG_GTEST_EXTRA_DEPS}
+      GTest::gtest_main
+      GTest::gmock_main
```
geoffviola, line 71: Both main targets would have two `main` functions. You should only need `GTest::gmock_main`. 

`gmock_main` has main and [depends on gmock](https://github.com/google/googletest/blob/b4aaf97d8f7eaffab79aa15e10a91b331b941fe2/googlemock/CMakeLists.txt#L101). [gmock depends on gtest](https://github.com/google/googletest/blob/b4aaf97d8f7eaffab79aa15e10a91b331b941fe2/googlemock/CMakeLists.txt#L98).

geoffviola (approved):
Changes make sense. Test still passes.

```bash
docker run --rm \
/bin/bash -c \
'apt update && apt install -y clang cmake git && git clone -b chiphogg/cmake-header-only-library#215 https://github.com/aurora-opensource/au.git && cd au && cmake -S . -B cmake/build -DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE && cmake --build cmake/build --target all all_verify_interface_header_sets test'

cmake/HeaderOnlyLibrary.cmake
chiphogg: Nice catch!  Now I'm a little curious why it actually worked before (or at least, appeared to), with two `main` functions.

PR #264, opened 2024-07-11
Require C++14 support for Au targets

chiphogg:
Helps #215.


geoffviola (approved):
This makes sense and should help downstream users.

The tests are run with gcc 11's default std version which is C++17. That make sense given that C++14 is the minimum requirement.

PR #268, opened 2024-07-22
Properly declare googletest dependency

chiphogg:
This lets us support installing the library to the system, and then
using it via `find_package()`.

To test this, I ran `sudo cmake --install cmake/build`, observed that it
completed with no errors, and then made a new CMake package that simply
used `find_package(Au)`.  Everything worked!

Helps #215.


geoffviola (approved):
Makes sense. Supporting 1find_package1 is great!

PR #269, opened 2024-07-22
Add `/W4` to Windows CI builds

chiphogg:
This should help catch a wider variety of warnings.

### DIFF
--- a/.github/workflows/single-file-build-and-test.yml
+++ b/.github/workflows/single-file-build-and-test.yml
@@ -43,5 +43,5 @@ jobs:
       - name: Build and run test
         shell: cmd
         run: |
-          cl.exe /std:c++14 single-file-test.cc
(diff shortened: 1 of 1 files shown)


.github/workflows/single-file-build-and-test.yml
```diff
@@ -43,5 +43,5 @@ jobs:
       - name: Build and run test
         shell: cmd
         run: |
-          cl.exe /std:c++14 single-file-test.cc
+          cl.exe /std:c++14 /W4 single-file-test.cc
```
geoffviola, line 46: I was thinking of adding the warnings in CMake and building and running through that. We would probably want to add `/WX` to treat the [warnings as errors](https://learn.microsoft.com/en-us/cpp/build/reference/wx-treat-linker-warnings-as-errors?view=msvc-170).

PR #271, opened 2024-07-23
Prep for 0.3.5 release

chiphogg:
I updated the CMake version number to 0.3.5.  Note that the URL doesn't
exist yet.  This is expected (see `RELEASE.md`); we'll make it exist as
part of the release process.

Separately, in the docs, I fixed up our bazel installation instructions,
because they were referring to a very old release.  The major change
here was to point users towards the manually uploaded archive file,
rather than the automatically generated one, because the former is
stable but the latter is not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,9 +26,9 @@ cmake_minimum_required(VERSION 3.24...3.29)
 
 project(
    Au
-   VERSION 0.3.4
+   VERSION 0.3.5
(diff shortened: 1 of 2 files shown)


chiphogg:
I did a compile time comparison last night on Aurora's AV repo.  The results came back squeaky clean: no measurable regressions at all.  (We weren't expecting any in this release.)

geoffviola (approved):

PR #275, opened 2024-07-30
Permit integer-dividing quantity-equivalent units

### LINKED ISSUE #249: Permit integer division between quantities of quantity-equivalent units
Integer division for quantity types has scary pitfalls (see: [slides](https://chogg.name/cppcon-2021-units/#/39); [video](https://youtu.be/5dhFtSu3wCo?si=GRzZA4_yiVRQ2xoG&t=2186)).

Initially, we always prevented integer division when quantities were involved.

We later realized that we had forgotten to handle divisions involving one quantity and one raw number, so we tried tightening the policy. 
 After surveying a bunch of use cases, we found a lot of places where the denominator was a raw integer number.  And these actually seemed fine!  So we refined our policy to only prevent cases where the _denominator_ has units.

And now we have #248, and its code example suggests yet another carve-out.  If we integer-divide two quantities, but their units are quantity-equivalent, then this should be completely fine.  The integer-division problems for quantities come when we can independently adjust the units for one quantity, such that the actual numeric variable we're dividing by can become arbitrarily different.  _If the units are constrained to be equivalent_, then this can't happen, and we're unlikely to violate users' intuition.  But if we continue to prevent this, we'll also prevent legitimate use cases like the example in #248!

So let's add this carve-out to the policy.

chiphogg:
This means we need to know the unit inside of
`warn_if_integer_division()`, which seems OK.  For the raw number cases,
we synthesize a unitless unit.

One weird thing is that sometimes the template parameters for
`warn_if_integer_division()` are for the numerator, and sometimes for
the denominator.  But it still gives the correct answer in all cases. If
both are `Quantity` types, then it doesn't matter which is which,
because "both integers and quantity-inequivalent units" is symmetric.
And if one is a raw number, then `warn_if_integer_division()` is _not_
symmetric, but that's fine because we simply don't _call it_ from the
`Q/ N` variant (which is always fine); we only call it from the `N / Q`
variant (which might not be fine).

Fixes #249.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -310,7 +310,7 @@ class Quantity {
     }
     template <typename T, typename = std::enable_if_t<IsQuotientValidRep<T, RepT>::value>>
     friend constexpr auto operator/(T s, Quantity a) {
-        warn_if_integer_division<T>();
+        warn_if_integer_division<UnitProductT<>, T>();
         return make_quantity<decltype(pow<-1>(unit))>(s / a.value_);
     }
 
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
Changes make sense. Since the units are quantity equivalent, there shouldn't be any additional conversion on the underlying values.

PR #276, opened 2024-08-02
Tweak corresponding quantity for main chrono types

### LINKED ISSUE #216: Use prefixed seconds when possible for chrono-equivalent units
See this link for current behaviour:

https://godbolt.org/z/Kxe8vrh48

We could create a utility template that says, effectively, "if any of these prefixes produces a quantity-equivalent unit, then use that instead of just the unit-and-mag".

Of course, once #85 gets fixed, we won't see `[UNLABELED_UNIT]`; we'll see something like `[1 / 1000000000] s`.  But the prefixes would be a nicer user experience.

chiphogg:
Currently, we have a single generic correspondence for all chrono
durations.  This is perfectly unit safe, but it means we have no labels
for any duration type, including common types like `std::chrono::hours`.
This means that `as_quantity(std::chrono::hours{3})` would print as
`"3 [UNLABELED UNIT]"` instead of `"3 h"`.

This PR defines a special correspondence for each commonly used chrono
duration type, which fixes the above behaviour.

As part of doing this, we found a missing dependency on `:quantity`, so
we added it in the BUILD file.

Fixes #216.


au/code/au/chrono_interop_test.cc
```diff
@@ -36,6 +46,36 @@ TEST(DurationQuantity, InterconvertsWithIndirectlyEquivalentChronoDuration) {
     EXPECT_THAT(from_chrono, SameTypeAndValue(seconds(1.234)));
 }
 
+TEST(DurationQuantity, EquivalentOfChronoNanosecondsHasNsLabel) {
+    constexpr auto from_chrono_ns = as_quantity(std::chrono::nanoseconds{123});
+    EXPECT_EQ(stream_to_string(from_chrono_ns), "123 ns");
```
hoffbrinkle, line 51: Any reason not to use
```suggestion
    EXPECT_STREQ(stream_to_string(from_chrono_ns), "123 ns");
```
or
```suggestion
    EXPECT_THAT(stream_to_string(from_chrono_ns), StrEq("123 ns"));
```
?

hoffbrinkle (approved):

au/code/au/chrono_interop_test.cc
chiphogg: No special reason; I'll make the switch.  Thanks!

geoffviola (approved):
Looks useful. Tests look good.

Issue #260, opened 2024-07-10
Update GitHub actions versions


chiphogg:
We're getting a bunch of errors (e.g., because `checkout` uses node12, which is deprecated).  We should consider batching the security reviews.

timhirsh:
Good call.  We already completed a review for a newer version, so #261 should bump us to node20.

timhirsh:
Remaining deprecation warnings:
- `ilammy/msvc-dev-cmd@cec98b9d092141f74527d0afa6feb2af698cfe89` ([run](https://github.com/aurora-opensource/au/actions/runs/9877004774?pr=261))
- `viperproject/check-license-header@971caaf7d1641bb2c6a49acd06b018a5ce51ad6e` ([run](https://github.com/aurora-opensource/au/actions/runs/9877004778?pr=261))
- `zgosalvez/github-actions-ensure-sha-pinned-actions@af2eb3226618e2494e3d9084f515ad6dcf16e229` ([run](https://github.com/aurora-opensource/au/actions/runs/9877004773?pr=261))

I'll kick off a review for these as well.

chiphogg:
> Remaining deprecation warnings:
> 
> * `ilammy/msvc-dev-cmd@cec98b9d092141f74527d0afa6feb2af698cfe89` ([run](https://github.com/aurora-opensource/au/actions/runs/9877004774?pr=261))
> * `viperproject/check-license-header@971caaf7d1641bb2c6a49acd06b018a5ce51ad6e` ([run](https://github.com/aurora-opensource/au/actions/runs/9877004778?pr=261))
> * `zgosalvez/github-actions-ensure-sha-pinned-actions@af2eb3226618e2494e3d9084f515ad6dcf16e229` ([run](https://github.com/aurora-opensource/au/actions/runs/9877004773?pr=261))
> 
> I'll kick off a review for these as well.

Thanks @timhirsh !  Any updates on the review for these?

timhirsh:
Yup, PR's incoming ☝️ 

We also need `actions/setup-python@3542bca2639a428e1796aaa6a2ffef0c0f575566`, so that'll be the last one.

PR #282, opened 2024-08-05
[ci] Bump actions/setup-python to v5.1.1

timhirsh:
### Summary

See #260 for background.

Update the [`actions/setup-python`](https://github.com/actions/setup-python) action to the most recent version: [v5.1.1](https://github.com/actions/setup-python/releases/tag/v5.1.1).  Quick summary of major versions:
* [v4](https://github.com/actions/setup-python/releases/tag/v4.0.0): there is no longer a default `python-version`, and we already specify one, so we should be ok
* [v5](https://github.com/actions/setup-python/releases/tag/v5.0.0) contained the Node20 upgrade

### Test Plan
- [ ] Verify new hash on this PR
- [ ] Checked job summary and saw no deprecation warnings related to `actions/setup-python`
- [X] Checked the changelog and didn't see any backwards incompatible changes

### DIFF
--- a/.github/workflows/single-file-build-and-test.yml
+++ b/.github/workflows/single-file-build-and-test.yml
@@ -32,7 +32,7 @@ jobs:
         uses: ilammy/msvc-dev-cmd@cec98b9d092141f74527d0afa6feb2af698cfe89 #v1.12.1
 
       - name: Set up python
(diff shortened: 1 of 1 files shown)


chiphogg (approved):
Double checked hash from [release](https://github.com/actions/setup-python/releases/tag/v5.1.1).

It's pretty rare that Au feels the need for a submit queue, but this would have been one of those times.  :slightly_smiling_face:

PR #286, opened 2024-08-12
Change compiler naming scheme for MSVC

### LINKED ISSUE #283: MSVC 19.35 CI isn't fixed-version; updates to latest
The latest runs of the MSVC 19.35 workflow are using 19.40. It is probably good to keep testing with latest MSVC but the CI should be relabeled. Could call it MSVC 2022 because it's on the windows-2022 runner.

- https://github.com/aurora-opensource/au/actions/runs/10061250249/job/27810846466#step:6:55
- https://github.com/aurora-opensource/au/actions/workflows/msvc-x64-19-35-32217-1.yml

The MSVC 19.29 workflow is staying on that compiler version properly, but I assume this is because the windows-2019 runner is no longer getting major compiler version updates.
- https://github.com/aurora-opensource/au/actions/runs/10255370911/job/28372173649#step:6:58
- https://github.com/aurora-opensource/au/blob/main/.github/workflows/msvc-x64-19-29-30151.yml

chiphogg:
We've been calling the version on Windows Server 2022, "MSVC x64 19.35".
Well, apparently it's now running 19.40, not 19.35.  If this can change
without our knowing, then putting the version number in the name is
probably not a great idea.

An alternative suggestion was to use the version of Windows Server in
the name, since that won't ever change.  We do that here.

This will break the continuity for the page giving historical results
per job, i.e., it will create a "new job".  I'm not especially concerned
about that.

Fixes #283.

### DIFF
--- a/.github/workflows/msvc-2019-x64.yml
+++ b/.github/workflows/msvc-2019-x64.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: "MSVC x64 19.29"
+name: "MSVC 2019 x64"
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


timhirsh (approved):
👍

PR #291, opened 2024-09-14
Automatically include <iostream> in godbolt link

### LINKED ISSUE #290: Include iostream in canonical godbolt link
I have to add it manually every single time and it's just so annoying.

chiphogg:
This eliminates the main annoyance in my personal use of the godbolt
link, and I suspect many others' too.

Fixes #290.


geoffviola (approved):
I think it makes sense to optimize your experience. I typically use `cout` as well.

There may be some `std::fputs` users that prefer a smaller binary size. If there are issues, we can revisit that later.

chiphogg:
Yep, I don't think it matters too much; all we need is a reasonable default.

Issue #292, opened 2024-09-16
clang-format job is broken


chiphogg:
See:

- DoozyX/clang-format-lint-action#73
- DoozyX/clang-format-lint-action#75
- DoozyX/clang-format-lint-action#78

For now, we will just disable this job.  I think this is a fine intermediate solution, because I am the one making most or all of the PRs, and I run clang-format automatically.

Once we decide which release we want to use, we can kick off an Aurora security review on the diff with the current version.

PR #293, opened 2024-09-16
Temporarily disable clang-format lint

chiphogg:
We'll need to get a known-good release through security review, and then
we can re-enable it.  Meanwhile, #292 will track this issue and remind
us that we'll need to do this.

### DIFF
--- a/.github/workflows/clang-format-lint.yml
+++ b/.github/workflows/clang-format-lint.yml
@@ -14,11 +14,11 @@
 
 name: clang-format-lint
 
-on:
-  push:
-    branches:
-      - main
-  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/clang-format-lint.yml
```diff
@@ -14,11 +14,11 @@
 
 name: clang-format-lint
 
-on:
-  push:
-    branches:
-      - main
-  pull_request:
+# on:
```
timhirsh, line 17:
```suggestion
# Temporarily disabled. See #292
# on:
```

timhirsh (approved):

PR #296, opened 2024-09-27
Default construct rep inside quantity

### LINKED ISSUE #273: Default constructor and default value
Hi,
I just found a very shocking news in the documentation regarding default-constructor:

> A default-constructed Quantity is initialized to some value, which helps avoid certain kinds of memory safety bugs. However, the value is contractually unspecified. You can of course look up that value by reading the source code, but we may change it in the future, and we would not consider this to be a breaking change. The only valid operation on a default-constructed Quantity is to assign to it later on.

This behaviour would differ from the underlying native type like `double`, `int` etc. which is contra-intuitive and thus error-prone! 
We are using the default-constructor spread in our entire code base, with the assumption, that the default value equals to the default value of the native underlying type!
You cannot expect that users read every single line in your documentation. And even if I know this now, my collegues in my company still don't know this and it is hard to communicate this problem to every developer in a big company.
Documentating it, doesn't solve the problem, as I said: You must never expect, that user are reading carefully every single line of your documentation. I experience myself, that many collegues simply don't read my documentation and they expect, that everything works like they expect it intuitively.

chiphogg:
This won't change any actual values when the rep is a fundamental
arithmetic type, but it is more stylistically consistent with the idea
of a "default constructor".

We also update our contract to officially promise that we do this.  This
lets us get rid of the scary/threatening language that "we might change
this out from under you!", even though we never had any intention to
actually do so.  We retain the encouragement to use `au::ZERO`, and to
avoid using a default-constructed `Quantity` for anything other than
future assignment.

Fixes #273.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -403,7 +403,7 @@ class Quantity {
 
     constexpr Quantity(Rep value) : value_{value} {}
 
-    Rep value_{0};
+    Rep value_{};
 };
 
 // Force integer division beteween two integer Quantities, in a callsite-obvious way.
--- a/docs/reference/quantity.md
+++ b/docs/reference/quantity.md
@@ -133,11 +133,17 @@ class Quantity {
 };
 ```
 
-A default-constructed `Quantity` is initialized to some value, which helps avoid certain kinds of
-memory safety bugs.  However, **the value is contractually unspecified**.  You can of course look up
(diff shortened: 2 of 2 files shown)


geoffviola (approved):
Makes sense. This feature is a departure from initializing fundamental types in C++. But it seems like the right trade-off in terms of safety and clarity.

PR #297, opened 2024-09-27
Make building unit tests optional to simplify packaging

ericriff:
I intent to package this library with Conan.

I don't want the conan package to depend on GTest, so instead of patching it out on the conan recipe I decided to propose this here instead.

I maintained the tests enabled by default, but a user can opt out of them with `-DENABLE_TESTING:BOOL=OFF`.

I don't think that gtest has any business on the installed targets. I disabled the `find_package()` call on the generated config file if we're building unit tests, but there are other references to gtest on the generated files. That's unusual. I think GTest should be treated as an internal tool, used for testing, and do not get propagated downstream.

-Eric

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,28 +32,35 @@ project(
    LANGUAGES CXX
 )
 
+option(ENABLE_TESTING "Build Unit Tests" ON)
+
 # The export set for all of our headers.
 set(AU_EXPORT_SET_NAME AuHeaders)
 
-enable_testing()
+if(ENABLE_TESTING)
+   message(STATUS "Unit tests enabled")
+   enable_testing()
 
-# Bring in GoogleTest so we can build and run the tests.
-include(FetchContent)
-FetchContent_Declare(
-   googletest
-   GIT_REPOSITORY https://github.com/google/googletest.git
-   GIT_TAG 58d77fa8070e8cec2dc1ed015d66b454c8d78850 # Release 1.12.1
-   FIND_PACKAGE_ARGS
-      1.12.1
-      NAMES GTest
-)
+   # Bring in GoogleTest so we can build and run the tests.
+   include(FetchContent)
+   FetchContent_Declare(
+      googletest
+      GIT_REPOSITORY https://github.com/google/googletest.git
+      GIT_TAG 58d77fa8070e8cec2dc1ed015d66b454c8d78850 # Release 1.12.1
+      FIND_PACKAGE_ARGS
+         1.12.1
+         NAMES GTest
+   )
 
-# https://google.github.io/googletest/quickstart-cmake.html
-# For Windows: Prevent overriding the parent project's compiler/linker settings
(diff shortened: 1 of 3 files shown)


chiphogg:
Thanks for making this PR!  Conan support has been a longstanding wishlist item of mine, and it's really exciting to see folks from the community stepping up to help make it happen.  :rocket: 

As I mentioned in the conversation for #215, we do have one target that legitimately depends on GTest.  Not sure how we want to handle that here.  As I _also_ mentioned there, since we don't use CMake/conan on a day to day basis, there's a high prior probability of us just making weird choices out of ignorance.  :slightly_smiling_face: 

LMK your thoughts on some of our best options.  I hope I'll get a chance to look more closely at this either later today or sometime in the next couple days.

CMakeLists.txt
```diff
@@ -32,28 +32,37 @@ project(
    LANGUAGES CXX
 )
 
+option(ENABLE_TESTING "Build Unit Tests" ON)
```
chiphogg, line 35: Nit: let's add the `AU_` prefix here.  Note that I did this in ericriff/au#1, which you're welcome to merge into this PR according to the instructions in the summary. :slightly_smiling_face:

cmake/HeaderOnlyLibrary.cmake
```diff
@@ -75,7 +75,7 @@ function(header_only_library)
   )
 
   # Add the test, if requested.
-  if (DEFINED ARG_GTEST_SRCS)
+  if (DEFINED ARG_GTEST_SRCS AND ENABLE_TESTING)
```
chiphogg, line 78: This line has a merge conflict with the new setup on `main`, because we broke this function apart into two separate functions: one for libraries, one for tests.  Accordingly, there's no such thing as `ARG_GTEST_SRCS` anymore.

ericriff/au#1 resolves this merge conflict, by simply adding an early return when this variable is not true.

chiphogg (approved):
This looks good!  I think this PR makes sense as the "add the first of two options" PR.

There is a merge conflict, because we've already landed the project CMake restructuring PR (#300).  In order to make it as easy as possible to land this present PR, I made my own PR _into your PR's branch_: ericriff/au#1.  My PR both resolves the merge conflict, and addresses my sole review comment.  The instructions in the summary indicate how to use it.  I **believe** that if you merge ericriff/au#1, we'll be able to land #297 basically as-is (although I admit I've never tried this strategy before).

After that, I'll put up a PR to add the second CMake option, which will let users opt out of all CMake dependencies.  And then, I believe Au's CMake setup should be able to support a conan recipe, without any patches!

chiphogg:
Superseded by #303, which includes all of your commits here (and I've confirmed that your authorship shows up in the commit log in main).

Thanks a ton for your contributions @ericriff!  It's because of your efforts (and those of @AbrilRBS) that Au is now finally available on conan.  :sunglasses:

Issue #257, opened 2024-07-04
Add Au support for conan


AbrilRBS:
Hi!
I'm trying to package the library to make it available in Conan Center Index, and I noticed that the pre-built header-only versions provided are only for the main branch, and are not present in the releases

I'd love to give users the ability to use this file for the version they choose (instead of building the package), but as of right now the release tar does not contain the files, so I'd like to request that in future releases, those files are added as part of the release at that point, thanks! :)

chiphogg:
Thanks for writing!  I've wanted to see Au available in Conan for a while.  I thought we would need to support CMake first --- an effort which is actively underway (#215).  But it sounds like you know an alternative way that doesn't require this, which is intriguing.  (I don't know one way or another, because I don't use Conan myself.)

Note that we do have single-file versions going all the way back to 0.3.1, because our documentation website hosts docs simultaneously for different releases.  For example, the single-file releases for 0.3.2 can be found at <https://aurora-opensource.github.io/au/0.3.2/install/#pre-built-single-file>.

That said --- I'm inferring from your conan use case that it wouldn't be enough to simply be able to download the files.  It sounds like you would need to have the single-file versions present inside of the release tarball.  Is that accurate?  I don't know how this works.  (Incidentally, if you have any Conan docs that would help me understand how I would need to package the single-file releases, that would be very helpful!)

By the way, one more question: would this whole issue go away if Au supported CMake?  I'm assuming it would be easier to make a "full install" available in Conan, so the demand for the single-file alternative would be reduced.

AbrilRBS:
Hi @chiphogg sorry for the delay, busy week :)

> I thought we would need to support CMake first

Conan has integrations with many build systems, and Bazel is one of those! We have less experience packaging those than CMake ones, but Conan is perfectly capable of using your already-existing build system to generate packages! (But yes, CMake is usually a bit easier to work with)

> I'm inferring from your conan use case that it wouldn't be enough to simply be able to download the files. It sounds like you would need to have the single-file versions present inside of the release tarball. Is that accurate?

[...]

PR #300, opened 2024-10-04
Consolidate Au CMake targets

chiphogg:
This breaks the 1:1 correspondence between target and test, so we split
up our helper accordingly into two.

We make `au` the single mega-target that holds everything that doesn't
depend on GTest.  It turns out that there are two targets that do.
`testing` is for external users, and contains custom GMock matchers.
`chrono_policy_validation` is an internal-only library that makes it
easy to test that Au is like chrono where we want to be, and unlike
chrono otherwise.  Each of these targets also depends on the `au`
mega-target.

Helps #257.

### DIFF
--- a/au/code/au/CMakeLists.txt
+++ b/au/code/au/CMakeLists.txt
@@ -20,7 +20,87 @@ include("${PROJECT_SOURCE_DIR}/cmake/HeaderOnlyLibrary.cmake")
 
 header_only_library(
   NAME au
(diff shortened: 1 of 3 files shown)


chiphogg:
Note to the reviewer:

This PR uses "checkpoint commits" to make it easier to verify correctness.  Currently, there are 4.  Start at the [first commit](https://github.com/aurora-opensource/au/pull/300/commits/97ba5e41ad496c50f871b5335f99b3342163bcfa), and use `n` and `p` to go to next and previous.  Here is what the commits do:

1. Bulk-update rules:
    1. Instead of `header_only_library` doing both the header and the test, we split it into two rules (by adding the new `gtest_based_test`), because tests and libraries aren't 1:1 anymore.
    1. We now have only three `header_only_library` instances: `au` (public, main); `testing` (public, optional); `chrono_policy_validation` (private).
    3. There are two aspects of this commit that make for easier diffing: (a) `GTEST_SRCS` is **not** renamed to `SRCS` yet, and (b) all of the `HEADERS` in `au` are on one single line, so that git isn't tempted to match them to their old rules.
    1. We get rid of `GTEST_SRCS` and `GTEST_EXTRA_DEPS` from the header-only libraries, because again, tests aren't 1:1 with library targets anymore.
    1. Every **old** `header_only_library` instance becomes a `gtest_based_test`
    1. `_test` gets appended to the name
    1. `HEADERS` goes away
    1. Old `DEPS` goes away
    2. New `DEPS`: delete any targets that went away, and add `au` (the new "mono-target") as a dep basically every time
2. Do the rest of the change.
    1. The `HEADERS` go one per line
    1. `GTEST_SRCS` gets renamed to just `SRCS`
3. Fix the breakage in the previous commit.
    1. Look.  I ran all the tests before pushing.  But CMake was so sloppy that it just reused the previously built executables!  It's good that we're providing CMake support, but this just reinforces what a good thing it is that we're bazel-first.
4. Update docs!

geoffviola (approved):
Looks good. I can see that the dependency on gmock has moved.

The following test still looks good.
```console
docker run --rm \
/bin/bash -c \
'apt update && apt install -y clang cmake git && git clone -b chiphogg/consolidate-cmake-libs#257 https://github.com/aurora-opensource/au.git && cd au && cmake -S . -B cmake/build -DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE && cmake --build cmake/build -j $(proc) --target all all_verify_interface_header_sets && ctest -j $(nproc) --test-dir cmake/build'
```

PR #303, opened 2024-10-20
Gate CMake tests behind `AU_ENABLE_TESTING` option

chiphogg:
This brings in the external contributions from #297, merges in the
latest main, and prepends an `AU_` prefix to the option.

A future PR will add another option to let users opt out of the GTest
dependency entirely, losing access to the `Au::testing` target (which
wouldn't be any loss to people who don't use GTest).

To test:

```sh
cmake \
  -S . \
  -B cmake/build \
  -DAU_ENABLE_TESTING=FALSE \
  -DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE \
&& cmake \
  --build cmake/build \
  --target all
```

Helps #257.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,12 +32,12 @@ project(
    LANGUAGES CXX
 )
 
-option(ENABLE_TESTING "Build Unit Tests" ON)
+option(AU_ENABLE_TESTING "Build Unit Tests" ON)
 
 # The export set for all of our headers.
 set(AU_EXPORT_SET_NAME AuHeaders)
 
-if(ENABLE_TESTING)
+if(AU_ENABLE_TESTING)
    message(STATUS "Unit tests enabled")
    enable_testing()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,29 +40,29 @@ set(AU_EXPORT_SET_NAME AuHeaders)
 if(AU_ENABLE_TESTING)
    message(STATUS "Unit tests enabled")
    enable_testing()
-
-   # Bring in GoogleTest so we can build and run the tests.
-   include(FetchContent)
-   FetchContent_Declare(
-      googletest
-      GIT_REPOSITORY https://github.com/google/googletest.git
-      GIT_TAG 58d77fa8070e8cec2dc1ed015d66b454c8d78850 # Release 1.12.1
-      FIND_PACKAGE_ARGS
-         1.12.1
-         NAMES GTest
-   )
-
-   # https://google.github.io/googletest/quickstart-cmake.html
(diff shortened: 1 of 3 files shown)


geoffviola (changes requested):
It looks like targets in au/code/au/CMakeLists.txt like `chrono_policy_validation` still depend on `GTest::gtest`. Is that expected?

I see an error when I run
```
docker run --rm \
/bin/bash -c \
'apt update && apt install -y clang cmake git && git clone -b chiphogg/update_vars#257 https://github.com/aurora-opensource/au.git && cd au && cmake -S . -B cmake/build -DAU_ENABLE_TESTING=FALSE -DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE && cmake --build cmake/build -j $(proc) --target all'
```

chiphogg:
> It looks like targets in au/code/au/CMakeLists.txt like `chrono_policy_validation` still depend on `GTest::gtest`. Is that expected?
> 
> I see an error when I run
> 
> ```
> docker run --rm \
> ubuntu@sha256:2e863c44b718727c860746568e1d54afd13b2fa71b160f5cd9058fc436217b30 \
> /bin/bash -c \
> 'apt update && apt install -y clang cmake git && git clone -b chiphogg/update_vars#257 https://github.com/aurora-opensource/au.git && cd au && cmake -S . -B cmake/build -DAU_ENABLE_TESTING=FALSE -DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE && cmake --build cmake/build -j $(proc) --target all'
> ```

Nice catch!

The current [plan of record](https://github.com/aurora-opensource/au/issues/257#issuecomment-2408686674) includes making two options: one for running the tests, and one for including the gtest dependency at all.  In the end, if the latter is false, then we will force the former to be false as well.

This present PR just adds the first option.  The main reason it exists is so that we can land @ericriff's contributions to main, and give him credit.  But his PR predates the fuller understanding reflected in our plan of record: it conflates the two options, and assumes that users skipping the tests would also want to remove the GTest dependency.  I fixed it up in the latest commit.

Of course, this reveals a hole in our CI testing: we didn't add tests for this new option.  I generally prefer to do this in the same PR, but editing `.github` will add another reviewer to this PR, and I think it makes more sense to do this holistically once both options have been added.  I filed #304 so we won't lose track of this.

chiphogg:
(See the updated PR summary for a testing command.)

geoffviola (approved):
LGTM.

Also, we probably don't need `-DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE` when we run cmake with `-DAU_ENABLE_TESTING=FALSE`.

PR #305, opened 2024-10-21
Use official vcpkg link

chiphogg:
[Evidently], `vcpkg.link` is an unofficial third party URL.  We now use
the official URL instead.

[Evidently]: https://github.com/aurora-opensource/au/commit/06b074cd972234fca5b9abe429a7299e9c95af93#r148182262


chiphogg:
CC: @andre-nguyen

PR #309, opened 2024-10-25
Document reasons for omitting raw number ctor

### LINKED ISSUE #274: Missing documentation about missing native based constructors
Hi,
I would be able to explain my collegues, why something like `au::QuantityD<au::Seconds> a{1};` is not allowed like in nholtshaus.
My guess is, that you can always alias a type like `using Timeout = au::QuantityD<au::Milli<au::Seconds>>;` and then do false assumptions about the magnitude (i.e., seconds vs. milliseconds).
But maybe the reason is much deeper and I would like to read this in the documentation.

chiphogg:
Tested by rendering locally with `au-docs-serve`.

Fixes #274.


geoffviola (approved):
Thanks! These are good examples and references.

Issue #105, opened 2023-02-15
Indicate relative magnitudes in common unit labels


chiphogg:
A core part of Au's philosophy is "no preferred units".  So when we subtract, say, MPH from KPH, the result is labeled as essentially "the common unit of MPH and KPH": `COM[km / h, mi / h]`.

I think this is a mixed bag.  It's good that we don't conjure up any unmentioned units.  `COM[km / h, mi / h]` _does_ unambiguously identify what unit is being used.  But this isn't a commonly used unit, and it's clearly bad that the label doesn't give any clue as to its _magnitude_, relative to any other unit.

I've found a new candidate solution.  We can express the common unit in terms of _each_ constituent unit --- and, for clarity, replace the `,` with `==`.  For example: `COM([1 / 15625] km / h == [1 / 25146] mi / h)`.

- **PRO:** Doesn't put any unit on special footing.
- **PRO:** Clearly indicates the common unit's magnitude in terms of _each_ constituent unit.
- **CON:** A little verbose.

I think this is a clear improvement.  Common-unit quantities generally only occur as the result of intermediate computations.  They're not the kind of thing that end users are going to _want_ to print, unmodified.  But when they _do_ end up being output, they'd better be unambiguous!

This solution depends on #85, because we need to be able to print magnitudes.

chiphogg:
Related: mpusz/units#438.

chiphogg:
Note: in discussions relating to the standard units project, we ended up converging on an `EQUIV` syntax:

```
Old:
COM([1 / 15625] km / h == [1 / 25146] mi / h)

New:
EQUIV{[1 / 15625] km / h, [1 / 25146] mi / h}
```

This was the brainchild of @JohelEGP.  I like it a lot!

Actually, we want to include the unscaled unit label inside of the square brackets, so it will be more like this:

```
EQUIV{[1 / 15625 km / h], [1 / 25146 mi / h]}
```

chiphogg:
> Actually, we want to include the unscaled unit label inside of the square brackets, so it will be more like this:
> 
> ```
> EQUIV{[1 / 15625 km / h], [1 / 25146 mi / h]}
> ```

For even better clarity, if the magnitude label has a `/`, we should enclose it in parentheses:

```
EQUIV{[(1 / 15625) km / h], [(1 / 25146) mi / h]}
```

PR #310, opened 2024-10-26
Eliminate redundant units from `CommonUnit<...>`

### LINKED ISSUE #295: Try reducing new units when incorporating into common units
See [this example](https://godbolt.org/z/1GPnq4v6j):

```cpp
int main(int, char**) {
    std::cout << "1) " << (1 * cm + 1 * mi + 1 * m) << std::endl;
    std::cout << "2) " << (1 * cm + 1 * m + 1 * mi) << std::endl;
}
```

Output:

```
1) 805177 COM[m, cm, mi]
2) 805177 COM[cm, mi]
```

In (2), the first addition does not create a common unit, because `cm` divides `m`.  In (1), by contrast, we start with `COM[cm, mi]`, and then incorporate `m`.  It would be nice if we could check `m` against each of `cm` and `mi` in turn, and simply avoid adding it if it's already taken care of.

We could go even further, and produce `COM[cm, mi]` when adding `COM[m, mi]` and `cm`, effectively replacing `m` with `cm`.

chiphogg:
The core here is the addition to `:quantity_test`.  Making this pass is
the goal of this PR.

To get there, we need a new trait/metafunction that can take a pack of
units, and eliminate all the "redundant" ones.  This means we also need
a definition of "redundant", where the canonical example would be that
"feet" is redundant between "feet" and "inches".  Here's what we came up
with:

- If two units are _identical_ (same type), then each is redundant
  w.r.t. the other.
- If two units are _distinct_, but _quantity-equivalent_, then the
  redundant one is whichever comes later in the "standard pack
  ordering".  (We try to give more "recognizable" units higher
  priority.)
- Otherwise, for _quantity-inequivalent_ units, a redundant unit is an
  exact integer multiple of another unit.

All of this assumes that the units have the same dimension.  We simply
wouldn't call this machinery in the first place if that weren't true.

Finally, we need one more piece of machinery: a simple "drop all X"
metafunction.  This is because when an earlier unit makes later units
redundant, we can't easily "reach into" that "tail pack" of units and
remove them directly.  Instead, we need to mark them with a "tombstone"
(`void`), and then remove all the tombstoned units.

Fixes #295.

### DIFF
--- a/au/code/au/quantity_test.cc
+++ b/au/code/au/quantity_test.cc
@@ -22,6 +22,8 @@
 #include "gtest/gtest.h"
 
 using ::testing::DoubleEq;
+using ::testing::Each;
+using ::testing::Eq;
 using ::testing::StaticAssertTypeEq;
 
 namespace au {
@@ -29,7 +31,10 @@ namespace au {
 struct Feet : UnitImpl<Length> {};
 constexpr auto feet = QuantityMaker<Feet>{};
 
-struct Miles : decltype(Feet{} * mag<5'280>()) {};
+struct Miles : decltype(Feet{} * mag<5'280>()) {
+    static constexpr const char label[] = "mi";
+};
+constexpr const char Miles::label[];
 constexpr auto mile = SingularNameFor<Miles>{};
 constexpr auto miles = QuantityMaker<Miles>{};
 
@@ -715,6 +720,32 @@ TEST(Quantity, MixedTypeAdditionUsesCommonRepType) {
 
 TEST(Quantity, MixedTypeSubtractionUsesCommonRepType) {
     EXPECT_THAT(feet(2.f) - feet(1.5), QuantityEquivalent(feet(0.5)));
+}
+
+TEST(Quantity, CommonUnitAlwaysCompletelyIndependentOfOrder) {
+    auto check_units = [](auto unit_a, auto unit_b, auto unit_c) {
+        const auto a = unit_a(1LL);
+        const auto b = unit_b(1LL);
+        const auto c = unit_c(1LL);
+        auto stream_to_string = [](auto x) {
+            std::ostringstream oss;
+            oss << x;
+            return oss.str();
+        };
+        std::vector<std::string> results = {
+            stream_to_string(a + b + c),
+            stream_to_string(a + c + b),
+            stream_to_string(b + a + c),
+            stream_to_string(b + c + a),
+            stream_to_string(c + a + b),
+            stream_to_string(c + b + a),
+        };
+        EXPECT_THAT(results, Each(Eq(results[0])))
+            << "Inconsistency found for (" << a << ", " << b << ", " << c << ")";
+    };
+
+    check_units(centi(meters), miles, meters);
+    check_units(kilo(meters), miles, milli(meters));
 }
 
 TEST(Quantity, CommonTypeRespectsImplicitRepSafetyChecks) {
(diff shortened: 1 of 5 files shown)


au/code/au/quantity_test.cc
```diff
@@ -717,6 +722,32 @@ TEST(Quantity, MixedTypeSubtractionUsesCommonRepType) {
...
+            stream_to_string(c + a + b),
+            stream_to_string(c + b + a),
+        };
+        EXPECT_THAT(results, Each(Eq(results[0])))
+            << "Inconsistency found for (" << a << ", " << b << ", " << c << ")";
+    };
+
+    check_units(centi(meters), miles, meters);
+    check_units(kilo(meters), miles, milli(meters));
+}
```
chiphogg, line 749: Note to reviewer: if we added only this test case on main, without the fixes, here is what the failure would look like:

```
[ RUN      ] Quantity.CommonUnitAlwaysCompletelyIndependentOfOrder
au/code/au/quantity_test.cc:743: Failure
Value of: results
Expected: only contains elements that is equal to "805177 COM[cm, m, mi]"
  Actual: { "805177 COM[cm, m, mi]", "805177 COM[cm, mi]", "805177 COM[cm, m, mi]", "805177 COM[cm, m, mi]", "805177 COM[cm, mi]", "805177 COM[cm, m, mi]" }, whose element #1 doesn't match
Inconsistency found for (1 cm, 1 mi, 1 m)
```

chiphogg:
Note: compile time impact tests came back squeaky clean.  Aurora-internal AV repo tests are still running but look very promising so far and I expect them to pass as well.

au/code/au/utility/type_traits.hh
```diff
@@ -39,11 +44,29 @@ struct AlwaysFalse : std::false_type {};
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Implementation details below.
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// `Prepend` implementation.
```
geoffviola, line 50: What is the point of this comment? Do we need a comment for all inline implementations?

au/code/au/quantity_test.cc
geoffviola: Instead of checking that they are all consistent with the first element, can we check that they are all equal to a specific unit? That might help test readibility.

geoffviola (approved):
Great change! This should make common units more understandable.

au/code/au/quantity_test.cc
chiphogg: I see the appeal of this suggestion.  I struggled with this for a little bit, but I couldn't see my way to an implementation that did this while keeping the test implementation clean.

One way would be to compare to a desired unit label.  But the unit label for common units will change dramatically after 0.4.0 (#105), and I'd hate to have to update this test for that change.

We could also do it by providing the expected _type_ of the common unit.  But `CommonUnit<...>` isn't supposed to be named directly by end users: we're supposed to use `CommonUnitT<...>`, which performs the de-duping and simplification that is the very thing under test!

Ultimately, I think the current approach is the best option available.  Phrasing it in terms of _order independence_, as mentioned in https://github.com/aurora-opensource/au/issues/295#issuecomment-2374768684, is the most elegant way I know to express what we're really after here.  The downside, as you point out, is that a test failure can't tell you _which one is correct_; all it can do is point out that they should all be the same, and they're not.  But I think in practice this will be good enough to point maintainers in the right direction.

au/code/au/utility/type_traits.hh
chiphogg: Often, the implementation for "public"-ish interface declared up top in a file will depend on new helpers, which I will declare-and-define right next to the "core" implementation.  This can get a little messy and confusing, so I use these "section comments" to group things and clarify the purpose.  That said, I'm certainly rather inconsistent about using them in practice.

PR #311, opened 2024-10-26
Add CMake option to fully exclude gtest dependency

### LINKED ISSUE #257: Add Au support for conan

chiphogg:
If users activate this new option, `AU_EXCLUDE_GTEST_DEPENDENCY`, then
it will force `AU_ENABLE_TESTING` to `OFF`.  We will not attempt to pull
in the googletest dependency by any means, and we won't define the extra
targets that depend on it.

To test, run:

```sh
rm -Rf cmake/build/ \
&& cmake \
  -S . \
  -B cmake/build \
  -DAU_EXCLUDE_GTEST_DEPENDENCY=TRUE \
  -DCMAKE_VERIFY_INTERFACE_HEADER_SETS=TRUE \
&& cmake --build cmake/build --target all
```

This provides the `-- Unit tests disabled` message (verifying that the
`AU_ENABLE_TESTING` interaction is working), and it finishes very
quickly without building any tests, or any targets that depend on GTest.

Fixes #257.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ include(CMakeDependentOption)
 
 option(
    AU_EXCLUDE_GTEST_DEPENDENCY
-   "Avoid taking any dependency on googletest (and, implicitly, do not build tests)"
+   "Avoid taking any dependency on googletest. This option implies that tests will be skipped"
    OFF
 )
 cmake_dependent_option(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,20 @@ project(
    LANGUAGES CXX
 )
 
-option(AU_ENABLE_TESTING "Build Unit Tests" ON)
+include(CMakeDependentOption)
+
+option(
+   AU_EXCLUDE_GTEST_DEPENDENCY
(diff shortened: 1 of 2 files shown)


chiphogg:
@AbrilRBS, I believe that after this change lands, we will have everything we need to be able to remove the patch file that was introduced in conan-io/conan-center-index#25447.  Of course, we can't _actually_ remove the patch file until we cut our next release, [0.4.0](https://github.com/aurora-opensource/au/milestone/6), and we're not _super_ close to that.  But if it's easy for you to check that yes, we _will be able_ to remove it once we cut that release, that'd be great!

AbrilRBS:
@chiphogg sorry for the late replay, I've been on holidays these last few days. I can confirm that this would be enough on Conan's side for the next release. Great work! :)

Issue #232, opened 2024-04-03
Provide forward declarations


chiphogg:
It would be nice to provide authoritatively correct forward declarations.  Then users could write APIs in their `.hh` files that take, say, `Quantity` by const-ref, without forcing their dependents to pay the cost of including the library.  Granted, we try to keep that cost as small as possible, but it's not nothing!

chiphogg:
We need to at least try to get a solution for composing units in forward declarations.

PR #312, opened 2024-10-27
Authoritatively forward declare core types

chiphogg:
From now on, anyone can just include `"au/fwd.hh"`, and get
authoritatively correct forward declarations.  These will be implicitly
tested automatically by adding an `#include` of this new file at the top
of every file that has a definition for which we now provide a forward
declaration: if we get a forward declaration wrong, we'll get a compiler
warning.

We also include `fwd.hh` in a few places where it lets us get rid of
some manual forward declarations.

The rep-named aliases move upstream into `fwd.hh`, because they'll be
directly useful for forward declaration use cases, and because you can't
"forward declare" an alias: you just have to move the definition
upstream.

Helps #232.  To resolve it completely, we will need to provide a forward
declaration file for each individual unit file.

### DIFF
--- a/au/code/au/zero.hh
+++ b/au/code/au/zero.hh
@@ -17,6 +17,8 @@
 #include <chrono>
 #include <type_traits>
 
+#include "au/fwd.hh"
+
 namespace au {
 
 // A type representing a quantity of "zero" in any units.
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -41,6 +41,13 @@ cc_test(
 )
 
 cc_library(
+    name = "fwd",
+    hdrs = ["code/au/fwd.hh"],
+    includes = ["code"],
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
     name = "io",
     hdrs = ["code/au/io.hh"],
     includes = ["code"],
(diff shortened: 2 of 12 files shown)


au/code/au/zero.hh
```diff
@@ -17,6 +17,8 @@
 #include <chrono>
 #include <type_traits>
 
+#include "au/fwd.hh"
```
geoffviola, line 20: Is this needed?

geoffviola (approved):
Looks useful 😄

au/code/au/zero.hh
chiphogg: I tried to include the forward declaration file in every file whose types are forward declared.  This should give us better test coverage for cases where the forward declaration goes out of sync with the true definition, such as if we forward declare as `class Zero;` but define as `struct Zero {...};`.

That said, I tried creating exactly this error condition, and I wasn't able to get a warning to come up from `//au:zero_test` alone.  I _do_ get errors from some other test in `//...:all`, but I get these regardless of whether this file includes the `fwd.hh` file or not.

All in all: I think it's a good best practice, and I'm going to keep it for that reason, although I'm mystified about why I can't seem to trigger the warnings as easily as I would expect.

PR #313, opened 2024-10-28
Test options in CMake CI jobs

### LINKED ISSUE #304: Add CMake CI tests for different options
In #303, we added a new option to skip tests, but had no automated CI builds/tests to make sure the option worked.  We need to enhance our CMake CI job to run with this configuration.

chiphogg:
Currently, we only run one CMake configuration in CI.  It's the default
one, where we permit the GTest dependency, and we build and run tests.
Now we are adding two more configurations, one for each of the new
options:

- Set `AU_ENABLE_TESTING` to `FALSE`
- Set `AU_EXCLUDE_GTEST_DEPENDENCY` to `TRUE` (implies the former)

In both these instances, the `test` target won't exist, so we exclude it
from running.

We do however retain the `CMAKE_VERIFY_INTERFACE_HEADER_SETS` option,
because that gives us useful testing for our interface targets.

Fixes #304.

### DIFF
--- a/.github/workflows/cmake-build-and-test.yml
+++ b/.github/workflows/cmake-build-and-test.yml
@@ -26,8 +26,8 @@ jobs:
       matrix:
         options: [
           { extra_cmake_args: '', extra_targets: 'test' },
-          { extra_cmake_args: ' -DAU_ENABLE_TESTING=FALSE', extra_targets: 'test' },
-          { extra_cmake_args: ' -DAU_EXCLUDE_GTEST_DEPENDENCY=TRUE', extra_targets: 'test' }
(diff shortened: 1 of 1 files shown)


chiphogg:
@timhirsh, see [this failed build](https://github.com/aurora-opensource/au/actions/runs/11557439050/job/32167461645).  Note that evidently, when one matrix build fails, it cancels the others.  This seems OK, since they all need to be passing all the time.

timhirsh (approved):
👍

chiphogg:
Now instead of getting 11 notifications from the GitHub app, every time I make, update, or land a PR... I can get 13!  :laughing:

PR #314, opened 2024-10-28
Add forward declarations for individual units

### LINKED ISSUE #232: Provide forward declarations

chiphogg:
Individual `"au/units/*.hh"` files still pull in some decently heavy
machinery.  The sensible, scalable solution is to provide a
`"au/units/*_fwd.hh"` file corresponding to each of them.  Each
"regular" units header will include its `_fwd.hh` header as the first
include (and we add a comment to that effect to prevent clang-format
from reordering them).

Happily, this enables us to write a somewhat meaningful test for the
forward declaraction machinery.  We can now make a header file that
provides an interface with only the forward declarations.  Then we call
that API inside of a test file.

Finally, this is worth updating the docs for:

- **Installation:** mention the `fwd.hh` files.
- **Alternatives:** mention this feature, as we're the first units
  library I know of that provides it.

All of these changes are in separate commits for the convenience of the
reviewer: I recommend starting at the first, and using `n` and `p` to
navigate them.

Fixes #232.

### DIFF
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -47,6 +47,23 @@ cc_library(
     visibility = ["//visibility:public"],
 )
 
+cc_test(
+    name = "fwd_test",
+    size = "small",
+    srcs = [
+        "code/au/fwd_test.cc",
+        "code/au/fwd_test_lib.cc",
+        "code/au/fwd_test_lib.hh",
+    ],
+    deps = [
+        ":fwd",
+        ":io",
+        ":quantity",
+        ":units",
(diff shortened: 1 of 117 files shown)


geoffviola (approved):
Makes sense. Implementation, test, and documentation changes look good.

PR #315, opened 2024-10-29
Replace `integer_quotient` with `unblock_int_div`

### LINKED ISSUE #253: Provide more flexible alternative to integer_quotient()
The point of `integer_quotient()` is to make it extremely obvious that we are about to do a possibly-truncating integer division.  We have seen that this is very error prone when the denominator has units, _and_ when those units are not quantity-equivalent to the numerator's units.  So it's good that we've put up this guard rail.

However, `integer_quotient()` is very inflexible.  For one thing, it's horrible for generic programming: there's no real way to write a division operation that sometimes uses integers (and you're OK with the truncation), but sometimes uses floating point.  For another, it _looks_ completely different to the division operation.

I think we can get all of the safety benefits of `integer_quotient()` by introducing a wrapper type for the denominator.  I don't know what the name should be yet, but it would be something like this:

```cpp
constexpr auto distance = meters(60);
constexpr auto speed = (miles / hour)(65);

// Old way:
{
    // Using `integer_quotient` says "yes, I really want this to truncate (even to 0)".
    constexpr auto time = integer_quotient(distance, speed);
}

// New way?
{
    constexpr auto time = distance / unblock_integer_division(speed);
}
```

chiphogg:
Instead of `integer_quotient(a, b)`, we now always write
`a / unblock_int_div(b)`.  This gives us everything we used to have with
`integer_quotient`, but with two big advantages:

1. The form for the units code becomes more similar to the non-units
   code that it replaces (i.e., we're using the `/` symbol).

2. We can now support _templated code_ that works for both integral and
   non-integral types: `a / unblock_int_div(b)` works equally well for,
   say, `b` with floating point rep.

There's no longer any use case for `integer_quotient`.  We therefore
immediately deprecate it.  We will keep it deprecated for the entirety
of the 0.4.0 cycle, and remove it for 0.5.0.

Fixes #253.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -393,7 +393,8 @@ class Quantity {
         constexpr bool are_units_quantity_equivalent =
             AreUnitsQuantityEquivalent<UnitT, OtherUnit>::value;
         static_assert(are_units_quantity_equivalent || !uses_integer_division,
-                      "Integer division forbidden: use integer_quotient() if you really want it");
+                      "Integer division forbidden: wrap denominator in `unblock_int_div()` if you "
+                      "really want it");
     }
 
     constexpr Quantity(Rep value) : value_{value} {}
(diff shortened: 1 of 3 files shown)


geoffviola (approved):
Nice! I do like the `/` operator over `integer_quotient`. It is more familiar in terms of operator precedence and argument order.

PR #317, opened 2024-10-30
Clarify a few declarations for static analyzers

djehuti:
Some static analyzers can have trouble with some of these templates; this PR separates them to help parsers that might have difficulty.

### DIFF
--- a/au/code/au/prefix.hh
+++ b/au/code/au/prefix.hh
@@ -134,7 +134,7 @@ constexpr auto mega = PrefixApplier<Mega>{};
 
 template <typename U>
 struct Kilo : decltype(U{} * pow<3>(mag<10>())) {
-    static constexpr detail::ExtendedLabel<1, U> label{detail::corncatenate("k", UnitLabel<U>::value).char_array()};
+    static constexpr detail::ExtendedLabel<1, U> label{detail::concatenate("k", UnitLabel<U>::value).char_array()};
 };
 template <typename U>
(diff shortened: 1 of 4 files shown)


au/code/au/prefix.hh
```diff
@@ -134,7 +134,7 @@ constexpr auto mega = PrefixApplier<Mega>{};
 
 template <typename U>
 struct Kilo : decltype(U{} * pow<3>(mag<10>())) {
-    static constexpr detail::ExtendedLabel<1, U> label = detail::concatenate("k", unit_label<U>());
+    static constexpr detail::ExtendedLabel<1, U> label{detail::corncatenate("k", UnitLabel<U>::value).char_array()};
```
geoffviola, line 137: Nit: spelling 🌽

chiphogg (changes requested):
If we end up landing these kinds of changes to appease static analyzers, we'll do it in a more production ready way.  I already have a good idea of how I plan to do most of these across the codebase, if it comes to that.

That said, I don't want to put any effort into reviewing/landing these changes until we understand whether the broader effort can succeed.  Given that this particular static analyzer does not understand function objects (!!), and that function objects are a core component of our public facing APIs, we need to find out whether they are able to fix their product or provide a workaround.

djehuti:
I deeply regret not having thought of "decorntaminate" before pushing that commit.

chiphogg:
We ended up deciding we didn't need this PR.

PR #318, opened 2024-10-31
Support Quantity non-type template parameters

### LINKED ISSUE #316: Support NTTP for integral rep
This is kind of a neat idea: for any `Quantity` type `Q`, we can have a `Q::NTTP` type, based on `enum class`, which implicitly converts in both directions with `Q`, but is valid as a template parameter.

chiphogg:
Technically, we can't get non-type template parameters of custom class
type until C++20.  But for a `Quantity<U, R>` whose rep `R` is integral,
we can get the next best thing!

This PR introduces a public member typedef, `Quantity<U, R>::NTTP`,
which _can_ be used as a template parameter, because it's an
enumeration.  We support _bidirectional, implicit_ conversion between
`Quantity<U, R>` and `Quantity<U, R>::NTTP`, as long as there's an exact
match for `U` and `R`.  In all other cases, users must use the usual
`Quantity` conversion operators to get it into the right type.  They can
also explicitly request conversion from the NTTP to the `Quantity` by
passing the former to `from_nttp(...)`.

It's a niche use case, but it's finally possible to use a `Quantity` as
a template parameter in a unit-safe way!  There just was no other
comparable solution before this.

Compile time measurements are ongoing, but so far they suggest a small
but nonzero penalty.  It appears that it varies from file to file, but
is generally less than 50 ms.  I think that's an OK price to pay for
this feature.

Fixes #316.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -385,6 +385,34 @@ class Quantity {
             CorrespondingQuantityT<T>{*this}.in(typename CorrespondingQuantity<T>::Unit{}));
     }
 
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Pre-C++20 Non-Type Template Parameter (NTTP) functionality.
+    //
+    // If `Rep` is a built in integral type, then `Quantity::NTTP` can be used as a template
+    // parameter.
+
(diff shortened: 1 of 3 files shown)


geoffviola (approved):
Cool. Safe, compile-time operations are great 😎

Issue #217, opened 2024-01-20
Large numbers in magnitudes cause compiler crashes or long compilation times


Tehforsch:
This is more of a curiosity than an actual issue, so feel free to close this if it doesn't seem relevant. 
I was inspired by the nice talk you (@chiphogg ) gave at CppCon to implement vector space magnitudes in my rust compile time units library (not because the library is mature enough to really need them, but because rusts const-exprs like integers a lot better than floats). 
While doing so I stumbled over the problem that factorizing the integers in the magnitude in the standard/naive way via trial division runs into performance problems (especially if the user really wants to run into performance problems). In my library, the magnitudes are currently read from the exponent / mantissa of a float, but the largest primes that fit into the 52+1 bit mantissa of a double (9007199254740881 for example), can already take a really long time. While implementing some more sophisticated algorithms, I  checked what `au` does to solve this issue and realized that it also uses the standard factorization algorithm, so I tried to confirm my suspicion and compiled the following code:

```c++
struct LargePrimes : decltype(Meters{} * mag<9007199254740881>());
```

if compiled with `clang`, this gives
```
note: constexpr evaluation hit maximum step limit; possible infinite loop?
```
and with `gcc`, I get
```
au.hh:442:5: error: ‘constexpr’ loop iteration count exceeds limit of 262144 (use ‘-fconstexpr-loop-limit=’ to increase the limit)
```

it eventually compiles with 
`g++ -fconstexpr-loop-limit=2147483647 -fconstexpr-ops-limit=2147483647`
but takes about ~100s on my machine. Putting in a ~64bit prime should take about an hour.

This is probably not much of a real-world concern since I maliciously wrote in large prime numbers, but if somebody, for example, were to  autogenerate their unit definitions from a bunch of sensor data, they might get unlucky and accidentally stumble into this. Probably not.

There are much faster algorithms that handle 64 bit integers just fine. If you're at all interested in fixing this, I am by no means knowledgeable in number theory, but I implemented the [Miller Rabin test](https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test) to check for primality first and then the [Pollard's Rho algorithm](https://en.wikipedia.org/wiki/Pollard%27s_rho_algorithm) to factor the composite numbers and with some tweaking, that seems to work fine.

chiphogg:
Very well spotted!  This is a known issue, although not previously tracked in this project.

[...]

PR #320, opened 2024-11-08
Add utilities for modular arithmetic

chiphogg:
These utilities all operate on unsigned 64-bit integers.  The main
design goal is to produce correct answers for all inputs while avoiding
overflow.

Most operations are standard: addition, subtraction, multiplication,
powers.  The one non-standard one is "half_mod_odd".  This operation is
`(x/2) % n`, but only if `n` is odd.  If `x` is also odd, we add `n`
before dividing by 2, which gives us an integer result.  We'll need this
operation for the strong Lucas probable prime checking later on.

These are Au-internal functions, so we use unchecked preconditions: as
long as the caller makes sure the inputs are already reduced-mod-n,
we'll keep them that way.

Helps #217.

### DIFF
--- a/au/code/au/utility/mod.hh
+++ b/au/code/au/utility/mod.hh
@@ -73,6 +73,10 @@ constexpr uint64_t mul_mod(uint64_t a, uint64_t b, uint64_t n) {
 //
 // Precondition: (a < n).
 // Precondition: (n is odd).
+//
+// If `a` is even, this is of course simply `a / 2` (because `(a < n)` as a precondition).
+// Otherwise, we give the result one would obtain by first adding `n` (guaranteeing an even number,
+// since `n` is also odd as a precondition), and _then_ dividing by `2`.
 constexpr uint64_t half_mod_odd(uint64_t a, uint64_t n) {
     return (a / 2u) + ((a % 2u == 0u) ? 0u : (n / 2u + 1u));
 }
--- a/au/code/au/utility/mod.hh
+++ b/au/code/au/utility/mod.hh
@@ -74,7 +74,7 @@ constexpr uint64_t mul_mod(uint64_t a, uint64_t b, uint64_t n) {
 // Precondition: (a < n).
 // Precondition: (n is odd).
 constexpr uint64_t half_mod_odd(uint64_t a, uint64_t n) {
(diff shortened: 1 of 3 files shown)


au/code/au/utility/mod.hh
```diff
@@ -0,0 +1,95 @@
...
+#include <limits>
+
+namespace au {
+namespace detail {
+
+// (a * b) % n
+//
+// Precondition: (a < n).
+// Precondition: (b < n).
+constexpr uint64_t add_mod(uint64_t a, uint64_t b, uint64_t n) {
```
geoffviola, line 27: `((3 * 2) % 4)` is 2, but `add_mod(3, 2, 4)` is 1. Is that expected?

geoffviola (changes requested):

au/code/au/utility/mod.hh
chiphogg: Yes, because `((3 + 2) % 4)` is `1`.  :wink:

(I expect you'll find that `mul_mod(3, 2, 4)` gives `2`.)

chiphogg: Ah, I see --- documentation fail.  Let me go fix the comment...

chiphogg: Fixed.  Also, thought of a new test case, and found one last overflow bug!  This one was only possible to hit when testing the max int $(2^{64} - 1)$.

au/code/au/utility/mod.hh
```diff
@@ -0,0 +1,98 @@
...
+    } else {
+        return n - (b - a);
+    }
+}
+
+// (a * b) % n
+//
+// Precondition: (a < n).
+// Precondition: (b < n).
+constexpr uint64_t mul_mod(uint64_t a, uint64_t b, uint64_t n) {
```
geoffviola, line 51: This implementation seems right. It gets the right answer for the `JUST_UNDER_HALF, 10u, MAX` case as opposed to https://www.geeksforgeeks.org/how-to-avoid-overflow-in-modular-multiplication/

geoffviola (approved):
Implementation looks good. Tests look good. Thanks for the comments.

PR #321, opened 2024-11-08
Add Miller-Rabin probable prime algorithm

chiphogg:
We put this in a new file, `probable_primes.hh`.  Algorithms in this
class can return one of three outcomes:

- "composite" for numbers that are definitely composite
- "probably prime" for _all_ primes, and possibly _some_ composite
  numbers
- "bad input" for numbers that aren't composite or prime (0 or 1), or
  simply aren't appropriate inputs for that particular algorithm.

The wikipedia page[1] gives a good overview of the algorithm.

How do we know that the implementation is correct?  By these main
strategies:

- [x] Miller-Rabin should mark _every_ prime number as "probably prime",
  so test a large number of known primes, each with a variety of bases.

- [x] Test all (odd) inputs up to a certain maximum, making sure that it
  marks every prime as "probably prime", _every pseudoprime to that
  base_ as "probably prime", and every other number as "composite".

- [x] Test a list of known pseudoprimes for a few given bases, and make
  sure that these are all marked as "probably prime" for that base.
  (Making sure that we make the "right mistakes" is a good way to build
  confidence that we implemented the algorithm correctly.)

- [x] Test an extremely large prime (close to the `uint64_t` limit), to
  check that overflow hasn't crept into any of our calculations.

Given that the implementation passes all of these tests, it seems very
likely that our implementation is correct.

Helps #217.

[1] https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test

### DIFF
--- a/au/code/au/utility/test/probable_primes_test.cc
+++ b/au/code/au/utility/test/probable_primes_test.cc
@@ -1,0 +1,205 @@
+// Copyright 2024 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "au/utility/probable_primes.hh"
+
+#include <array>
+#include <string>
+#include <type_traits>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+using ::testing::AllOf;
+using ::testing::ElementsAre;
+using ::testing::Eq;
+using ::testing::Gt;
+using ::testing::Lt;
+
+namespace au {
+namespace detail {
+// Make test output for `PrimeResult` easier to read.
+std::ostream &operator<<(std::ostream &out, const PrimeResult &m) {
+    switch (m) {
+        case PrimeResult::COMPOSITE:
+            return (out << "COMPOSITE");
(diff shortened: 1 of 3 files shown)


au/code/au/utility/test/probable_primes_test.cc
```diff
@@ -0,0 +1,205 @@
...
+    const auto primes = first_n_primes<3'000u>();
+    const auto pseudoprimes = miller_rabin_pseudoprimes_to_base_2();
+
+    // Make sure that we are both _into the regime_ of the pseudoprimes, and that we aren't off the
+    // end of it.
+    ASSERT_THAT(primes.back(), AllOf(Gt(pseudoprimes.front()), Lt(pseudoprimes.back())));
+
+    std::size_t i_prime = 2u;  // Skip 2 and 3; they're too small for Miller-Rabin.
+    std::size_t i_pseudoprime = 0u;
+    for (uint64_t n = primes[i_prime]; i_prime < primes.size(); n += 2u) {
```
geoffviola, line 162: Nit: Can we make `n` `const`?
Q: Why do we add by `2u` when  the primes are all odd after index 0, value 2?

geoffviola (approved):
Good to see solid numerical code in `constexpr`.

au/code/au/utility/test/probable_primes_test.cc
chiphogg: Sorry if it's a little confusing.

- We can't make `n` `const` because we're changing it as a loop variable.
- We're incrementing _`n`_ by two, not the _prime index_ by two.  `n` is how we _advance_ the loop, but `i_prime` is how we _end_ the loop: we're relying on `i_prime` being incremented internally.

It's a non-standard setup, but I think it's a good test: check every odd number until we run out of primes (having already made sure we run out of primes before pseudoprimes).

chiphogg: Actually, this test is just needlessly confusing, and can become a lot more straightforward with a single simple tweak.

PR #322, opened 2024-11-12
Add utilities to compute Jacobi symbols

chiphogg:
This is a core need for the Baillie-PSW implementation: we will use
Jacobi symbols to set the parameters for the Strong Lucas test.

The [Jacobi symbol] `(a/n)` is a multiplicative function of two numbers:
`a` being any integer, and `n` any positive odd integer.  It only takes
values in `{-1, 0, 1}`.  It's defined in terms of another function,
Legendre symbols, which makes it very intimidating to compute... but
actually, computation is straightforward, because the Jacobi symbol has
symmetries and reduction rules that let us skip computing the Legendre
symbols.

I wrote this implementation by reading the linked wikipedia page, and
following the rules.  To test it, I wrote a few manual tests for cases
where the right answer was obvious.  I also tested against an
implementation which I found on Wikipedia.  I decided not to check the
latter test into version control, because I was unsure about the
licensing implications.  However, the Jacobi symbol code will still get
robust testing _indirectly_ in the future, because our Baillie-PSW
implementation will depend on it.

Helps #217.

[Jacobi symbol]: https://en.wikipedia.org/wiki/Jacobi_symbol

### DIFF
--- a/au/code/au/utility/probable_primes.hh
+++ b/au/code/au/utility/probable_primes.hh
@@ -82,5 +82,84 @@ constexpr PrimeResult miller_rabin(std::size_t a, uint64_t n) {
     return PrimeResult::COMPOSITE;
 }
 
+constexpr uint64_t gcd(uint64_t a, uint64_t b) {
+    while (b != 0u) {
+        const auto remainder = a % b;
+        a = b;
+        b = remainder;
+    }
+    return a;
+}
+
+// The conversions `true` -> `1` and `false` -> `0` are guaranteed by the standard.
+//
+// This is a branchless implementation, which should generally be faster.
+constexpr int bool_sign(bool x) { return x - (!x); }
+
+//
+// The Jacobi symbol (a/n) is defined for odd positive `n` and any integer `a` as the product of the
+// Legendre symbols (a/p) for all prime factors `p` of n.  There are several rules that make this
+// easier to calculate, including:
+//
+//  1. (a/n) = (b/n) whenever (a % n) == (b % n).
+//
+//  2. (2a/n) = (a/n) if n is congruent to 1 or 7 (mod 8), and -(a/n) if n is congruent to 3 or 5.
+//
+//  3. (1/n) = 1 for all n.
+//
+//  4. (a/n) = 0 whenever a and n have a nontrivial common factor.
+//
+//  5. (a/n) = (n/a) * (-1)^x if a and n are both odd, positive, and coprime.  Here, x is 0 if
+//     either a or n is congruent to 1 (mod 4), and 1 otherwise.
+//
+constexpr int jacobi_symbol_positive_numerator(uint64_t a, uint64_t n, int start) {
+    int &result = start;
+
+    while (a != 0u) {
+        // Handle even numbers in the "numerator".
+        const int sign_for_even = bool_sign(n % 8u == 1u || n % 8u == 7u);
+        while (a % 2u == 0u) {
+            a /= 2u;
+            result *= sign_for_even;
+        }
+
+        // `jacobi_symbol(1, n)` is `1` for all `n`.
+        if (a == 1u) {
+            return result;
+        }
+
+        // `jacobi_symbol(a, n)` is `0` whenever `a` and `n` have a common factor.
+        if (gcd(a, n) != 1u) {
+            return 0;
+        }
+
+        // At this point, `a` and `n` are odd, positive, and coprime.  We can use the reciprocity
+        // relationship to "flip" them, and modular arithmetic to reduce them.
+
+        // First, compute the sign change from the flip.
+        result *= bool_sign((a % 4u == 1u) || (n % 4u == 1u));
+
+        // Now, do the flip-and-reduce.
+        const uint64_t new_a = n % a;
+        n = a;
+        a = new_a;
+    }
+    return 0;
+}
+constexpr int jacobi_symbol(int64_t raw_a, uint64_t n) {
+    // Degenerate case: n = 1.
+    if (n == 1u) {
+        return 1;
+    }
+
+    // Starting conditions: transform `a` to strictly non-negative values, setting `result` to the
+    // sign we pick up from this operation (if any).
+    int result = bool_sign((raw_a >= 0) || (n % 4u == 1u));
+    auto a = static_cast<uint64_t>(raw_a * bool_sign(raw_a >= 0)) % n;
+
+    // Delegate to an implementation which can only handle positive numbers.
+    return jacobi_symbol_positive_numerator(a, n, result);
+}
+
 }  // namespace detail
 }  // namespace au
--- a/au/code/au/utility/test/probable_primes_test.cc
+++ b/au/code/au/utility/test/probable_primes_test.cc
@@ -200,6 +200,51 @@ TEST(MillerRabin, SupportsConstexpr) {
     static_assert(result == PrimeResult::PROBABLY_PRIME, "997 is prime");
 }
 
+TEST(Gcd, ResultIsAlwaysAFactorAndGCDFindsNoLargerFactor) {
+    for (auto i = 0u; i < 500u; ++i) {
+        for (auto j = 1u; j < i; ++j) {
+            const auto g = gcd(i, j);
+            EXPECT_EQ(i % g, 0u);
+            EXPECT_EQ(j % g, 0u);
+
+            // Brute force: no larger factors.
+            for (auto k = g + 1u; k < j / 2u; ++k) {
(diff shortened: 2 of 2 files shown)


au/code/au/utility/probable_primes.hh
```diff
@@ -82,5 +82,84 @@ constexpr PrimeResult miller_rabin(std::size_t a, uint64_t n) {
...
+constexpr int jacobi_symbol(int64_t raw_a, uint64_t n) {
+    // Degenerate case: n = 1.
+    if (n == 1u) {
+        return 1;
+    }
+
+    // Starting conditions: transform `a` to strictly non-negative values, setting `result` to the
+    // sign we pick up from this operation (if any).
+    int result = bool_sign((raw_a >= 0) || (n % 4u == 1u));
+    auto a = static_cast<uint64_t>(raw_a * bool_sign(raw_a >= 0)) % n;
```
geoffviola, line 158: Isn't `raw_a * bool_sign(raw_a >= 0)` the same as `std::abs(raw_a)`? Maybe, that would be simpler.

I took a look at the [assembly](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,selection:(endColumn:26,endLineNumber:22,positionColumn:26,positionLineNumber:22,selectionStartColumn:26,selectionStartLineNumber:22,startColumn:26,startLineNumber:22),source:'%23include+%3Ccmath%3E%0A%23include+%3Ciostream%3E%0A%0Aconstexpr+int+bool_sign(bool+x)+%7B+return+x+-+(!!x)%3B+%7D%0A%0Auint64_t+foo_0(int64_t+raw_a)+%7B%0A++++return+static_cast%3Cuint64_t%3E(std::abs(raw_a))%3B%0A%7D%0A%0Auint64_t+foo_1(int64_t+raw_a)+%7B%0A++++return+static_cast%3Cuint64_t%3E(raw_a+*+bool_sign(raw_a+%3E%3D+0))%3B%0A%7D%0A%0Aint+main()+%7B%0A++++std::cout+%3C%3C+foo_0(0)+%3C%3C+!'%5Cn!'%3B%0A++++std::cout+%3C%3C+foo_1(0)+%3C%3C+!'%5Cn!'%3B%0A%0A++++std::cout+%3C%3C+foo_0(1)+%3C%3C+!'%5Cn!'%3B%0A++++std::cout+%3C%3C+foo_1(1)+%3C%3C+!'%5Cn!'%3B%0A%0A++++std::cout+%3C%3C+foo_0(-1)+%3C%3C+!'%5Cn!'%3B%0A++++std::cout+%3C%3C+foo_1(-1)+%3C%3C+!'%5Cn!'%3B%0A%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%232',t:'0')),header:(),k:51.16297763490145,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang1910,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-O0+-std%3Dc%2B%2B23',overrides:!(),selection:(endColumn:12,endLineNumber:30,positionColumn:12,positionLineNumber:30,selectionStartColumn:9,selectionStartLineNumber:30,startColumn:9,startLineNumber:30),source:2),l:'5',n:'0',o:'+x86-64+clang+19.1.0+(Editor+%232)',t:'0')),k:48.83702236509856,l:'4',m:54.6806649168854,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+clang+15.0.0',editorid:2,fontScale:14,fontUsePx:'0',j:2,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+19.1.0+(Compiler+%232)',t:'0')),header:(),l:'4',m:11.986001749781277,n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:clang1910,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3+-std%3Dc%2B%2B23',overrides:!(),runtimeTools:!(),source:2,stdinPanelShown:'1',tree:0,wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+clang+19.1.0+(C%2B%2B,+Editor+%232)',t:'0')),l:'4',m:33.33333333333333,n:'0',o:'',s:0,t:'0')),k:48.83702236509856,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). They are the same on `-O2` and `-O3`. For `-O0`, `std::abs` is the one without the branch. `raw_a >= 0` causes a `cmp` operation to be emitted. It may branch.

au/code/au/utility/probable_primes.hh
```diff
@@ -82,5 +82,84 @@ constexpr PrimeResult miller_rabin(std::size_t a, uint64_t n) {
...
+//
+//  3. (1/n) = 1 for all n.
+//
+//  4. (a/n) = 0 whenever a and n have a nontrivial common factor.
+//
+//  5. (a/n) = (n/a) * (-1)^x if a and n are both odd, positive, and coprime.  Here, x is 0 if
+//     either a or n is congruent to 1 (mod 4), and 1 otherwise.
+//
+constexpr int jacobi_symbol_positive_numerator(uint64_t a, uint64_t n, int start) {
+    int &result = start;
```
geoffviola, line 116: Is a reference to an `int` intended? It seems a value type would be trivially copyable, easy to optimize, and easier to examine for correctness.

au/code/au/utility/test/probable_primes_test.cc
```diff
@@ -200,6 +200,52 @@ TEST(MillerRabin, SupportsConstexpr) {
...
+    for (auto i = 3u; i < 99u; i += 2u) {
+        EXPECT_EQ(jacobi_symbol(1, i), 1) << "jacobi(1, " << i << ") should be 1";
+    }
+}
+
+TEST(BoolSign, ReturnsCorrectValues) {
+    EXPECT_EQ(bool_sign(true), 1);
+    EXPECT_EQ(bool_sign(false), -1);
+}
+
```
geoffviola, line 248: Can we get some tests when `jacobi_symbol` is either `-1` or `1`?

geoffviola (approved):
Looks good. I just left some minor comments.

au/code/au/utility/probable_primes.hh
chiphogg: I was trying to create an alias, but you're right --- this is just a weird choice, and removing the `&` is better all around.

au/code/au/utility/probable_primes.hh
```diff
@@ -82,5 +82,84 @@ constexpr PrimeResult miller_rabin(std::size_t a, uint64_t n) {
...
+constexpr uint64_t gcd(uint64_t a, uint64_t b) {
+    while (b != 0u) {
+        const auto remainder = a % b;
+        a = b;
+        b = remainder;
+    }
+    return a;
+}
+
+// The conversions `true` -> `1` and `false` -> `0` are guaranteed by the standard.
```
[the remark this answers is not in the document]
chiphogg: Done.

au/code/au/utility/probable_primes.hh
chiphogg: Good catch --- this was needless obfuscation and premature optimization.

au/code/au/utility/test/probable_primes_test.cc
chiphogg: Done.

While playing around locally, I tested against the C++ snippet on the Wikipedia page.  We reproduced the correct answer for every value in the table.  But I didn't want to check that code in due to questions about licensing a snippet from Wikipedia.

What I've done now is to take specific test cases from the text of the page, each of which is either 1 or -1.

au/code/au/utility/probable_primes.hh
chiphogg: Actually --- nope!  It turns out we can't use `std::abs` here, because it's not `constexpr` until C++23.  :upside_down_face: 

I'll roll the fix into the final landing PR.

Live and learn!

PR #323, opened 2024-11-12
Implement Strong Lucas test and Baillie-PSW

chiphogg:
The [Baillie-PSW] is the star of the show: it's fully deterministic for
all 64-bit integers, and there are no known counterexamples of any size.
In order to reach it, we need to combine the existing Miller-Rabin test
with a [Strong Lucas] test, the latter of which is the "hard work" part
of this PR.

The first step is to find the right value of the "D" parameter to use:
the first from the sequence {5, -7, 9, -11, ...} whose Jacobi symbol
(see parent PR) is -1.  We represent this as an `uint64_t`-and-`bool`
struct, rather than a simple `int`, for two reasons.  First, it's easier
to write this incrementing/alternating pattern with a separate sign.
Second, when we use `D.mag` in calculations, it's convenient if it's
already a `uint64_t` like almost all of the rest of our types.

Oh --- and, this step won't work if `n` is a perfect square, so we add a
separate guarding step to filter this situation out.

Now for the meat of the calculation.  We need to compute elements of the
Lucas Sequences, $U_k$ and $V_k$ for certain specific indices.  To do
that efficiently, we use the fact that $U_1 = V_1 = 1$, and apply the
index-doubling and index-incrementing relations:

Equations for U and V

The above are derived assuming the Lucas parameters $P=1$, $Q=(1-D)/4$,
as is always done for the best-studied parameterization of Baillie-PSW.
We use a bit-representation of the desired indices to decide when to
double and when to increment, also commonly done (see the Implementation
section in the [Strong Lucas] reference).

Finally, we are able to combine the new Strong Lucas test with the
existing Miller-Rabin test to obtain a working Baillie-PSW test.

We use the same kinds of verification strategies for Strong Lucas and
Baillie-PSW as we used for Miller-Rabin, except that we don't test
pseudoprimes for Baillie-PSW because nobody has ever found any yet.

Helps #217.

[Baillie-PSW]: https://en.wikipedia.org/wiki/Baillie%E2%80%93PSW_primality_test
[Strong Lucas]: https://en.wikipedia.org/wiki/Lucas_pseudoprime#Strong_Lucas_pseudoprimes

### DIFF
--- a/au/code/au/utility/probable_primes.hh
+++ b/au/code/au/utility/probable_primes.hh
@@ -84,6 +84,27 @@ constexpr PrimeResult miller_rabin(std::size_t a, uint64_t n) {
     return PrimeResult::COMPOSITE;
 }
 
+//
+// Test whether the number is a perfect square.
+//
+constexpr bool is_perfect_square(uint64_t n) {
+    if (n < 2u) {
+        return true;
+    }
+
+    uint64_t prev = n / 2u;
+    while (true) {
+        const uint64_t curr = (prev + n / prev) / 2u;
+        if (curr * curr == n) {
+            return true;
+        }
+        if (curr >= prev) {
+            return false;
+        }
+        prev = curr;
+    }
+}
+
 constexpr uint64_t gcd(uint64_t a, uint64_t b) {
     while (b != 0u) {
         const auto remainder = a % b;
@@ -162,6 +183,161 @@ constexpr int jacobi_symbol(int64_t raw_a, uint64_t n) {
 
     // Delegate to an implementation which can only handle positive numbers.
     return jacobi_symbol_positive_numerator(a, n, result);
+}
+
+struct LucasDParameter {
+    uint64_t mag = 5u;
+    bool is_neg = false;
+
+    friend constexpr int as_int(const LucasDParameter &D) { return D.is_neg ? -D.mag : D.mag; }
+    friend constexpr void increment(LucasDParameter &D) {
+        D.mag += 2u;
+        D.is_neg = !D.is_neg;
+    }
+};
+
+//
+// The first `D` in the infinite sequence {5, -7, 9, -11, ...} whose Jacobi symbol is (-1) is the
+// `D` we want to use for the Strong Lucas Probable Prime test.
+//
+// Requires that `n` is *not* a perfect square.
+//
+constexpr LucasDParameter find_first_D_with_jacobi_symbol_neg_one(uint64_t n) {
+    LucasDParameter D{5u, false};
+    while (jacobi_symbol(as_int(D), n) != -1) {
+        increment(D);
+    }
+    return D;
+}
+
+//
+// Elements of the Lucas sequence.
+//
+// The default values give the first element (i.e., k=1) of the sequence.
+//
+struct LucasSequenceElement {
+    uint64_t U = 1u;
+    uint64_t V = 1u;
+};
+
+// Produce the Lucas element whose index is twice the input element's index.
+constexpr LucasSequenceElement double_strong_lucas_index(const LucasSequenceElement &element,
+                                                         uint64_t n,
+                                                         LucasDParameter D) {
+    const auto &U = element.U;
+    const auto &V = element.V;
+
+    uint64_t V_squared = mul_mod(V, V, n);
+    uint64_t D_U_squared = mul_mod(D.mag, mul_mod(U, U, n), n);
+    uint64_t V2 =
+        D.is_neg ? sub_mod(V_squared, D_U_squared, n) : add_mod(V_squared, D_U_squared, n);
+    V2 = half_mod_odd(V2, n);
+
+    return LucasSequenceElement{
+        mul_mod(U, V, n),
+        V2,
+    };
+}
+
+// Find the next element in the Lucas sequence, using parameters for strong Lucas probable primes.
+constexpr LucasSequenceElement increment_strong_lucas_index(const LucasSequenceElement &element,
+                                                            uint64_t n,
+                                                            LucasDParameter D) {
+    const auto &U = element.U;
+    const auto &V = element.V;
+
+    auto U2 = half_mod_odd(add_mod(U, V, n), n);
+
+    const auto D_U = mul_mod(D.mag, U, n);
+    auto V2 = D.is_neg ? sub_mod(V, D_U, n) : add_mod(V, D_U, n);
+    V2 = half_mod_odd(V2, n);
+
+    return LucasSequenceElement{U2, V2};
+}
+
+// Compute the strong Lucas sequence element at index `i`.
+constexpr LucasSequenceElement find_strong_lucas_element(uint64_t i,
+                                                         uint64_t n,
+                                                         LucasDParameter D) {
+    LucasSequenceElement element{};
+
+    bool bits[64] = {};
+    std::size_t n_bits = 0u;
+    while (i > 1u) {
+        bits[n_bits++] = (i % 2u == 1u);
+        i /= 2u;
+    }
+
+    for (std::size_t j = n_bits; j > 0u; --j) {
+        element = double_strong_lucas_index(element, n, D);
+        if (bits[j - 1u]) {
+            element = increment_strong_lucas_index(element, n, D);
+        }
+    }
+
+    return element;
+}
+
+//
+// Perform a strong Lucas primality test on `n`.
+//
+constexpr PrimeResult strong_lucas(uint64_t n) {
+    if (n < 2u || n % 2u == 0u) {
+        return PrimeResult::BAD_INPUT;
+    }
+
+    if (is_perfect_square(n)) {
+        return PrimeResult::COMPOSITE;
+    }
+
+    const auto D = find_first_D_with_jacobi_symbol_neg_one(n);
+
+    const auto params = decompose(n + 1u);
+    const auto &s = params.power_of_two;
+    const auto &d = params.odd_remainder;
+
+    auto element = find_strong_lucas_element(d, n, D);
+    if (element.U == 0u) {
+        return PrimeResult::PROBABLY_PRIME;
+    }
+
+    for (std::size_t i = 0u; i < s; ++i) {
+        if (element.V == 0u) {
+            return PrimeResult::PROBABLY_PRIME;
+        }
+        element = double_strong_lucas_index(element, n, D);
+    }
+
+    return PrimeResult::COMPOSITE;
+}
+
+//
+// Perform the Baillie-PSW test for primality.
+//
+// Returns `BAD_INPUT` for any number less than 2, `COMPOSITE` for any larger number that is _known_
+// to be prime, and `PROBABLY_PRIME` for any larger number that is deemed "probably prime", which
+// includes all prime numbers.
+//
+// Actually, the Baillie-PSW test is known to be completely accurate for all 64-bit numbers;
+// therefore, since our input type is `uint64_t`, the output will be `PROBABLY_PRIME` if and only if
+// the input is prime.
+//
+constexpr PrimeResult baillie_psw(uint64_t n) {
+    if (n < 2u) {
+        return PrimeResult::BAD_INPUT;
+    }
+    if (n < 4u) {
+        return PrimeResult::PROBABLY_PRIME;
+    }
+    if (n % 2u == 0u) {
+        return PrimeResult::COMPOSITE;
+    }
+
+    if (miller_rabin(2u, n) == PrimeResult::COMPOSITE) {
+        return PrimeResult::COMPOSITE;
+    }
+
+    return strong_lucas(n);
 }
 
 }  // namespace detail
--- a/au/code/au/utility/test/probable_primes_test.cc
+++ b/au/code/au/utility/test/probable_primes_test.cc
@@ -200,6 +200,110 @@ TEST(MillerRabin, SupportsConstexpr) {
     static_assert(result == PrimeResult::PROBABLY_PRIME, "997 is prime");
 }
 
+TEST(IsPerfectSquare, ProducesCorrectAnswers) {
+    auto next_sqrt = 0u;
+    for (auto n = 0u; n < 400'000u; ++n) {
+        const auto next_square = next_sqrt * next_sqrt;
+
+        const auto is_square = (n == next_square);
+        if (is_square) {
+            ++next_sqrt;
+        }
+
+        EXPECT_EQ(is_perfect_square(n), is_square) << "n = " << n;
+    }
+}
+
+std::vector<uint64_t> strong_lucas_pseudoprimes() {
+    // https://oeis.org/A217255
+    return {5459u,   5777u,   10877u,  16109u,  18971u,  22499u,  24569u,  25199u,  40309u,
+            58519u,  75077u,  97439u,  100127u, 113573u, 115639u, 130139u, 155819u, 158399u,
+            161027u, 162133u, 176399u, 176471u, 189419u, 192509u, 197801u, 224369u, 230691u,
+            231703u, 243629u, 253259u, 268349u, 288919u, 313499u, 324899u};
+}
+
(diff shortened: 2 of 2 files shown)


au/code/au/utility/test/probable_primes_test.cc
```diff
@@ -200,6 +200,115 @@ TEST(MillerRabin, SupportsConstexpr) {
...
+        if (p > 2u) {
+            EXPECT_THAT(strong_lucas(p), Eq(PrimeResult::PROBABLY_PRIME)) << p;
+        }
+    }
+}
+
+TEST(StrongLucas, GetsFooledByKnownPseudoprimes) {
+    for (const auto &p : strong_lucas_pseudoprimes()) {
+        ASSERT_EQ(miller_rabin(2u, p), PrimeResult::COMPOSITE) << p;
+        EXPECT_THAT(strong_lucas(p), Eq(PrimeResult::PROBABLY_PRIME)) << p;
```
hoffbrinkle, line 242: A nice side effect of using the matcher syntax is that things tend to align visually.

```suggestion
        ASSERT_THAT(miller_rabin(2u, p), Eq(PrimeResult::COMPOSITE)) << p;
        EXPECT_THAT(strong_lucas(p), Eq(PrimeResult::PROBABLY_PRIME)) << p;
```

au/code/au/utility/test/probable_primes_test.cc
```diff
@@ -200,6 +200,115 @@ TEST(MillerRabin, SupportsConstexpr) {
...
+    auto next_sqrt = 0u;
+    for (auto n = 0u; n < 400'000u; ++n) {
+        const auto next_square = next_sqrt * next_sqrt;
+
+        const auto is_square = (n == next_square);
+        if (is_square) {
+            ++next_sqrt;
+        }
+
+        EXPECT_EQ(is_perfect_square(n), is_square) << "n = " << n;
```
hoffbrinkle, line 213:
```suggestion
        EXPECT_THAT(is_perfect_square(n), Eq(is_square)) << "n = " << n;
```

au/code/au/utility/test/probable_primes_test.cc
```diff
@@ -200,6 +200,115 @@ TEST(MillerRabin, SupportsConstexpr) {
...
+    // https://oeis.org/A217255
+    return {5459u,   5777u,   10877u,  16109u,  18971u,  22499u,  24569u,  25199u,  40309u,
+            58519u,  75077u,  97439u,  100127u, 113573u, 115639u, 130139u, 155819u, 158399u,
+            161027u, 162133u, 176399u, 176471u, 189419u, 192509u, 197801u, 224369u, 230691u,
+            231703u, 243629u, 253259u, 268349u, 288919u, 313499u, 324899u};
+}
+
+TEST(LucasDParameter, CanConvertToInt) {
+    EXPECT_EQ(as_int(LucasDParameter{5u, true}), 5);
+    EXPECT_EQ(as_int(LucasDParameter{7u, false}), -7);
```
hoffbrinkle, line 227:
```suggestion
    EXPECT_THAT(as_int(LucasDParameter{5u, true}), Eq(5));
    EXPECT_THAT(as_int(LucasDParameter{7u, false}), Eq(-7));
```

au/code/au/utility/test/probable_primes_test.cc
```diff
@@ -200,6 +200,115 @@ TEST(MillerRabin, SupportsConstexpr) {
...
+    ASSERT_THAT(miller_rabin(2u, n), Eq(PrimeResult::PROBABLY_PRIME));
+    EXPECT_THAT(baillie_psw(n), Eq(PrimeResult::COMPOSITE));
+}
+
+TEST(BailliePSW, HandlesVeryLargePrimes) {
+    for (const auto &p : {
+             225'653'407'801ull,
+             334'524'384'739ull,
+             9'007'199'254'740'881ull,
+             18'446'744'073'709'551'557ull,
```
hoffbrinkle, line 306: Rather than relying on `ull` to be equivalent to a `uint64_t`;

```suggestion
             uint64_t{225'653'407'801},
             uint64_t{334'524'384'739},
             uint64_t{9'007'199'254'740'881},
             uint64_t{18'446'744'073'709'551'557},
```

geoffviola: +1. Luckily, `ull` is produces literals of a size of 8 bytes on clang, gcc, and MSVC. It's not guaranteed. Literals with `ul` are 8 bytes on clang and gcc, but 4 bytes on MSVC. [godbolt](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsQAJnL0WTUKgCkBgEJXr5FY1RECQ6s0wBhdPQCubJkMADnIPABkCJmwAOX8AI2xSEAA2cgAHdCViVyZvPwDg9MzsoQio2LYEpNTHbGcckSIWUiI8/0CDENr6oUbmojKY%2BMSUhyaWtoLOsf7BiqqUgEoHdF9SVE4uKwBmSNQ/HABqC23PVCUiQmET3AstAEEdvYPsY9PXC9Jsdhu7x4fIkRDmwWJEIItjgB2WwPQ5ww4XTAgNCrIEnTzohEEABe2HQNAgvkByWkAH0iBD0ZiwFsAKyeJg0k4w%2B7whGXZEYXxo06YrK4/EQHi%2BeiU3mnQ5M%2BmMrbbFlsxGc1FvDES/l4gnC%2BiilXUukMplyv5/NlfIhrJiHLTM42QgAiXGW9G4tP4gS4OnI6G4GJsdgRq3Wrx2Bn4RG0juWSG%2BOCS4PIAGsQLStAA6SGQ2nbHhBaQGbYGZKQ5LbWnGbjSfhsQxacjuz3erj8JQgWvhj2O8hwWAoDBsNIMRKUah9geMJIAN1QaTSpInPAAnKTpFpSQA1EQ8SGknir9RBVJ0ehERItiBxCPkOKRZoAT24fCvN9It4A8nFdHV24%2B%2BxxhK%2BmHoe8O3IHA4l8YBPAkegW14fgcBBMxJBAwgvnqCdsFgz1sHUOpuU2R9AWwZ0QPoAg4lIO9vBwS8iFIAhqzg8gMNIOJMmwO1sEQ4AyLMCNlhoExgCUNcCGwAB3V80mYB9%2BEEYQxAkTgZDkYRlDUTQQP0HhjFMcw7DsYxyJbSBlnQNIXCEWD%2BHQFj6JwEz426SzAggDxJkCHTwkiIZKhGHSMiyFyPKkIogpyOZhiSHTnIacZWh8dpQti3p4sivzopmCZEoKGK0p8%2BZ/OWJRAw2QwnRdN1L0bQ592SYElCnQ55wXVMV0ODct1TXdDggfBiDIY58wMRYw34qMYxGeMSMrchqx4ZJUwLZIsy0ZIeAMAxpGkBc62q7hm1bch2x0ZZu2QEAcLwkgKCoCBmmE1RTGIkQkHQcT3R/dB%2BwYFgXKeqJ6Fe97634UdB0nadZxa5dV067dd1JOryHB8donYTYwe%2BsdElfblgY%2By8rtQe5SGEg7Qlw1BGiuCn5NEcRJBU%2Bn1I0S99CMEw%2BMsP0bCMuJHLMiycmsr07IIBz4GWUhfGEBjsAAFXQHxBZWNYyqMRFIgBl63sJpi6OwTha3Eyi0lkiquFdPaQMbXAqe5Qa6oapqWrarQOs3SFuo9iBfVsPnDn6m6hoLUbjvG8suFm6t81TIttmkZIF22IIM1pWkdNBr0KZbNt%2BK7RAIF7bGIeHCBUZGYAZCMI8T1IM8LxA69WBfWSn1bt8Py/dvf2YIgAKAy8wIgqDtVgx8EL05DPVQr8CAwrD%2BGJ/D26IkjPTIiiqKwTZPTohj25YtiVE47jeNADsBKEkSxMk6TPrk%2BRFKZ2QWZUNmtJAbZdO5gy%2Ba3qrcyLlRa2USPZTCUsHDEXnm4NyTAvA5U8qEBB6UFgBWKMFJBoVAolCYGg/y0CnAuT6NlfIyCUpMFIQMAqUVkrxRCnlWYtCMpSGKqVTgI0o7W2zjVOqABaEkhxgCoFQIcGQqYDC9WDoNEM4cTqRijjHGsNsGy5yOgos6RcUArxuuXB6SgdZAz1o/FGpcTD/WesYkGl5K5JGrnmMxP00YYwpnYvGogTFEypqTcmTZKZ1BppEOmz9GbKTfvIVmmlPT6B/lzfSvN7CAKgcAkWFMwGkAgarGWcsOBKxVlAkq6tOFjCuEYgmpjDbG3IKbFg5s4KWx4ftLg9trpOwPII6QwjRHiOkJI3q/tDJB0ICHORY0r6W2UQYWsvD1H5wmYmZMaZ07ZlzMNIsJYywkW2FVW2FNNGF3gOdYuKJnFDjunYtA8TFy7gEAweujdLwtzvO3Z5L53yfmcL3b6f4B6AWArPbA4FILQQnvBLi0897wQIGhFwi9vHXQIvwdel4t6URfNRJFx16KMUfMfdiZ9p6REvqdAQN9RISSkjJJi9MX7hNUooD%2B0S9CGF/gkgOSTjIpOFlZdJ4tJamSIbA1y7lsFeVQSw9BYU8GMOlS5AhmVKHUNlUq/K5Q6FMLIUlTVND1WsJ4Ow4p5VuG7LUVwWqHShH7D0uI1qPBUy%2BxkaQUOI1xmnWWEmaQQRUxBG2JCLQm1tgrS0AuIItJUgzSrMmPpyQghrRJAuBaQQw2yFmf4vOEcr6HJ7CAVEaRuTlzsejDg3ABFWvieI2kDqHXLxGWQCWrLaVhKkBEtSTL2astqfUvgjTTU2W4B4/NQJ8QWuSJ0w41qzC2u6g6gZ5jEguvkZHaMLBYzUEmVGng9rQ3JC0FoLcO1IRBA2hmVR/b00aILscnRDs9F3QMeUrxTFUZ/RyI%2BmxIFLmTurom2sRbXH%2BPcfjJ92EfFk1FsTIJ7on4KWbczSJ7av46XiWgf%2BnKBbcpAdwfhr5tiHH4YiE4doeYcvzDZflkDBU5JcHk5WMFCkcNZVrJg779aPiqUdLtFsTVnpzi0297Sx3lptYuGdvtBmBydYut1ijI1zW/tsJa2xlMqdUyp3jjZDrzPdYsnakitqxp4LSYskICypyjjsjT%2BzI4kVDFZ/xmjlgsSyG4aQQA%3D%3D)

chiphogg: Literals generally are a funny and subtle thing.  The type of an integral literal will be the default integral type (usually `int` or `uint`), _unless_ the value won't fit in that type, in which case, the type will be the smallest type where it _will_ fit, being signed for no suffix and unsigned for `u`.  (I don't remember where I learned this.)

The problem this causes is that if you use a simple braced list like this, and just use a `u` suffix for everything, the compiler will _deduce inconsistent types_ if some of those literals can fit in a `uint`, and some can't.  Therefore, I decided to apply a literal suffix that would cover all of them.

But, the feedback in this thread shows why that's not a great idea.  Explicit `uint64_t` seems much more readable.  (The other alternative would be to make an explicit `std::vector<uint64_t>`, which would make mixing-and-matching OK.)

au/code/au/utility/probable_primes.hh
```diff
@@ -164,5 +185,166 @@ constexpr int jacobi_symbol(int64_t raw_a, uint64_t n) {
...
+constexpr LucasSequenceElement find_strong_lucas_element(uint64_t i,
+                                                         uint64_t n,
+                                                         LucasDParameter D) {
+    LucasSequenceElement element{};
+
+    bool bits[64] = {};
+    std::size_t n_bits = 0u;
+    while (i > 1u) {
+        bits[n_bits++] = (i % 2u == 1u);
+        i /= 2u;
```
geoffviola, line 274: Comment: In my mind, it makes more sense to mask the bit and shift. The compiler produces nearly identical instructions even without optimizations: [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:27,positionColumn:2,positionLineNumber:27,selectionStartColumn:2,selectionStartLineNumber:27,startColumn:2,startLineNumber:27),source:'%23include+%3Ccstdint%3E%0A%23include+%3Ciostream%3E%0A%0Avoid+foo_0(uint64_t+i,+bool+*const+bits)+%7B%0A++++std::size_t+n_bits+%3D+0u%3B%0A++++while+(i+%3E+1u)+%7B%0A++++++++bits%5Bn_bits%2B%2B%5D+%3D+(i+%25+2u+%3D%3D+1u)%3B%0A++++++++i+/%3D+2u%3B%0A++++%7D%0A%7D%0A%0Avoid+foo_1(uint64_t+i,+bool+*const+bits)+%7B%0A++++std::size_t+n_bits+%3D+0u%3B%0A++++while+(i+%3E+1u)+%7B%0A++++++++bits%5Bn_bits%2B%2B%5D+%3D+(i+%26+1u)%3B%0A++++++++i+%3E%3E%3D+1%3B%0A++++%7D%0A%7D%0A%0Aint+main()+%7B%0A++++bool+bits%5B64%5D+%3D+%7B%7D%3B%0A%0A++++foo_0(100,+bits)%3B%0A++++foo_1(100,+bits)%3B%0A%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%232',t:'0')),header:(),k:50.775318423267635,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang1910,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-O0',overrides:!(),selection:(endColumn:24,endLineNumber:60,positionColumn:24,positionLineNumber:60,selectionStartColumn:24,selectionStartLineNumber:60,startColumn:24,startLineNumber:60),source:2),l:'5',n:'0',o:'+x86-64+clang+19.1.0+(Editor+%232)',t:'0')),k:48.83702236509856,l:'4',m:54.6806649168854,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+clang+15.0.0',editorid:2,fontScale:14,fontUsePx:'0',j:2,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+19.1.0+(Compiler+%232)',t:'0')),header:(),l:'4',m:11.986001749781277,n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:clang1910,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3+-std%3Dc%2B%2B23',overrides:!(),runtimeTools:!(),source:2,stdinPanelShown:'1',tree:0,wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+clang+19.1.0+(C%2B%2B,+Editor+%232)',t:'0')),l:'4',m:33.33333333333333,n:'0',o:'',s:0,t:'0')),k:49.22468157673238,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)
```suggestion
        bits[n_bits++] = (i % 1u);
        i >>= 1;
```

geoffviola (approved):
Very cool. Good work! 🎉

PR #325, opened 2024-11-19
Weaken `find_first_factor` contract to any prime

chiphogg:
It turns out, none of our use cases require us to find the _first_
factor.  All we really need is to find _any prime_ factor.  (And if the
number itself is prime, of course, then the only option will be to
return the number itself.)

Weakening this contract enables us to take advantage of faster factoring
methods that aren't guaranteed to find the _smallest_ factor. Obviously,
we could use these faster methods to build a function that satisfies our
old contract, by repeatedly applying them to _fully_ factor a number,
and then taking the smallest one.  But this adds extra computation for
no clear benefit.

Helps #217.

### DIFF
--- a/au/code/au/utility/test/factoring_test.cc
+++ b/au/code/au/utility/test/factoring_test.cc
@@ -17,6 +17,8 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+using ::testing::AnyOf;
+using ::testing::Eq;
 using ::testing::Gt;
 using ::testing::Le;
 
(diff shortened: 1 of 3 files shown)


au/code/au/utility/test/factoring_test.cc
```diff
@@ -40,35 +42,39 @@ TEST(FirstPrimes, HasOnlyPrimesInOrderAndDoesntSkipAny) {
...
 
-    EXPECT_EQ(find_first_factor(196961u), 196961u);
+    EXPECT_EQ(find_prime_factor(196961u), 196961u);
 }
 
-TEST(FindFirstFactor, FindsFirstFactor) {
-    EXPECT_EQ(find_first_factor(7u * 11u * 13u), 7u);
-    EXPECT_EQ(find_first_factor(cube(196961u)), 196961u);
+TEST(FindFactor, FindsFirstFactorWhenFirstFactorIsSmall) {
+    EXPECT_EQ(find_prime_factor(7u * 11u * 13u), 7u);
```
geoffviola, line 56: Q: Should these tests use `AnyOf` as well as opposed to equaling the smallest factor?

geoffviola (approved):
Relaxing the contract makes sense.

au/code/au/utility/test/factoring_test.cc
chiphogg: Probably, yeah.  I was just a little bit lazy when I put the PR up initially.  :slightly_smiling_face:

Issue #85, opened 2023-02-06
Solve most `[UNLABELED_UNIT]` instances by generating magnitude labels


chiphogg:
Right now, the unit resulting from scaling a unit by a magnitude gets the default label, `[UNLABELED_UNIT]`.  It's important that it _not_ have the same label as the unscaled unit.  Still, we can do much better if we design magnitude labels: automatic string constants that represent magnitudes.

Possible examples:

- `"1 / 18"`
- `"Pi / 180"`
- `"5,280"`
- `"(3 * sqrt(3)) / (5 * Pi)"`

To solve this, we would need to come up with a collection of canonical examples we can use as acceptance tests, and work out the general rules.  We also need to figure out how to handle things like parentheses, powers, and roots.

chiphogg:
Based on discussion on the discord DMs for the C++ standard units library, I think a scaled unit should have the format `[M U]`, where `M` is the magnitude label, and `U` is the label for the unscaled unit.  That is: the new label should be fully enclosed in square brackets.

PR #331, opened 2024-11-27
Add labels for integer and rational magnitudes

chiphogg:
Think of this as "just enough of #85 to unblock #105".  The fact that
our common unit labels don't tell you the _size_ of the unit (#105) has
really, really been bugging me for a long time.  Recently, I realized we
don't need to do all of #85 to get it!  Instead, all we need to do is:

1. Build a _mechanism_ that we can easily _extend_.

2. Cover the most important use cases.

This PR creates the `MagnitudeLabel` trait mechanism (also accessible
via a function/value interface as `mag_label`).  We enumerate the
various categories of magnitudes that we can label, defaulting to
"unsupported".  The first two supported categories are _integers_ (that
fit in `std::uintmax_t`), and _rationals_.

We also add a trait, `has_exposed_slash`, looking forward to the obvious
use case of auto-generating labels for scaled units.  Those labels will
have the form `"[M U]"` for a unit of label `"U"` scaled by a magnitude
of label `"M"`.  If `has_exposed_slash` is `true` for a given magnitude
label, then we'll know to make this `"[(M) U]"` instead.

Finally, we move a couple of `StringConstant`-ish utilities into
`"string_constant.hh"`, so that we can use them in our implementation.

Helps #85.

### DIFF
--- a/au/code/au/magnitude.hh
+++ b/au/code/au/magnitude.hh
@@ -22,6 +22,7 @@
 #include "au/power_aliases.hh"
 #include "au/stdx/utility.hh"
 #include "au/utility/factoring.hh"
+#include "au/utility/string_constant.hh"
 #include "au/zero.hh"
 
 // "Magnitude" is a collection of templated types, representing positive real numbers.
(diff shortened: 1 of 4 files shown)


geoffviola (approved):
Implementation, tests and documentation looks good. Thanks!

Issue #185, opened 2023-10-29
Always return Quantity, not raw numbers (breaking change)


chiphogg:
Right now, the result of a Quantity computation whose unit is equivalent to 1 is a raw number, rather than a Quantity equivalent to `unos`.  I made this decision because people generally expect this kind of computation to produce a raw number.  We do provide _implicit convertibility_ to the raw number type, but this can sometimes fail to trigger.  Experience shows that the only thing that _always_ acts like a raw numeric type is that type itself.

I believe that further experience has shown this to be a mistake.  It makes generic programming too hard.  I used to think this was fine, because generic programming is mainly the domain of more experienced users, who will know how to handle this.  I was wrong: the reason it's still a problem is that _users can't be confident whether they'll get a Quantity or raw number just from inspecting a computation_.  If we refactor and change the unit of one of the participating quantities, it can even switch from one to the other!

The way to solve this is to provide some idiomatic function --- say, `as_raw_number(...)` --- which "unpacks" a dimensionless quantity.  This helps other (non-unity) dimensionless use cases, too.  Alternatively, we could just tell everyone to write `.in(unos)`.

**This is a breaking change.**  The best time to do this would be when we're nearly ready to release 0.4.0.  We will also need to use Aurora's AV repo to see how "bad" this change would be in terms of what it does to the code.

jaredjxyz:
+1 for this change... There are certainly tradeoffs, but I'll also add that this could unlock the ability to create more dimensionless units - for example, `.in(dozens)`, `.in(bakers_dozens)`

chiphogg:
One slight complication to consider is that some users might have functions that are generic not just across different `Quantity` types, but also between `Quantity` types and raw numbers (and possibly other types).  See #248 for an example.

This means that `.in(unos)` probably won't work.  However, we could design the `as_raw_number(...)` API so that:
- it returns its argument for anything other than a `Quantity`
- it returns the value in unos for a dimensionless `Quantity`
- it produces an error for any _other_ type of `Quantity`

PR #333, opened 2024-11-29
Add `as_raw_number(...)` utility

chiphogg:
This lets us pave the way to change the behavior of quantity arithmetic
when all units cancel out (#185).  If we make this change all at once,
it'll generally be too big and hard to handle.  But if we provide this
utility now, then people can migrate individual callsites gradually over
time.  Then, at the end (probably 0.5.0?), making the switch to the new
behavior won't be such a big change.

Helps #185.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -534,6 +534,20 @@ constexpr auto operator%(Quantity<U1, R1> q1, Quantity<U2, R2> q2) {
     return make_quantity<U>(q1.in(U{}) % q2.in(U{}));
 }
 
+// Callsite-readable way to convert a `Quantity` to a raw number.
+//
+// Only works for dimensionless `Quantities`; will return a compile-time error otherwise.
+//
+// Identity for non-`Quantity` types.
+template <typename U, typename R>
+constexpr R as_raw_number(Quantity<U, R> q) {
+    return q.as(UnitProductT<>{});
+}
+template <typename T>
+constexpr T as_raw_number(T x) {
+    return x;
+}
+
 // Type trait to detect whether two Quantity types are equivalent.
 //
 // In this library, Quantity types are "equivalent" exactly when they use the same Rep, and are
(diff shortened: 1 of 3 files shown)


chiphogg:
> Sounds like this change will help migrate.
> 
> I feel like `.in(unos)` isn't great if the quantity wasn't a `unos` to begin with. But I do like keeping in the tradition of mentioning the type somewhere in the conversion statement. For example,
> 
> ```
> auto foo_p = percent(0.7);
> ...
> double foo_d = foo_p.in(percent);
> ```
> 
> seems better than
> 
> ```
> auto foo_p = percent(0.7);
> ...
> double foo_d = as_raw_number(foo_p);
> ```

Great analysis, and I agree.  The main use case for using `as_raw_number` with non-unitless dimensionless quantities is for _results of calculations_.  This will let somebody write `as_raw_number(frequency * duration)`, and have it continue to work even when we refactor to use more convenient units for `frequency` or `duration` --- as long as the combination can be safely converted to a raw number.

chiphogg:
> Sounds like this change will help migrate.
> 
> I feel like `.in(unos)` isn't great if the quantity wasn't a `unos` to begin with. But I do like keeping in the tradition of mentioning the type somewhere in the conversion statement. For example,
> 
> ```
> auto foo_p = percent(0.7);
> ...
> double foo_d = foo_p.in(percent);
> ```
> 
> seems better than
> 
> ```
> auto foo_p = percent(0.7);
> ...
> double foo_d = as_raw_number(foo_p);
> ```

Oh!  And just to be crystal-clear, `foo_p.in(percent)` has _different semantics_ than `as_raw_number(foo_p)` --- the former gives `0.7`, while the latter gives `0.007` (or, in each case, as nearly as we can represent those values in `double`).

[Non-unitless dimensionless units are really hard](https://aurora-opensource.github.io/au/main/discussion/concepts/dimensionless/#implicit-conversions).  :sweat_smile:

PR #339, opened 2024-12-02
Convert `CommonPointUnit` to EQUIV-style labels

### LINKED ISSUE #105: Indicate relative magnitudes in common unit labels

chiphogg:
This is a kind of "good enough for now" placeholder.  It's a big
improvement over what we had before, because it shows the actual values.
There may be some subtleties that we could improve on, but we will save
them for a future release.

Fixes #105.

### DIFF
--- a/au/code/au/quantity_point_test.cc
+++ b/au/code/au/quantity_point_test.cc
@@ -21,8 +21,17 @@
 
 using ::testing::Not;
 using ::testing::StaticAssertTypeEq;
+using ::testing::StrEq;
 
 namespace au {
+namespace {
+template <typename T>
+std::string stream_to_string(const T &t) {
+    std::ostringstream oss;
+    oss << t;
+    return oss.str();
+}
+}  // namespace
 
 struct Meters : UnitImpl<Length> {};
 constexpr QuantityMaker<Meters> meters{};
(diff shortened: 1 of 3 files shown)


geoffviola (approved):
Good improvement. Thanks.

PR #340, opened 2024-12-02
Document constants

### LINKED ISSUE #90: Support constants

chiphogg:
We update the installation guide, both to point users to the headers for
the constants, and to tell them how to include constants in the
single-file script.  This made me realize that we had forgotten to make
constants available in `@au//au`, so I updated the BUILD file.

On the constant reference page, we now list the built-in constants.  I
like the way the table looks!

On the monovalue types page, constants are a great example of monovalue
types, so I add them as an example.

We also add a how-to guide for making new constants.

Finally, now that we include built-in constants with the library, I
think Au's constant support is now best-in-class, so I am updating our
Alternatives page to list us as such.

Fixes #90.

### DIFF
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -25,7 +25,9 @@ cc_library(
     deps = [
         ":chrono_interop",
         ":constant",
+        ":constants",
         ":math",
+        ":units",
     ],
 )
 
--- a/docs/alternatives/index.md
+++ b/docs/alternatives/index.md
@@ -908,15 +908,12 @@ features.
             href="https://mpusz.github.io/mp-units/2.0/users_guide/framework_basics/faster_than_lightspeed_constants/">"Faster
             than lightspeed" constants</a>
         </td>
-        <td class="good">
+        <td class="best">
             <ul>
                 <li class="check">Constants as types</li>
                 <li class="check">Perfect conversion policy</li>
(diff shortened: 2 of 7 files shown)


geoffviola (approved):
Docs look great. Thanks for specifying what already exists and how to add new ones.

PR #343, opened 2024-12-06
Add how-to guide for `Quantity` template params

chiphogg:
These are documented already, but under the somewhat obscure heading of
"NTTP" inside of the Quantity docs.  I think it will be easier for end
users to discover this if we make a how-to guide, and if we name it
"quantity template parameters" --- a name that is more relatable and
easier to understand.

Follow-up to #316.


geoffviola (approved):
The documentations looks good. I like the details in the under the hood section and the example was nice.

PR #348, opened 2024-12-08
Add `STANDARD_GRAVITY` Constant

chiphogg:
It took me a little while to figure out exactly what I wanted to do
here.

Initially, I was going to _deprecate_ the standard gravity unit.  Then I
realized we need it to define certain other units, such as the
correspondence between "pounds" as a force and a mass --- and I didn't
want to introduce a circular dependency between `:units` and
`:constants`.

Then I was going to deprecate only the _quantity maker_,
`standard_gravity`.  But then I realized it would be weird to create a
situation where certain units have quantity makers, and certain other
ones don't.

In the end, I think the right move is to just provide standard gravity
as a new constant.  This is the preferred way to interact with it, but
the old ways will still be there for the foreseeable future.

Follow-on to #90.


chiphogg:
I should think this through some more.

Issue #353, opened 2024-12-10
MSVC19 - Spaceship Operator triggers "C2338	static_assert failed: 'NTTP functionality only works when rep is built-in integral type'


MatthiasJeuthe:
Hello Chip,

I noticed, that as soon as I export a class using the spaceship operator, I receive the error as mentioned in the title.
Minimal example: 

```
#define DLL_API __declspec(dllexport)

	struct DLL_API TestClass
	{
		auto operator<=>(const TestClass&) const = default;

		au::Quantity<au::Meters, double> thickness;
	};
```

It would be great, if you could take a look at it.

chiphogg:
Thanks for raising this.  Very timely, as we are on the verge of cutting the next release!

I was extremely confused, but I finally figured out what is happening.  It turns out that `operator<=>` can't be automatically defined for `Quantity`, because its data member is `private`.  Normally, we would have defined it ourselves as a hidden friend, but we maintain C++14 compatibility, and we literally can't express `operator<=>` in C++14.  However, `Quantity` can now be _implicitly converted_ to the enum class `Quantity::NTTP`, which presumably _does_ support `<=>`.  Of course, since `Quantity<Meters, double>` is of floating point type, we (correctly and safely) produce a hard compiler error when we trigger this conversion.

You can immediately unblock yourself by adding this function somewhere after you include `"au/quantity.hh"` (directly, or indirectly via `"au/au.hh"`), and before you define `TestClass`:

```cpp
template <typename U1, typename R1, typename U2, typename R2>
constexpr auto operator<=>(const Quantity<U1, R1> &lhs, const Quantity<U2, R2> &rhs) {
    using U = CommonUnitT<U1, U2>;
    return lhs.in(U{}) <=> rhs.in(U{});
}
```

On the Au side, I think we can make a kind of `"au/au_cpp20.hh"` file, with appropriate CI testing.  Users on newer standards can include it to get library-supported fixes to handle features in newer standards.  Users on older standards won't be affected at all.

chiphogg:
A colleague, @andrey-petr, pointed out that you could also change your return type from `auto` to `std::strong_ordering` in this case.  That would have the benefit of being a very simple change.  However, I think that `auto` better expresses author intent in this case, and probably would also scale better to structs with multiple members.

MatthiasJeuthe:
Thank you for your quick response! The workaround you suggested works perfectly. 
[...]

PR #355, opened 2024-12-11
Tighten NTTP creation: require `to_nttp(...)`

chiphogg:
This backpedals on the very low friction pathway to pass `Quantity` instances as template parameters.  This causes problems with C++20's spaceship operator `<=>`.  We could fix this with a special C++20 compatibility layer, and we might still do exactly that.  However, this is just one instance of a broader class of vulnerability.

It turns out that giving the very safe `Quantity` an **implicit** conversion to a raw numeric type may simply be inviting trouble.  This gives the compiler a pathway to generate all kinds of operations that we probably don't want to generate.

We _could_ just make the conversion _explicit_... but then, how would we trigger it?  I can't think of a way that would be at all comparably concise to `to_nttp`.

So let's just use `to_nttp` for now.  It will be a lot easier to loosen this later if we're confident that's correct, than to tighten it after we've cut an official release.

Helps #353: it gives a more sensible error.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -376,16 +376,11 @@ class Quantity {
                       "NTTP functionality only works when rep is built-in integral type");
     }
 
-    constexpr operator NTTP() const {
+    friend constexpr NTTP to_nttp(const Quantity &q) {
         static_assert(std::is_integral<Rep>::value,
                       "NTTP functionality only works when rep is built-in integral type");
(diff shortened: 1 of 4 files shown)


chiphogg:
I created this pull request to show, for posterity, how we would have migrated to a forced `to_nttp` interface.  I'm closing it now because I'm leaning towards keeping the current interface, but if that decision ever changes, we can reopen it.

PR #358, opened 2024-12-12
Prefer specific feature test macros

matthew-reynolds:
Rather than testing for the entire C++ standard revision using
`__cplusplus`, use specific feature test macros for testing the presence
of the particular features being used.

Testing the entire C++ standard revision is insufficient, as toolchains
may implement and enable features incrementally during implementation of
the next revision. A toolchain could feasibly implement three-way
comparisons but have not yet incremented `__cplusplus` as other features
remain unimplemented. [^1]

I also believe this approach to be somewhat more self-describing, which
is appealing even in the face of its verbosity/clunkiness.

[^1]: https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#explanation-and-rationale-for-the-approach

### DIFF
--- a/au/code/au/cpp20_test.cc
+++ b/au/code/au/cpp20_test.cc
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if __cplusplus >= 202002L
-
 #include <compare>
 
 #include "au/quantity.hh"
@@ -42,5 +40,3 @@ TEST(Quantity, SupportsSpaceship) { EXPECT_LT(Foo{5 * m}, Foo{6 * m}); }
 TEST(QuantityPoint, SupportsSpaceship) { EXPECT_LT(FooPt{meters_pt(5)}, FooPt{meters_pt(6)}); }
 
 }  // namespace au
-
-#endif
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -815,8 +815,7 @@ constexpr auto operator>=(QLike q1, Quantity<U, R> q2) -> decltype(as_quantity(q
     return as_quantity(q1) >= q2;
 }
 
-// Spaceship operator provides C++20 compatibility.
-#if __cplusplus >= 202002L
+#if defined(__cpp_impl_three_way_comparison) && __cpp_impl_three_way_comparison >= 201907L
 template <typename U1, typename R1, typename U2, typename R2>
 constexpr auto operator<=>(const Quantity<U1, R1> &lhs, const Quantity<U2, R2> &rhs) {
     using U = CommonUnitT<U1, U2>;
--- a/au/code/au/quantity_point.hh
+++ b/au/code/au/quantity_point.hh
@@ -370,8 +370,7 @@ constexpr auto operator-(QuantityPoint<U1, R1> p1, QuantityPoint<U2, R2> p2) {
     return detail::using_common_point_unit(p1, p2, detail::minus);
 }
 
-// Spaceship operator provides C++20 compatibility.
-#if __cplusplus >= 202002L
+#if defined(__cpp_impl_three_way_comparison) && __cpp_impl_three_way_comparison >= 201907L
 template <typename U1, typename R1, typename U2, typename R2>
 constexpr auto operator<=>(const QuantityPoint<U1, R1> &lhs, const QuantityPoint<U2, R2> &rhs) {
     using U = CommonPointUnitT<U1, U2>;


chiphogg (changes requested):
Thanks for this!

You should update the remaining `__cplusplus` instance, which I think is in the `cpp20_test.cc` file.  In fact... maybe we should just remove it?  Maybe the right move would be to specifically exclude any configurations that we _don't expect_ to support the feature.  We don't actually have such configurations, and likely never will, but I think this would leave us in a better place: right now, if it happened that _all_ configurations didn't support this (maybe we made a mistake with the macros), we would just silently not be running tests.

In retrospect, I wish I had taken this approach in the first place... but, it would have been more work, because we would have needed to _engineer_ a solution for gcc 10, and now we don't because with your change it will just work out of the box.  :slightly_smiling_face:

matthew-reynolds:
> You should update the remaining __cplusplus instance, which I think is in the cpp20_test.cc file.

I went back and forth on this, sounds like I should have included this diff after all :slightly_smiling_face:
```diff
diff --git a/au/code/au/cpp20_test.cc b/au/code/au/three_way_compare_test.cc
similarity index 88%
rename from au/code/au/cpp20_test.cc
rename to au/code/au/three_way_compare_test.cc
index 8469837..dce5caf 100644
--- a/au/code/au/cpp20_test.cc
+++ b/au/code/au/three_way_compare_test.cc
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if __cplusplus >= 202002L
+#if (defined(__has_include) && __has_include(compare)) && \
+     (defined(__cpp_lib_three_way_comparison_) && __cpp_lib_three_way_comparison >= 201711L)
 
 #include <compare>
 
```

> In fact... maybe we should just remove it? Maybe the right move would be to specifically exclude any configurations that we don't expect to support the feature. We don't actually have such configurations, and likely never will, but I think this would leave us in a better place: right now, if it happened that all configurations didn't support this (maybe we made a mistake with the macros), we would just silently not be running tests.
>
> In retrospect, I wish I had taken this approach in the first place... but, it would have been more work, because we would have needed to engineer a solution for gcc 10, and now we don't because with your change it will just work out of the box. :slightly_smiling_face:

Oh, is GCC 10 in the bucket of "Has three way compare but not the rest of C++20?" Gotcha. In that case, removing the checks altogether seems reasonable to me, maybe with a comment at the top of the file stating that all toolchains used for testing support three way comparisons even if they don't support the full C++20. I also think this is a good way to go, I will update the PR

matthew-reynolds:
>> Maybe the right move would be to specifically exclude any configurations that we don't expect to support the feature. We don't actually have such configurations, and likely never will
>
> Oh, is GCC 10 in the bucket of "Has three way compare but not the rest of C++20?" Gotcha.

I think I must have misunderstood. The non-C++20 configurations of all toolchains don't support this feature. Can you clarify what you're envisioning here?

chiphogg:
> > > Maybe the right move would be to specifically exclude any configurations that we don't expect to support the feature. We don't actually have such configurations, and likely never will
> > 
> > 
> > Oh, is GCC 10 in the bucket of "Has three way compare but not the rest of C++20?" Gotcha.
> 
> I think I must have misunderstood. The non-C++20 configurations of all toolchains don't support this feature. Can you clarify what you're envisioning here?

I just meant: remove all checks from the test file.  If a configuration isn't compatible, we'll exclude it manually.

We already have it set up in GH actions such that we won't run that test target unless `--std=c++20` is set.

matthew-reynolds:
> If a configuration isn't compatible, we'll exclude it manually.
> We already have it set up in GH actions such that we won't run that test target unless --std=c++20 is set.

Isn't this still opt-in? We've moved it up to the Bazel level with the `manual` tag, but it's still skipped by default and only included manually for particular configs. And since we don't have Bazel config settings for the standard version, we can't use `target_compatible_with` or similar.

I'm fine with dropping the checks from the test file and relying only on the Bazel target selection, it removes one level of redundancy and means accidentally including the file in the wrong test config will fail rather than silently passing, but I don't think this addresses "specifically exclude configurations that don't support the feature."

matthew-reynolds:
> I meant that if we had a toolchain that didn't have a `-std=c++20` mode, then we would need to tweak our GitHub Actions file to specifically exempt that toolchain from trying to run the C++20 tests.

Gotcha, I'm following now.

Sounds good. I will remove the feature checks from `cpp20_test.cc`. I'm also inclined to rename it `three_way_compare_test.cc` - Any objections?

chiphogg:
> > I meant that if we had a toolchain that didn't have a `-std=c++20` mode, then we would need to tweak our GitHub Actions file to specifically exempt that toolchain from trying to run the C++20 tests.
> 
> Gotcha, I'm following now.
> 
> Sounds good. I will remove the feature checks from `cpp20_test.cc`. I'm also inclined to rename it `three_way_compare_test.cc` - Any objections?

Thanks for asking --- I prefer the current name, because I want to limit the potential proliferation of special test files.  (There sure are a _lot_ of feature test macros!)  I'm fine only _running the tests_ on toolchains with full C++20 support; the important thrust of this PR is to improve support for _library end users_.  Also, if we provided concepts with the library someday, I'd want this file to be the clear home for those tests.

chiphogg (approved):
Thanks for making this accessible to a wider and more precise set of users!

chiphogg:
It's becoming an Au tradition to get a PR landed _immediately after_ a release is cut.  :grin:

PR #360, opened 2024-12-15
Add test client package

### LINKED ISSUE #359: au/static_cast_checkers.hh missing from CMakeLists.txt
I started messing around with releasing a new conan version, and got an unexpected build error.  Turns out that, yep, #347 did not update `CMakeLists.txt`.  I need to figure out why our CI didn't catch it.  Using CMake locally, everything does still pass.  I suspect it's related to having an _external project use Au as a dependency_.  We'll need to cut a new release for this ASAP.

chiphogg:
This makes sure that clients can use a version of Au that is installed
by CMake.

First, we add the new workflow, to make sure that it [breaks] in exactly
the way that reveals #359.  Then, in a new commit in this same PR, we
will fix the bug.

[breaks]: https://github.com/aurora-opensource/au/actions/runs/12341997471/job/34441316121?pr=360

Fixes #359.

### DIFF
--- a/au/code/au/CMakeLists.txt
+++ b/au/code/au/CMakeLists.txt
@@ -39,6 +39,7 @@ header_only_library(
     quantity.hh
     quantity_point.hh
     rep.hh
+    static_cast_checkers.hh
     unit_of_measure.hh
     unit_symbol.hh
     wrapper_operations.hh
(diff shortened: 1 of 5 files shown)


au/code/au/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ header_only_library(
     quantity.hh
     quantity_point.hh
     rep.hh
+    static_cast_checkers.hh
```
chiphogg, line 42: The changes in this file are the ones that fix the actual bug.

PR #362, opened 2024-12-18
Add `common_unit` and `make_common` utilities

chiphogg:
We also add the "common point" versions of these.

The absence of `common_unit` and `common_point_unit` was mostly just an
oversight.  I recently tried to call one in Godbolt, and got surprised
when I couldn't.  And it turned out that I had already written this
function anyway, except that it was hidden away inside of a test file.

As I thought about these functions more, I thought it was interesting
that the inputs are unit slots, so we could have a wide variety of
things (simple units, quantity makers, unit symbols, etc.), but the
output would always be a simple unit.  Wouldn't it be nice if we could
also combine `meters` and `feet`, or `m` and `ft`, and have the output
act like its inputs?  This idea became the `make_common` and
`make_common_point` utilities.

Docs included (and rendered and tested).

### DIFF
--- a/au/code/au/unit_of_measure_test.cc
+++ b/au/code/au/unit_of_measure_test.cc
@@ -36,12 +36,6 @@ using ::testing::StaticAssertTypeEq;
 using ::testing::StrEq;
 
 namespace au {
-namespace {
-template <typename... Us>
-constexpr auto common_unit(Us...) {
-    return CommonUnitT<Us...>{};
-}
-}  // namespace
 
 struct Celsius : Kelvins {
     static constexpr auto origin() { return milli(kelvins)(273'150); }
@@ -49,6 +43,7 @@ struct Celsius : Kelvins {
 };
 constexpr const char Celsius::label[];
 constexpr auto celsius = QuantityMaker<Celsius>{};
+constexpr auto celsius_pt = QuantityPointMaker<Celsius>{};
 
 struct AlternateCelsius : Kelvins {
     static constexpr auto origin() { return micro(kelvins)(273'150'000); }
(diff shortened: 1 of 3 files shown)


au/code/au/unit_of_measure_test.cc
```diff
@@ -555,6 +554,40 @@ TEST(CommonPointUnit, UnpacksTypesInNestedCommonUnit) {
     StaticAssertTypeEq<Common, CommonPointUnitT<W, X, Y, Z>>();
 }
 
+TEST(CommonPointUnit, SupportsUnitSlots) {
+    StaticAssertTypeEq<decltype(common_point_unit(kelvins_pt, celsius_pt)),
+                       CommonPointUnitT<Kelvins, Celsius>>();
+}
+
+TEST(MakeCommon, PreservesCategory) {
+    constexpr auto feeters = make_common(feet, meters);
```
geoffviola, line 563: > feeters

😆

docs/reference/unit.md
```diff
@@ -603,9 +619,85 @@ A specialization will only exist if the inputs are all units, and will exist but
...
+    constexpr auto x = make_common(meters, feet)(18);
+
+    using symbols::m;
+    using symbols::ft;
+    constexpr auto y = 9.5f * make_common(m, ft);
+    ```
+
+### Making common point units {#make-common-point}
+
+**Result:** A new unit: the largest-magnitude, highest-origin unit which is "common" to the input
```
geoffviola, line 690: In [the test](https://github.com/aurora-opensource/au/pull/362/files#diff-a4df4f6d28089aabc817f1a4fb8860b4ae7efdeebe90424039467d7478685400R576) it says lowest origin. Should this say lowest?

docs/reference/unit.md
```diff
@@ -603,9 +619,85 @@ A specialization will only exist if the inputs are all units, and will exist but
...
+**Result:** A new unit: the largest unit that evenly divides its input units.  (Read more about the
+concept of [common units](../discussion/concepts/common_unit.md).)
+
+**Syntax:**
+
+- `make_common(us...)`
+
+??? example "Examples"
+    ```cpp
+    constexpr auto x = make_common(meters, feet)(18);
```
geoffviola, line 681: Instead of `auto`, can we be explicit about the type so that the category is more obvious?

geoffviola (approved):
Implementation, tests, and docs look good. Just left some minor comments.

docs/reference/unit.md
chiphogg: Good point about the lack of obviousness... but I don't think making the type explicit is the way to go.  It's really in the way we _use the result_ that we can see the effects.

I added some explanatory comments, as I think that was the clearest way to get the idea across.

docs/reference/unit.md
chiphogg: No.  :sweat_smile:  But it's really confusingly worded.

To be "common" (for points) to all input units, a unit's origin must _not be higher_ than any of them.  The unit with the _highest_ such origin will have its origin _equal_ to the _lowest origin_ of all input units.

The fix here was to simply abandon this phrasing and rephrase it more carefully.  :slightly_smiling_face:

PR #366, opened 2024-12-19
Improve error message for `Quantity` in unit slot

### LINKED ISSUE #327: Add error, with link, for passing Quantity to unit slot
Imagine trying to round some size to the nearest 10 bytes:

```cpp
auto size = bytes(1234);
size = round_as<int>(bytes * mag<10>(), size);
```

This works (it gives `bytes(1230)`).  _However_, users might try writing something like this:

```cpp
auto size = bytes(1234);
size = round_as<int>(bytes(10), size); // (Incorrect)
```

If so, they might get an error like this:

image

That's not the best.  It _does_ say that `IsUnit<...Quantity...>` isn't satisfied, but it's not as straightforward as it could be.

We could instead turn `AssociatedUnit<Quantity<...>>` into a hard compiler error, with a readable message.  Something like this:

image

That seems better: we can directly tell them that they're passing a Quantity to a unit slot, and link them to documentation on unit slots.  Actually, probably even better would be to make a new entry in the troubleshooting page.

chiphogg:
This is an easy mistake to make for users who haven't yet grasped the
idioms of the library.  Now, if they do this, they'll get a nice,
readable error, which directs them to the new section of the
troubleshooting doc (also added in this PR) that explains what is going
wrong, and what to do about it.

Fixes #327.

### DIFF
--- a/au/code/au/quantity.hh
+++ b/au/code/au/quantity.hh
@@ -424,6 +424,22 @@ class Quantity {
     Rep value_{};
 };
 
+// Give more readable error messages when passing `Quantity` to a unit slot.
+template <typename U, typename R>
+struct AssociatedUnit<Quantity<U, R>> {
+    static_assert(
+        detail::AlwaysFalse<U, R>::value,
+        "Can't pass `Quantity` to a unit slot (see: "
+        "https://aurora-opensource.github.io/au/main/troubleshooting/#quantity-to-unit-slot)");
+};
+template <typename U, typename R>
+struct AssociatedUnitForPoints<Quantity<U, R>> {
+    static_assert(
+        detail::AlwaysFalse<U, R>::value,
+        "Can't pass `Quantity` to a unit slot for points (see: "
+        "https://aurora-opensource.github.io/au/main/troubleshooting/#quantity-to-unit-slot)");
+};
+
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Machinery to explicitly unblock integer division.
 //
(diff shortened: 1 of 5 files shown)


au/code/au/quantity.hh
```diff
@@ -424,6 +424,22 @@ class Quantity {
     Rep value_{};
 };
 
+// Give more readable error messages when passing `Quantity` to a unit slot.
+template <typename U, typename R>
+struct AssociatedUnit<Quantity<U, R>> {
+    static_assert(
+        detail::AlwaysFalse<U, R>::value,
+        "Can't pass `Quantity` to a unit slot (see: "
+        "https://aurora-opensource.github.io/au/main/troubleshooting/#quantity-to-unit-slot)");
```
hoffbrinkle, line 433: Do we have any way to ensure that the link remains valid?

hoffbrinkle (approved):

au/code/au/quantity.hh
chiphogg: Good question!

The simple answer is "not directly".  But one thing we _can_ do is to add this same link _somewhere else_... or at least, the relative parts of the URL.  This will mean that if we ever change the anchor name, then `au-docs-serve` will throw up errors like the following, and we're likely to notice it at some point and fix the docs.

image

Incidentally... this screenshot shows actual doc errors that I had introduced in this PR.  :flushed:  So, I fixed those too.

And then I added a new section to the "unit slot" docs which points back to the troubleshooting guide, which feels like it "should have been there all along", and --- more to the point here --- which also uses the URL from these `static_assert` calls, giving it the best _partial_ coverage that I know how to get.

I'm not planning to ask for a re-review, but if you spot anything we should fix, we can always do a quick follow on PR.

hoffbrinkle (approved):

PR #371, opened 2024-12-30
Remove ad hoc constant from unit slot example.

hoffbrinkle:
When demonstrating that we can pass `Constant` to a unit slot, use the available
builtin constant for the speed of light rather than creating an ad hoc
constant.


geoffviola (approved):
Changes look good. This example is more readable and copyable. The changes match the suggestion made out of band.

PR #372, opened 2024-12-31
Update documentation on ad hoc constant limitations.

hoffbrinkle:
Ad hoc constants no longer have significant limitations with unit labels.

### DIFF
--- a/docs/reference/constant.md
+++ b/docs/reference/constant.md
@@ -130,23 +130,41 @@ The main advantage of doing this is its conciseness: the constant definition is 
 line.  The built constant also has all of the multiplication and division operators types that
 `Constant` supports, as well as its perfect conversion policy to any `Quantity` type.
 
-The only disadvantage is the unit label for an ad hoc `Constant` will not be specific to that
-`Constant`.
+The only disadvantage is that an ad hoc `Constant` gets an ad hoc unit symbol, instead of a simple,
+concise symbol.  The example below demonstrates the difference.
 
-```cpp
-    constexpr auto c = make_constant(meters / second * mag<299'792'458>());
+??? example "Example: ad hoc constants get ad hoc labels"
+    Let's take the following speed, and express it in terms of the speed of light, using a constant:
 
+    ```cpp
     constexpr auto v = (miles / hour)(65.0);
+    ```
+
+    First, an ad hoc constant:
+
+    ```cpp
+    constexpr auto c = make_constant(meters / second * mag<299'792'458>());
 
     std::cout << v.as(c) << std::endl;
     // Output:
     // "9.69257e-08 [299792458 m / s]"
+    ```
 
+    Note how the program is correct, and the label is accurate, but clunky: it prints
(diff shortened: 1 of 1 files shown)


docs/reference/constant.md
```diff
@@ -130,12 +130,23 @@ The main advantage of doing this is its conciseness: the constant definition is
 line.  The built constant also has all of the multiplication and division operators types that
 `Constant` supports, as well as its perfect conversion policy to any `Quantity` type.
 
-The only disadvantage is the missing label, which will make printed quantities hard to understand
-because the constant will be represented as `[UNLABELED_UNIT]` in the compound label.
+The only disadvantage is the unit label for an ad hoc `Constant` will not be specific to that
+`Constant`.
```
chiphogg, line 134:
```suggestion
The only disadvantage is that an ad hoc `Constant` gets an ad hoc unit symbol, instead of a simple,
concise symbol.  The example below demonstrates the difference.
```

I thought this read a little more clearly.

docs/reference/constant.md
```diff
@@ -130,12 +130,23 @@ The main advantage of doing this is its conciseness: the constant definition is
...
+
+    std::cout << v.as(c) << std::endl;
+    // Output:
+    // "9.69257e-08 [299792458 m / s]"
+
+    // Found in `"au/constants/speed_of_light.hh"`:
+    std::cout << v.as(SPEED_OF_LIGHT) << std::endl;
+    // Output:
+    // "9.69257e-08 c"
+```
```
chiphogg, line 149:
```suggestion
??? example "Example: ad hoc constants get ad hoc labels"
    Let's take the following speed, and express it in terms of the speed of light, using a constant:

    ```cpp
    constexpr auto v = (miles / hour)(65.0);
    ```

    First, an ad hoc constant:

    ```cpp
    constexpr auto c = make_constant(meters / second * mag<299'792'458>());

    std::cout << v.as(c) << std::endl;
    // Output:
    // "9.69257e-08 [299792458 m / s]"
    ```

    Note how the program is correct, and the label is accurate, but clunky: it prints
    `[299792458 m / s]` instead of a concise label such as `c`.

    By contrast, we can use a fully defined constant.  This is generally a little more effort, but
    some constants, such as `SPEED_OF_LIGHT`, are included out of the box:

    ```cpp
    // Found in `"au/constants/speed_of_light.hh"`:
    std::cout << v.as(SPEED_OF_LIGHT) << std::endl;
    // Output:
    // "9.69257e-08 c"
    ```

    There's no difference in the program that gets executed, but the printed label is a lot easier
    to understand.
```

I love this example!  Thanks for adding it.  I decided to turn it into an expandable Example admonition, and add some narrative flow.

chiphogg (approved):
Thanks for catching this!  This is good to land once the latest main gets merged in.  I've included two optional tweaks as well.

chiphogg:
> Thanks for catching this! This is good to land once the latest main gets merged in. I've included two optional tweaks as well.

Apparently, I'm able to merge the latest main myself.  I'll go ahead and land this after applying the suggestions.

PR #374, opened 2025-01-15
Support `SingularNameFor` in unit slots

### LINKED ISSUE #373: Support `SingularNameFor` in (quantity) unit slots
This doesn't work, but it should (https://godbolt.org/z/bY3r1bKzc):

```cpp
constexpr auto ONE_REV = make_constant(revolution);
QuantityD<Radians> two_pi_rad = ONE_REV;
```

chiphogg:
I thought it through, and realized this is pretty clearly for "quantity"
type slots only, not "point" type slots.  The reason this exists is to
participate in multiplications, divisions, and powers with quantity
makers.  Quantity points don't have multiplicative semantics.

Fixes #373.

### DIFF
--- a/au/code/au/unit_of_measure.hh
+++ b/au/code/au/unit_of_measure.hh
@@ -430,6 +430,10 @@ struct SingularNameFor {
     }
 };
 
+// Support `SingularNameFor` in (quantity) unit slots.
+template <typename U>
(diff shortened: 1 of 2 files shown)


geoffviola (approved):
Example, test, and implementation look good.

PR #376, opened 2025-01-15
Support powers and roots for unit wrappers

### LINKED ISSUE #375: Constant and SymbolFor do not support rational powers
We should be able to write `squared(c)`, but we currently can't.

chiphogg:
This unlocks some basic missing functionality for `Constant` and
`SymbolFor`: namely, we should be able to pass instances of them to
`squared()`, `sqrt()`, `pow<N>()`, `root<N>()`, and so on.

Fixes #375.

### DIFF
--- a/au/code/au/wrapper_operations_test.cc
+++ b/au/code/au/wrapper_operations_test.cc
@@ -33,6 +33,7 @@ struct UnitWrapper : MakesQuantityFromNumber<UnitWrapper, Unit>,
                      ScalesQuantity<UnitWrapper, Unit>,
                      ComposesWith<UnitWrapper, Unit, UnitWrapper, UnitWrapper>,
                      ComposesWith<UnitWrapper, Unit, QuantityMaker, QuantityMaker>,
+                     SupportsRationalPowers<UnitWrapper, Unit>,
                      CanScaleByMagnitude<UnitWrapper, Unit> {};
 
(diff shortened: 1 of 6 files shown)


au/code/au/wrapper_operations_test.cc
```diff
@@ -144,18 +145,22 @@ TEST(CanScaleByMagnitude, MakesScaledWrapperWhenDividingByMagnitude) {
...
+    constexpr auto mol = UnitWrapper<Moles>{};
+    StaticAssertTypeEq<decltype(pow<3>(mol)), UnitWrapper<decltype(pow<3>(Moles{}))>>();
+}
+
+TEST(SupportsRationalPowers, EnablesTakingRoots) {
+    constexpr auto mol = UnitWrapper<Moles>{};
+    StaticAssertTypeEq<decltype(root<8>(mol)), UnitWrapper<decltype(root<8>(Moles{}))>>();
+}
 
-    // UnitWrapper<Meters>{} * meters_pt;
```
geoffviola, line 150: Q: Are these tests irrelevant now?

geoffviola (approved):
Motivation, implementation, and tests look good.

au/code/au/wrapper_operations_test.cc
chiphogg: Oops... actually, no!  We deleted `ForbidsComposingWith` in #227, so I thought we forgot to remove this, but reading that PR summary I see that leaving these here was intentional.

PR #377, opened 2025-01-15
Add unit test for XKCD 3038 alt-text

chiphogg:
See for context: <https://xkcd.com/3038/>.  Presumably, the joke only
works if the speed limit in question corresponds to a realistic speed
limit for American roads, so we take the range to be from 25 MPH (many
residential streets) to 75 MPH (some roads in places I have seen such as
parts of MI or TX).

To make this work, it was convenient to add unit definitions for
`Arcminutes` and `Arcseconds`.  These have various choices for unit
symbol, but I decided to go with `am` and `as` because this will be the
most consistent with users who form symbols for `mas` and `uas`.

For the labels, in any case, I went with `'` and `"`.  I could be
persuaded to go with `''` for arcseconds, but I think `"` is likely to
be better overall.

We also now group all of our XKCD test cases into an `Xkcd` test family,
because of course we do.

### DIFF
--- a/au/code/au/CMakeLists.txt
+++ b/au/code/au/CMakeLists.txt
@@ -59,6 +59,10 @@ header_only_library(
     stdx/utility.hh
     units/amperes.hh
     units/amperes_fwd.hh
+    units/arcminutes.hh
+    units/arcminutes_fwd.hh
+    units/arcseconds.hh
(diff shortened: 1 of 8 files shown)


geoffviola (approved):
😂

Issue #255, opened 2024-07-03
Prevent CMake targets from going stale.


chiphogg:
Soon, we plan to support CMake (#215).  We won't use globbing, because that's a can of worms.  But several bazel targets _do_ use globbing, so we need a way to make sure the CMake build doesn't go stale.

Maybe there's some GitHub action where the user will have to check a box to land the PR.  And this blocking box can be triggered on:

- Any file being added in a directory, like `units`, that we glob.
- Any `BUILD` file being edited.

The latter would makes this issue more expansive than globs, so that it covers keeping CMake up to date generally.

PR #379, opened 2025-01-19
Add script to compare core bazel and CMake headers

chiphogg:
This gives us the most important portion of #255.  In fact, it's even
better than what we originally suggested, which was a blocking checkbox
in code review.  This solution uses `bazel query` to get the
authoritative set of headers included in our main targets, `//au` and
`//au:io`.  It compares them to the CMake target `au`, which encompasses
both of these targets.

The script returns an exit code of `0` if the headers are an exact
match, and `1` otherwise.  If any files differ, it prints out the list
of every file that is found in only bazel or only CMake.

A good way to test this is to run:

```sh
check-cmake-headers && echo $?
```

This should print `0` on the last line.  Then, if we edit
`au/code/au/CMakeLists.txt`, and _change the name_ of one of the files
(say, prepend `asdf`), we should see the discrepancy printed out, and
the printed exit code should change to `1`.

A future PR will add a GitHub action to call this script and gate PRs on
its result.


chiphogg:
For proof that the script _will actually work in CI_, see [this passed build](https://github.com/aurora-opensource/au/actions/runs/12856948865) for 1805ffa33e0d657cc403f20b1a93f44c4126ab23 (which temporarily made it gating), and [this failed build](https://github.com/aurora-opensource/au/actions/runs/12856959692) for 82e6727ff3d908a844046627a08d42d9dbc42f77 (which intentionally created an error condition).

tools/bin/check-cmake-headers
```diff
@@ -0,0 +1,127 @@
...
+    )
+    return 0 if ok else 1
+
+
+def _get_cmake_headers():
+    # Ensure that CMake has been set up appropriately.
+    subprocess.run(
+        [
+            "cmake",
+            "-S",
```
hoffbrinkle, line 35: I tend to lean towards long arguments when using scripts to make them more self documenting.  Not being a regular cmake user I have no idea what `-B` and `-S` do.  Well, I have have a decent guess as to `-S`.

hoffbrinkle (approved):

tools/bin/check-cmake-headers
chiphogg: Interestingly, `cmake --help` didn't seem to suggest a long-form alternative!

```
Usage

  cmake [options] <path-to-source>
  cmake [options] <path-to-existing-build>
  cmake [options] -S <path-to-source> -B <path-to-build>

Specify a source directory to (re-)generate a build system for it in the
current working directory.  Specify an existing build directory to
re-generate its build system.

Options
  -S <path-to-source>          = Explicitly specify a source directory.
  -B <path-to-build>           = Explicitly specify a build directory.
  -C <initial-cache>           = Pre-load a script to populate the cache.
  -D <var>[:<type>]=<value>    = Create or update a cmake cache entry.
  -U <globbing_expr>           = Remove matching entries from CMake cache.
...
```

Anyway, I don't know them from memory either, but this command was basically pasted from our [installation instructions](https://aurora-opensource.github.io/au/main/install/#__tabbed_1_2).  I've added a link here in a comment; thanks for pointing it out!

PR #381, opened 2025-01-21
Gate PRs on CMake and bazel headers matching

### LINKED ISSUE #255: Prevent CMake targets from going stale.

chiphogg:
This will prevent future issues similar to #359 from occurring in the
first place.

Fixes #255.


timhirsh (approved):
LGTM 👍

Issue #383, opened 2025-02-20
Unit-Safe Implementation of `std::hypot`


justin-kottinger:
Please consider the following scenario:
```cpp
struct PointR2 {
    au::QuantityD<au::Meters> x{au::ZERO};
    au::QuantityD<au::Meters> y{au::ZERO};
};

// ...

auto distance_to(const PointR2 & point_1, const PointR2 & point_2) -> au::QuantityD<au::Meters> {
  return au::meters(std::hypot((point_2.x - point_1.x).in(au::meters),
                               (point_2.y - point_1.y).in(au::meters)));
}
```
It is sub-optimal to remove unit-safety, call `std::hypot`, and then immediately go back to `au` land. Instead, it would be great if we could do something like the following:
```cpp
auto distance_to(const PointR2 & point_1, const PointR2 & point_2) -> au::QuantityD<au::Meters> {
  return au::hypot(point_2.x - point_1.x, point_2.y - point_1.y);
}
```
There are probably some intricacies in determining the return type of something like:
```cpp
const auto val_1{au::meters(1.0)};
const auto val_2{au::feet(1.0)};
const auto some_value{au::hypot(val_1, val_2)};
```
But it also might not matter as long as it returns an `au::Quantity`. One could always do `some_value.in(au::meters)` or `some_value.in(au::feet)` if / when they need to. 

Any chance we can get this feature into AU?

chiphogg:
Yep, seems like a good addition.  I'll pencil it in for the next release.  Thanks!

PR #384, opened 2025-02-21
Unit-Safe Implementation of `std::hypot`

justin-kottinger:
## Summary
Implements a unit-safe version of `std::hypot` within the AU library. This implementation ensures dimensional consistency when calculating the hypotenuse/Euclidean norm while preserving the underlying units.

## Implementation Details
- Extends the existing AU library with a unit-safe wrapper around `std::hypot`
- Maintains dimensional analysis correctness throughout calculations
- Returns results in the common unit and underlying representation as the input parameters

Related to #383

### DIFF
--- a/au/code/au/math.hh
+++ b/au/code/au/math.hh
@@ -182,15 +182,6 @@ auto hypot(Quantity<U1, R1> x, Quantity<U2, R2> y)
     return make_quantity<U>(std::hypot(x.in(U{}), y.in(U{})));
 }
 
-template <typename U1, typename R1, typename U2, typename R2>
-auto hypot(QuantityPoint<U1, R1> x, QuantityPoint<U2, R2> y)
-{
-    using U = CommonPointUnitT<U1, U2>;
-    using R = std::common_type_t<R1, R2>;
-    using ResultT = QuantityPoint<U, R>;
-    return make_quantity_point<U>(std::hypot(x.in(U{}), y.in(U{})));
-}
-
 // Copysign where the magnitude has units.
 template <typename U, typename R, typename T>
 constexpr auto copysign(Quantity<U, R> mag, T sgn) {
--- a/au/code/au/math.hh
+++ b/au/code/au/math.hh
@@ -174,8 +174,7 @@ constexpr auto clamp(QuantityPoint<UV, RV> v,
 }
 
 template <typename U1, typename R1, typename U2, typename R2>
-auto hypot(Quantity<U1, R1> x, Quantity<U2, R2> y)
-{
+auto hypot(Quantity<U1, R1> x, Quantity<U2, R2> y) {
     using U = CommonUnitT<U1, U2>;
     return make_quantity<U>(std::hypot(x.in(U{}), y.in(U{})));
 }
--- a/au/code/au/math.hh
+++ b/au/code/au/math.hh
@@ -33,6 +33,7 @@ using std::abs;
 using std::copysign;
 using std::cos;
 using std::fmod;
+using std::hypot;
 using std::isnan;
 using std::max;
 using std::min;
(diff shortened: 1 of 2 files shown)


au/code/au/math_test.cc
```diff
@@ -159,6 +159,60 @@ TEST(clamp, SupportsZeroForMultipleArguments) {
...
+    expect_consistent_with_std_hypot(0.0, 0.0);
+    expect_consistent_with_std_hypot(1.0, 0.0);
+    expect_consistent_with_std_hypot(2.0, 0.0);
+    expect_consistent_with_std_hypot(4.0, 2.0);
+}
+
+TEST(hypot, QuantityProducesResultsInCommonUnitOfInputs) {
+    EXPECT_THAT(hypot(centi(meters)(30), milli(meters)(400)),
+                SameTypeAndValue(milli(meters)(500.0)));
+}
```
chiphogg, line 187: There's a really cute test case you could consider adding here, that takes advantage of the pythagorean triple (5, 12, 13), and the fact that 1 ft == 12 in.  We could justify it by adding coverage for a non-`double` rep:

```suggestion

    EXPECT_THAT(hypot(inches(5.f), feet(1.f)), SameTypeAndValue(inches(13.f)));
}
```

chiphogg (changes requested):
Thanks for making this contribution!  I'll re-review once you get the CLA signed, and remove the `QuantityPoint` overloads.

au/code/au/math.hh
```diff
@@ -172,6 +173,15 @@ constexpr auto clamp(QuantityPoint<UV, RV> v,
     return (v < lo) ? ResultT{lo} : (hi < v) ? ResultT{hi} : ResultT{v};
 }
 
+template <typename U1, typename R1, typename U2, typename R2>
+auto hypot(Quantity<U1, R1> x, Quantity<U2, R2> y)
+{
+    using U = CommonUnitT<U1, U2>;
+    using R = std::common_type_t<R1, R2>;
+    using ResultT = Quantity<U, R>;
```
chiphogg, line 181:
```suggestion
```

It doesn't look like we're actually using either of these.

chiphogg (approved):
This LGTM with just one more small change needed.  (I spotted the change manually, and then in running the jobs, it looks like it is in fact causing failures.)

You'll also need to sign the CLA.  Let me know if you're having any difficulties with that, and I'll try to help troubleshoot!

justin-kottinger:
Thanks for the speedy review. Regarding the CLA, I signed the corporate version on Friday. It should be under my name with the company name of "Swarmbotics AI". It does not look like [CLAassistant](https://github.com/CLAassistant) is finding it. Do I need to sign the individual one as well?

chiphogg:
> Thanks for the speedy review. Regarding the CLA, I signed the corporate version on Friday. It should be under my name with the company name of "Swarmbotics AI". It does not look like [CLAassistant](https://github.com/CLAassistant) is finding it. Do I need to sign the individual one as well?

I have reached out internally to the folks who manage our CLA assistant.  I hope we can get this cleared up pretty soon.

timhirsh:
Hi @justin-kottinger, thanks for your contribution here!  For corporate CLA's, our options are to manually allowlist either:
- a public GitHub organization.  this would allow all public members of the organization to contribution to Aurora Opensource on behalf of the corporate CLA that you signed earlier.
- individual usernames of members of the organization "Swarmbotics AI"

Does "Swarmbotics AI" have a public GitHub organization that we can list here?  That would be preferable over maintaining a list of usernames.  Thanks!

chiphogg:
Also, note that while looking into this internally, we found that you're the first contributor to use the corporate CLA.  :smile:  So thanks for helping us work out the kinks, and I hope it explains why there _are_ any kinks!

justin-kottinger:
Thank you @timhirsh and @chiphogg for your help on this. My team generally prefers to keep our memberships private for security reasons. However, I have temporarily made my membership public. All tests are passing now. :tada:  Please let me know if you need anything else.

PR #389, opened 2025-04-05
Implement `cbrt()`

### LINKED ISSUE #388: Request support for more math functions.
Please add support for some of the less-common math functions, including, `std::cbrt`.

chiphogg:
Additionally, I noticed that we had a commented-out "hard" test case for
`sqrt`, which we have now satisfied for some time.  I uncommented that
test case in this same commit so we can start taking credit for it.

Fixes #388.

### DIFF
--- a/au/code/au/math_test.cc
+++ b/au/code/au/math_test.cc
@@ -78,14 +78,6 @@ TEST(cbrt, OutputRepDependsOnInputRep) {
     EXPECT_THAT(cbrt(cubed(meters)(8.L)), QuantityEquivalent(meters(2.L)));
 }
 
-TEST(cbrt, MixedUnitsSupportedWithCasting) {
-    constexpr auto x_in = inches(1);
-    constexpr auto y_cm = centi(meters)(2.54);
-    constexpr auto z_mm = milli(meters)(25.4);
-
-    EXPECT_THAT(cbrt(x_in * y_cm.as(inches) * z_mm.as(inches)), IsNear(x_in, nano(meters)(1)));
-}
-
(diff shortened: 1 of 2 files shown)


au/code/au/math_test.cc
```diff
@@ -70,6 +71,36 @@ TEST(abs, SameAsStdAbsForNumericTypes) {
...
+    EXPECT_EQ(cbrt(1.f), std::cbrt(1.f));
+    EXPECT_EQ(cbrt(1.L), std::cbrt(1.L));
+}
+
+TEST(cbrt, CanConvertIfConversionFactorRational) {
+    const auto geo_mean_length = cbrt(inches(1) * meters(1) * yards(1));
+
+    // Using Quantity-equivalent Unit just retrieves the stored value.
+    const auto retrieved_value = geo_mean_length.in(cbrt(inch * meter * yards));
+    EXPECT_THAT(retrieved_value, SameTypeAndValue(1.0));
```
hoffbrinkle, line 93: If it's just retrieving the stored value, does the value really come out as a `double`?

chiphogg: I noticed that too (this test is a copy-and-modify of the corresponding `sqrt` one).  Then I realized that it's getting passed through `std::cbrt`, which is why it becomes a `double`.

The comment is unclear, though.  At first glance, it makes it seem like the "stored" value should be `1 * 1 * 1`, but we're actually referring to the value stored in the result of `cbrt` (i.e., `geo_mean_length`).  I'll update all the relevant comments.

hoffbrinkle (approved):

PR #393, opened 2025-04-24
Add CI build with `-Wsign-conversion`

chiphogg:
We expect this to expose some violations of this warning.  We'll fix
those violations, and then we'll land this CI change to keep them fixed.


timhirsh (approved):
👍

Issue #392, opened 2025-04-23
`-Werror=sign-conversion` interference.


axelatharva:
I am working on a C++ project with the following compilation flags enabled : `-pedantic, -Wall, -Werror, -Wextra, -Wcast-align, -Wcast-qual, -Wdate-time, -Wdouble-promotion, -Wlogical-op, -Wmissing-include-dirs, -Wpacked, -Wpointer-arith, -Wredundant-decls, -Wold-style-cast, -Woverloaded-virtual, -Wshadow, -Wsign-conversion, -Wunused, -Wuseless-cast, -Wvla, -Wnon-virtual-dtor, -Wconversion, -Wduplicated-branches, -Wduplicated-cond, -Wimplicit-fallthrough, -Wmisleading-indentation, -Wnull-dereference`

If I just include a header file from `Au` and don't even use it (in this example `#include "au/units/volts.hh"`)
I get the compilation error - 

```
In file included from C:/.../MyProject/.build/_deps/au-src/au/code/au/magnitude.hh:25,
                 from C:/.../MyProject/.build/_deps/au-src/au/code/au/apply_rational_magnitude_to_integral.hh:19,
                 from C:/.../MyProject/.build/_deps/au-src/au/code/au/apply_magnitude.hh:17,
                 from C:/.../MyProject/.build/_deps/au-src/au/code/au/quantity.hh:19,
                 from C:/.../MyProject/.build/_deps/au-src/au/code/au/units/volts.hh:20,
                 from C:/.../MyProject/src/main.cpp:17:
C:/.../MyProject/.build/_deps/au-src/au/code/au/utility/string_constant.hh:253:59: error: conversion to 'long long unsigned int' from 'long long int' may change the sign of the result [-Werror=sign-conversion]
  253 |                     UIToA<static_cast<uint64_t>((N) >= 0) ? N : -N>::value);
      |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
C:/.../MyProject/.build/_deps/au-src/au/code/au/magnitude.hh:252:57: error: conversion to 'long long int' from 'uintmax_t' {aka 'long long unsigned int'} may change the sign of the result [-Werror=sign-conversion]
  252 |     using type = MagProductT<Magnitude<Pow<Prime<base>, power>>, PrimeFactorizationT<remainder>>;
      |                                                         ^~~~~
[...]

PR #394, opened 2025-04-25
Resolve all `-Wsign-conversion` errors

chiphogg:
The goal was to get this command to pass cleanly:

```sh
bazel build \
  --copt="-Werror" \
  --copt="-Wsign-conversion" \
  --config=gcc10 \
  //release/...:all \
  //au/...:all
```

The biggest change was to `string_constant`: we needed to fix up our
strategy for taking the absolute value of a number.  It was cleanest to
write a new helper, `abs_as_unsigned(T)`, which carefully and
efficiently handles all values (even the troublesome minimum
representable value, _whose negative can't be represented in `T`_). So
we're not just becoming warning-clean here; we're also handling values
that we couldn't handle before.  I found this cleaned up the
implementation a bit, too (no more `x = -x;`; no more ternary).

All of these new utilities got unit tests, to build confidence.

For the `unit_of_measure.hh` change, the right move was to change the
interface to accept `std::intmax` directly.  After all, literally the
only thing we do with `Deg` is to pass it to `UnitPowerT`, which takes
`std::intmax_t`.

For `magnitude.hh`, we simply insert an explicit static cast.  This is
very safe in this case, because `power` is the _log_ of a representable
number: it simply can't be very big at all.

Helps #392.  After this PR lands, #393 should become landable.

### DIFF
--- a/au/code/au/utility/string_constant.hh
+++ b/au/code/au/utility/string_constant.hh
@@ -15,7 +15,10 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 #include <utility>
+
+#include "au/stdx/type_traits.hh"
 
 namespace au {
 namespace detail {
@@ -87,6 +90,22 @@ struct IToA;
 // Implementation details below.
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// The absolute value of a signed integer, as an unsigned integer.
+//
+// This handles the special case where the lowest `int64_t` cannot be directly negated: it would be
+// too big to fit into `int64_t`.
+template <typename S>
+constexpr std::make_unsigned_t<S> abs_as_unsigned(S x) {
+    static_assert(stdx::conjunction<std::is_integral<S>, std::is_signed<S>>::value,
+                  "Only designed for signed integral types");
+    using U = std::make_unsigned_t<S>;
+    constexpr auto SMAX = static_cast<U>(std::numeric_limits<S>::max());
+    constexpr auto UMAX = std::numeric_limits<U>::max();
+
+    auto result = static_cast<U>(x);
+    return (result > SMAX) ? (UMAX - result + 1u) : result;
+}
+
 // The string-length needed to hold a representation of this unsigned integer.
 constexpr std::size_t string_size_unsigned(uint64_t x) {
     std::size_t digits = 1;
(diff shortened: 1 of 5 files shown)


au/code/au/utility/test/string_constant_test.cc
```diff
@@ -50,6 +64,13 @@ TEST(IToA, ValueHoldsStringVersionOfTemplateParameter) {
 
     EXPECT_STREQ(IToA<-1>::value.c_str(), "-1");
     EXPECT_STREQ(IToA<-83294>::value.c_str(), "-83294");
+
+    // This funny way of writing it is because `-9'223'372'036'854'775'808` isn't a literal.  The
+    // actual literal is `9'223'372'036'854'775'808`, which is too big (by one) to fit into
+    // `int64_t` --- even though `-9'223'372'036'854'775'808` can!  So instead, we negate the
+    // largest literal, and then subtract one to get the lowest literal.
+    constexpr int64_t min = -9'223'372'036'854'775'807 - 1;
+    EXPECT_STREQ(IToA<min>::value.c_str(), "-9223372036854775808");
```
hoffbrinkle, line 73: Do we want to start using the `StrEq` matcher for things like this?

```suggestion
    EXPECT_THAT(IToA<min>::value, StrEq("-9223372036854775808"));
```

au/code/au/utility/string_constant.hh
```diff
@@ -87,6 +90,22 @@ struct IToA;
...
+template <typename S>
+constexpr std::make_unsigned_t<S> abs_as_unsigned(S x) {
+    static_assert(stdx::conjunction<std::is_integral<S>, std::is_signed<S>>::value,
+                  "Only designed for signed integral types");
+    using U = std::make_unsigned_t<S>;
+    constexpr auto SMAX = static_cast<U>(std::numeric_limits<S>::max());
+    constexpr auto UMAX = std::numeric_limits<U>::max();
+
+    auto result = static_cast<U>(x);
+    return (result > SMAX) ? (UMAX - result + 1u) : result;
```
hoffbrinkle, line 106: I don't need the parens for readability (years of C I've learned ternary is pretty far down the precedence list, only above comma), but understand if you want to leave them in.

hoffbrinkle (approved):

au/code/au/utility/string_constant.hh
chiphogg: Yep: for ternary specifically, I never really feel uncertainty about precedence.  However, I do like the visual grouping that the parentheses bring.

PR #398, opened 2025-05-01
Avoid undesired integer promotion

### LINKED ISSUE #395: Compiler Error: Comparison of integer expressions of different signedness
Hi,
I wanted to upgrade to the latest 0.4.1 version.
I installed the prebuilt `au_all_units.hh` and get following compile error:

```console
../source/Common/Units/include/Units/au.h: In function ‘constexpr uintmax_t au::detail::find_prime_factor(uintmax_t)’:

[2025-04-30T11:22:07.560Z] ../source/Common/Units/include/Units/au.h:2883:19: error: comparison of integer expressions of different signedness: ‘int’ and ‘uintmax_t’ {aka ‘long unsigned int’} [-Werror=sign-compare]

[2025-04-30T11:22:07.560Z]  2883 |         if (p * p > n) {
[2025-04-30T11:22:07.560Z]       |             ~~~~~~^~~
```

This is a snippet of `constexpr std::uintmax_t find_prime_factor(std::uintmax_t n)`

It fails with different GCC versions like 9.3 and 13.2

My sub-issue is to enable warnings as errors in your CI: -Wall -Wextra -Werror

chiphogg:
Apparently, this has triggered a `-Wsign-compare` violation by an
obscure and indirect pathway on an end user's build system.  We have not
been able to reproduce this (see #395).  However, on examining the code,
it's clear that `p * p` should be promoted to `int`, which we do compare
with the unsigned `n`.

The simplest fix is to just store each prime in a `uintmax_t`, which is
what we would want it to be in to perform the computations anyway.  The
main benefit of `uint16_t` here is to cram a lot of numbers into the
storage space, and this change preserves that property.

Fixes #395.

### DIFF
--- a/au/code/au/utility/factoring.hh
+++ b/au/code/au/utility/factoring.hh
@@ -96,7 +96,7 @@ constexpr std::uintmax_t find_prime_factor(std::uintmax_t n) {
     //
     // Note that range-for isn't supported until C++17, so we need to use an index.
     for (auto i = 0u; i < FirstPrimes::values.size(); ++i) {
-        const auto &p = FirstPrimes::values[i];
+        const std::uintmax_t p = FirstPrimes::values[i];
 
         if (n % p == 0u) {
             return p;


geoffviola (approved):
The change looks good. There still a question of coverage.

chiphogg:
> The change looks good. There still a question of coverage.

Yep: based on the latest comments on #395, I'm thinking that the coverage may just be too hard to get for now.

I was going to wait to put this out for review "formally" until @StefanoD got back and could test this build, to see if there are any other errors that would be uncovered.  But the change is simple enough that I feel comfortable just landing it now.

PR #400, opened 2025-05-02
Add type-based aliases for power helpers

### LINKED ISSUE #397: Support "type versions" of `squared`, `sqrt`, etc.
Here are some options for spelling this now:

```cpp
using NanoSecondsSquared = decltype(squared(Nano<Seconds>{}));
using NanoSecondsSquared = UnitPowerT<Nano<Seconds>, 2>;
using NanoSecondsSquared = Squared<Nano<Seconds>>;
```

The first two are available today.  The third one is so natural and compact, though, that we wouldn't even be motivated to define the alias: we would just write `QuantityU64<Squared<Nano<Seconds>>>`.

We should do this for all of the "power aliases": `sqrt`, `cbrt`, `squared`, `cubed`, and `inverse`.  And we should try to make it work in an agnostic way, so that we support everything that these aliases support.

CC: @hoffbrinkle for vis.

chiphogg:
Consider the following three options for creating a unit of squared
nanoseconds:

```cpp
using NanoSecondsSquared = decltype(squared(Nano<Seconds>{}));
using NanoSecondsSquared = UnitPowerT<Nano<Seconds>, 2>;
using NanoSecondsSquared = Squared<Nano<Seconds>>;
```

The third one is the most appealing, but it doesn't currently exist.
This PR adds it.

Includes docs.

Fixes #397.

### DIFF
--- a/au/code/au/power_aliases.hh
+++ b/au/code/au/power_aliases.hh
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <cstdint>
+#include <utility>
 
 namespace au {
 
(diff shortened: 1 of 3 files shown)


au/code/au/power_aliases.hh
```diff
@@ -45,29 +46,39 @@ template <typename T>
...
+template <typename T>
+using Cubed = decltype(cubed(std::declval<T>()));
 
 // Make "sqrt" an alias for "root<2>" when the latter exists (for anything).
 template <typename T>
 constexpr auto sqrt(T x) -> decltype(root<2>(x)) {
     return root<2>(x);
 }
+template <typename T>
+using Sqrt = decltype(sqrt(std::declval<T>()));
```
hoffbrinkle, line 74: Broader question: Is this the name we want to use for these helpers.  I know when I see `cbrt` I have to look up what that means.  Our API is pretty fluent, so `SquareRoot` and `CubeRoot` wouldn't really be out of place.

chiphogg: Hmm... I think so?

The argument for the status quo is that `cbrt` should be consistent with `std::cbrt`, so we make `Cbrt` consistent with `cbrt`.  This makes it easier for people to "guess" the rest if they know any of them.

I do agree that for type names, `CubeRoot<Nano<Liters>>` would be nicer than `Cbrt<Nano<Liters>>` in a lot of ways, but for now at least I'm finding the consistency argument more persuasive.

hoffbrinkle: Fair enough.  The lower case matching up is obviously plays well with `std`.  And I don't think we should be polluting the namespace with a bunch of aliases.  Projects can always provide their own alias if desired.

hoffbrinkle (approved):

Issue #370, opened 2024-12-23
Support negative constants


chiphogg:
All I really want is negative constants.  I've previously [argued at length](https://github.com/mpusz/mp-units/issues/510) that we _probably_  don't want negative _units_, and _definitely_ don't want negative _magnitudes_.

In terms of the perfect conversion policy for `Constant`, this isn't too difficult.  We can factor out the guts of `Constant` into a `ConvertsTo` mixin, and then add `NegativeConstant`, with simple rules for keeping track of the sign when multiplying `Constant` and `NegativeConstant`.

But here's the problem.  Multiplying by a `Constant` simply adds a factor of that constant's _unit_ to the number or quantity in question.  If that constant is negative, but the unit itself is not, then we lose the sign information.  We also create a discrepancy between the constant itself, and its associated unit, which feels bug-prone.

OK, so maybe we should open the door for negative units?  Not that we'd _encourage_ them, but supporting them would keep negative constants very simple.

In our unit implementation, we reason in terms of two properties: the _dimension_ and _magnitude_.  Does this mean we need to add a _sign_ property, and keep track of it?  That seems like it could work.  On the other hand, "sign" and "magnitude" feel like two parts of a pretty coherent whole.

So maybe we should allow negative "magnitudes" after all?  Would that be so bad?  I think my objection was focused around the name: a "magnitude" is a "positive real number".  But a "nonzero real number" also makes sense.  It's still closed under products, and _most_ powers.  Some powers are _not_ closed (rather famously, $(-1)^{1/2}$), but we can simply return a hard compiler error if this gets attempted.  And I guess this answers [my objection](https://github.com/mpusz/mp-units/issues/510#issuecomment-1781087492) to the anticipated logarithm use case: we can simply produce a hard compiler error for the logarithm of a negative number, and the remaining use cases are unaffected.

The benefit of starting with `Magnitude`, of course, is that if we do it correctly and carefully, the downstream functionality should _just work_.  (Or at least, _can be made_ to work in a straightforward way.  We'll need to add the ability to reason about the sign of a magnitude, just as we can reason about whether it's an integer or rational.  And we'll need to double check everywhere that uses these properties.  But this should be doable via incremental steps.)

Finally, I'd like to get enough of a proof of concept done to unblock #369 before evaluating whether to land this.  I'm not completely sure all this is worth it until I can see for sure what it unlocks, and what it costs.

PR #402, opened 2025-05-03
Support negative magnitudes

chiphogg:
We add a new base, `au::Negative`.  This always goes at the beginning of
the pack.  We define rational powers such that `Negative` disappears
when the numerator (expressed in lowest terms) is even; otherwise, it
remains if the denominator is odd, and we produce a hard compiler error
if the denominator is even.  For products, the usual machinery works in
most cases, but if both inputs have `Negative`, then we remove them
both, and delegate to the product of what remains.

This brings new magnitude traits: `Abs` for the absolute value, and
`IsPositive` to tell whether a magnitude is positive: formerly, these
would have been the identity, and trivially true, respectively.

For obtaining magnitude representations, we add a new error state
outcome, `ERR_NEGATIVE_NUMBER_IN_UNSIGNED_TYPE`.  If this doesn't
happen, we simply delegate to computing the magnitude of the absolute
value, and negate the result.  This does mean we won't produce an answer
for the minimum value of signed integral types, but we can complete that
later.

Finally, we support magnitude labels by prepending a `-` sign.  We
always prepend this sign, even for unlabeled magnitudes, because it's
important to convey sign information.

**Note that this change enables users to create broken units and
quantities.**  If they use negative magnitudes _with units_, it will not
produce correct results.  But given that (a) nobody has ever been able
to form these magnitudes before, (b) the functionality is undocumented,
and (c) we will follow up very soon with changes that restore this
functionality, this does not seem like a serious concern.

Helps #370.

### DIFF
--- a/au/code/au/magnitude.hh
+++ b/au/code/au/magnitude.hh
@@ -763,7 +763,7 @@ template <typename BP, typename... Ts>
 struct PrependIfExpNegative<BP, Magnitude<Ts...>>
     : std::conditional<(ExpT<BP>::num < 0), Magnitude<BP, Ts...>, Magnitude<Ts...>> {};
 
-// If M is (N/D), DenominatorPartT<M> is D; we want 1/D.
+// Remove all positive powers from M.
 template <typename M>
 using NegativePowers = MagQuotientT<M, NumeratorPartT<M>>;
 }  // namespace detail
--- a/au/code/au/magnitude.hh
+++ b/au/code/au/magnitude.hh
@@ -61,6 +61,38 @@ using MagQuotientT = PackQuotientT<Magnitude, T, U>;
 template <typename T>
 using MagInverseT = PackInverseT<Magnitude, T>;
 
+// Enable negative magnitudes with a type representing (-1) that appears/disappears under powers.
+struct Negative {};
+template <typename... BPs, std::intmax_t ExpNum, std::intmax_t ExpDen>
+struct PackPower<Magnitude, Magnitude<Negative, BPs...>, std::ratio<ExpNum, ExpDen>>
+    : std::conditional<
+          (std::ratio<ExpNum, ExpDen>::num % 2 == 0),
+
+          // Even powers of (-1) are 1 for any root.
+          PackPowerT<Magnitude, Magnitude<BPs...>, ExpNum, ExpDen>,
+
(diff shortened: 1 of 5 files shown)


au/code/au/magnitude.hh
```diff
@@ -661,7 +765,7 @@ struct PrependIfExpNegative<BP, Magnitude<Ts...>>
 
 // If M is (N/D), DenominatorPartT<M> is D; we want 1/D.
 template <typename M>
-using NegativePowers = MagInverseT<DenominatorPartT<M>>;
+using NegativePowers = MagQuotientT<M, NumeratorPartT<M>>;
```
hoffbrinkle, line 768: Should we update the comment?

au/code/au/magnitude_test.cc
```diff
@@ -113,6 +147,11 @@ TEST(IntegerPart, PicksOutIntegersFromNumerator) {
     EXPECT_EQ(integer_part(PI * sqrt(mag<32>()) / mag<15>()), mag<4>());
 }
 
+TEST(IntegerPart, PreservesSign) {
+    EXPECT_EQ(integer_part(-mag<1>()), -mag<1>());
+    EXPECT_EQ(integer_part(-mag<8765>()), -mag<8765>());
```
hoffbrinkle, line 152: I personally don't have a problem mixing the `EXPECT_THAT` matcher syntax with the legacy `EXPECT_EQ` assertion syntax (in that I'll write new tests in the new syntax leaving existing alone, but I do understand the desire to maintain consistency.

hoffbrinkle (approved):

au/code/au/magnitude.hh
chiphogg: Done.

au/code/au/magnitude_test.cc
chiphogg: This is a bit of a time capsule.  I wrote this back in December, which is probably before you persuaded me to use matchers consistently!  I'll update these new test cases.  At some point, we should migrate the whole repo, too.

hoffbrinkle: I'm also fine not touching time capsule tests.  Sometimes it's nice to be looking at history and say "nice, this code has been here doing it's job correctly for X number of years." 😁

Issue #349, opened 2024-12-09
Improve efficiency of explicit-rep overflow checkers


chiphogg:
Our initial implementation for #110 will have separate checks for each of 3 stages: static cast to common type, unit conversion, and static cast to destination type.

Really, there should be a simple boundary between values that won't overflow, and those that will.  Therefore, a more efficient version would be to simply compute this boundary, and compare the input value to it.

PR #403, opened 2025-05-04
Handle negative magnitudes in conversions

chiphogg:
Our code was all written under the assumption that a magnitude was
always positive.  Now we need to go back and see what changes when
magnitudes can be negative.

For applying magnitudes to values, the _category_ tends not to change.
In fact, for everything except the "apply rational magnitudes to
integers" case, the code is identical, and all we do is add more test
cases.  For that rational/integer case, though, we need to update our
checks for whether the numerator can fit: it's specifically the
_absolute value_ of the numerator that matters.  Additionally, we "short
circuit" the max non-overflowing value computation whenever the
magnitude is negative and the _destination_ is unsigned.  (Someday, once
we do #349, _all_ "apply magnitude" cases will have this form, but for
now the rational/integer case is special.)

On the conversion policy code, we now use the absolute value to check
for overflow.  We also add a check for setting an unsigned destination
with a negative scale factor, which can sometimes be OK, but is
certainly not OK to do _implicitly_.

Regarding common units, we used to consider a unit redundant if it was
an integer multiple of another.  Now that we can have negative integer
multiples, this is not what we want.  Consider "neginches" and "feet":
the core algorithm of "take the lowest power of every base" means that
for the base (-1), we should take the power 0 (positive input) rather
than the power 1 (negative input).  So the common unit should be
(something quantity-equivalent to) _inches_, and we shouldn't
short-circuit and return neginches just because every unit is a
(sometimes negative) integer multiple of it.

Finally, this comes with a new capability for `Quantity`: if the scale
factor is exactly `-1`, we support implicit conversions without checking
for overflow.  We add a test to cover this.

Helps #370.


au/code/au/apply_magnitude_test.cc
```diff
@@ -78,6 +97,17 @@ TEST(ApplyMagnitude, DividesForIntegerDivide) {
     }
 }
 
+TEST(ApplyMagnitude, DividesForNegativeIntegerDivide) {
+    constexpr auto minus_one_thirteenth = -ONE / mag<13>();
+    ASSERT_EQ(categorize_magnitude(minus_one_thirteenth), ApplyAs::INTEGER_DIVIDE);
+
+    // This test would fail if our implementation multiplied by the float representation of (-1/13),
+    // instead of dividing by -13, under the hood.
```
hoffbrinkle, line 105: Would it make sense to add an explicit test case to validate this comment?

au/code/au/apply_magnitude_test.cc
```diff
@@ -40,23 +40,34 @@ TEST(CategorizeMagnitude, FindsIntegerMultiplyInstances) {
     EXPECT_EQ(categorize_magnitude(mag<2>()), ApplyAs::INTEGER_MULTIPLY);
     EXPECT_EQ(categorize_magnitude(mag<35>()), ApplyAs::INTEGER_MULTIPLY);
 
+    EXPECT_EQ(categorize_magnitude(-mag<35>()), ApplyAs::INTEGER_MULTIPLY);
```
hoffbrinkle, line 43: Usual comment about `EXPEC_THAT`. 😁

au/code/au/apply_rational_magnitude_to_integral.hh
```diff
@@ -110,6 +110,12 @@ constexpr bool is_known_to_be_less_than_one(Magnitude<BPs...>) {
 template <typename T, typename MagT, bool IsMagLessThanOne>
 struct MaxNonOverflowingValueImplWhenNumFits;
 
+// Implementation helper for "the value is zero" (which recurs a bunch of times).
+template <typename T>
+struct ValueIsZero {
+    static constexpr T value() { return T{0}; }
+};
```
hoffbrinkle, line 117: This one tripped me up a bit since "is" to me usually means an API that answers the question "is this thing that?"  Is this just `ValueZero`?

au/code/au/apply_rational_magnitude_to_integral.hh
```diff
@@ -142,20 +148,27 @@ struct MaxNonOverflowingValueImplWhenNumFits<T, MagT, false> {
...
+// circuit to set the max to zero.
+template <typename T, typename MagT, MagRepresentationOutcome NumOutcome>
+struct MaxNonOverflowingValueImpl<T, MagT, NumOutcome, true> : ValueIsZero<T> {};
+
 // If the numerator fits in the promoted type of `T`, delegate further based on whether the
 // denominator is bigger.
 template <typename T, typename MagT>
-struct MaxNonOverflowingValueImpl<T, MagT, MagRepresentationOutcome::OK>
-    : MaxNonOverflowingValueImplWhenNumFits<T, MagT, is_known_to_be_less_than_one(MagT{})> {};
+struct MaxNonOverflowingValueImpl<T, MagT, MagRepresentationOutcome::OK, false>
```
hoffbrinkle, line 165: I tend to avoid `bool` as a parameter type as it leads to this; at this call site what does `false` mean?  I usually get around this by making the parameter an `enum` even if that `enum` is just giving names to `true` and `false` that help the code be self-documenting at the call site.

au/code/au/apply_rational_magnitude_to_integral_test.cc
```diff
@@ -50,9 +50,16 @@ TEST(PromotedType, PromotesUint8TIntoLargerType) {
     EXPECT_GT(sizeof(PromotedU8), sizeof(uint8_t));
 }
 
-TEST(IsKnownToBeLessThanOne, ProducesExpectedResultsForMagnitudesThatCanFitInUintmax) {
-    EXPECT_TRUE(is_known_to_be_less_than_one(mag<1>() / mag<2>()));
-    EXPECT_TRUE(is_known_to_be_less_than_one(mag<999'999>() / mag<1'000'000>()));
+TEST(IsAbsKnownToBeLessThanOne, ProducesExpectedResultsForMagnitudesThatCanFitInUintmax) {
+    EXPECT_TRUE(is_abs_known_to_be_less_than_one(mag<1>() / mag<2>()));
```
hoffbrinkle, line 54: Obligatory `EXPECT_THAT` comment...

au/code/au/conversion_policy.hh
```diff
@@ -27,11 +27,11 @@ namespace au {
 template <typename Rep, typename... BPs>
 constexpr bool can_scale_without_overflow(Magnitude<BPs...> m, Rep value) {
     // Scales that shrink don't cause overflow.
-    if (get_value<double>(m) <= 1.0) {
+    if (get_value<double>(abs(m)) <= 1.0) {
         (void)value;
         return true;
     } else {
```
hoffbrinkle, line 33: Is the `else` redundant here?

```suggestion
        return true;
    }
```

Personally, the sooner I can drop the conditional context from what I've got to keep active in my "internal/mental model" stack when reading code, the easier it is parse.

au/code/au/conversion_policy_test.cc
```diff
@@ -47,6 +48,14 @@ TEST(ImplicitRepPermitted, TrueForIdentityMagnitude) {
     EXPECT_TRUE((ImplicitRepPermitted<uint8_t, Magnitude<>>::value));
 }
 
+TEST(ImplicitRepPermitted, TrueForNegativeOneUnlessUnsigned) {
+    EXPECT_TRUE((ImplicitRepPermitted<int64_t, Magnitude<Negative>>::value));
```
hoffbrinkle, line 52: Obligatory `EXPECT_THAT`...

au/code/au/quantity_point_test.cc
```diff
@@ -372,6 +372,11 @@ TEST(QuantityPoint, CanCompareUnitsWithDifferentOrigins) {
     EXPECT_THAT(celsius_pt(0), ConsistentlyLessThan(kelvins_pt(274)));
 }
 
+TEST(QuantityPoint, AddingPosUnitQuantityToNegUnitPointGivesPosUnitPoint) {
+    constexpr auto neg_celsius_pt = celsius_pt * (-mag<1>());
+    EXPECT_THAT(neg_celsius_pt(40) + celsius_qty(15), PointEquivalent(celsius_pt(-25)));
```
hoffbrinkle, line 377: "GivesPosUnitPoint"?  It looks like the unit point we're expecting is negative?

au/code/au/conversion_policy.hh
chiphogg: This whole setup feels a little goofy.  I decided to rewrite it as a simple `||` conditional.

au/code/au/quantity_point_test.cc
chiphogg: The _unit itself_ for the point we're expecting is positive, though: we get `celsius_pt` rather than `neg_celsius_pt`.

au/code/au/apply_magnitude_test.cc
chiphogg: I don't want to add a lot of noise in this PR, so here's my counter-offer:

- I filed #404 to track migrating the whole repo.
- I'll provide a fast-follow PR, after this one lands, that fully migrates every file where you left a comment like this one in this PR.

Good deal?  :slightly_smiling_face:

Issue #404, opened 2025-05-04
Use EXPECT_THAT/ASSERT_THAT consistently


chiphogg:
This is consistent with modern best practices for googletest.

hoffbrinkle:
Missed some instances in the release code, and may as well update our documentation while I'm at it.

hoffbrinkle:
Finally complete:

```bash
grep -HnR --perl-regexp '(EXPECT|ASSERT)_(?!THAT)' .
```

PR #417, opened 2025-05-07
Convert existing EQ assertion checks to matchers

hoffbrinkle:
This converts a bulk of the existing `ASSERT_EQ` and `EXPECT_EQ` assertion
checks to the matcher syntax.

Partial implementation for #404.

### DIFF
--- a/au/code/au/zero_test.cc
+++ b/au/code/au/zero_test.cc
@@ -21,6 +21,7 @@ using namespace std::chrono_literals;
 
 namespace au {
 
+using ::testing::Eq;
 using ::testing::IsFalse;
 using ::testing::IsTrue;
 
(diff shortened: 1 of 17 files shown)


au/code/au/zero_test.cc
```diff
@@ -54,20 +55,20 @@ TEST(WrappedInt, BasicInterfaceWorksAsExpected) {
...
 }
 
 TEST(Zero, ComparableToArbitraryQuantities) {
-    EXPECT_EQ(ZERO, WrappedInt{0});
+    EXPECT_THAT(ZERO, Eq(WrappedInt{0}));
     EXPECT_LT(ZERO, WrappedInt{1});
     EXPECT_GT(ZERO, WrappedInt{-1});
 
-    EXPECT_EQ(ZERO, WrappedInt{0});
+    EXPECT_THAT(WrappedInt{0}, Eq(ZERO));
```
chiphogg, line 71: Not sure why these are repeated, but I like that you're making them distinct by swapping the order of the args :sweat_smile:

au/code/au/zero_test.cc
```diff
@@ -84,18 +85,18 @@ TEST(Zero, ComparesEqualToZero) {
...
 }
 
 TEST(Zero, ImplicitlyConvertsToChronoDuration) {
     constexpr std::chrono::nanoseconds zero_ns = ZERO;
-    EXPECT_EQ(zero_ns, 0ns);
+    EXPECT_THAT(zero_ns, Eq(0ns));
 
     constexpr std::chrono::duration<float, std::milli> zero_ms_f = ZERO;
-    EXPECT_EQ(zero_ms_f, (std::chrono::duration<float, std::milli>{0.f}));
+    EXPECT_THAT(zero_ms_f, Eq((std::chrono::duration<float, std::milli>{0.f})));
```
chiphogg, line 99:
```suggestion
    EXPECT_THAT(zero_ms_f, Eq(std::chrono::duration<float, std::milli>{0.f}));
```

chiphogg (approved):

au/code/au/zero_test.cc
hoffbrinkle: Rats, I missed one! 😁

au/code/au/zero_test.cc
hoffbrinkle: I dug into it a bit more.  Came in with c86431a8aaf770bcf13157a1cc6f5436625b42cf.  I'm thinking that it was because the test immediately below it had a "true" block and a "false" block.  I'm thinking maybe we just delete the redundant tests?

chiphogg: Honestly, yeah, that's probably the best approach.

Issue #369, opened 2024-12-23
Maximize common unit for `QuantityPoint`


chiphogg:
Consider this test, from `:quantity_point_test`:

```cpp
TEST(QuantityPoint, CommonPointUnitLabel) {
    EXPECT_THAT(stream_to_string(celsius_pt(0) - kelvins_pt(0)),
                AnyOf(StrEq("5463 EQUIV{[(1 / 20) K], [(1 / 20) degC]}"),
                      StrEq("5463 EQUIV{[(1 / 20) degC], [(1 / 20) K]}")));
}
```

This passes... but only because we're using non-standard and optimized definitions for `Celsius` and `Kelvins` in that file.  The [_actual_ code](https://godbolt.org/z/baTTnG83M) on Au 0.4.1 gives `"27315 EQUIV{[(1 / 100) degC], [(1 / 100) K]}"`.  Yes, it's a little easier to recognize the correct value, but it's also a factor of 5 bigger, which means unnecessary additional overflow risk.

We _could_ try manually specifying every origin in the most efficient units.  However, this is a losing game, because we would need to consider all pairwise combinations of multiple units --- what's going to happen when we bring Fahrenheit into play?

My first attempt to solve this was to store the origin difference in a unit where its value was 1.  I was worried this wouldn't be possible at compile time, but since the origins are `constexpr` quantities, it actually is!  The problem with this is that such a big unit hits our overflow safety surface pretty easily.  But if it's possible to construct this unit in the first place (and it is), then it should be possible to use a `Constant`, and take advantage of its perfect conversion policy!

This is what I think we should do.  I've done enough exploring to convince myself that this is possible.  We're blocked on negative constants, though, and that has a **lot** of implications we'll need to work through.

PR #420, opened 2025-05-07
Add `UnitOfLowestOrigin` helper

chiphogg:
`UnitOfLowestOrigin<Us...>` picks out some unit from among `Us...` whose
origin is equal to the lowest origin.  It will always pick out the same
unit every time, regardless of the ordering of the inputs.  (To get this
behavior, we needed to add a couple more utilities to `:packs`.)

Helps https://github.com/aurora-opensource/au/issues/369.

### DIFF
--- a/au/code/au/unit_of_measure.hh
+++ b/au/code/au/unit_of_measure.hh
@@ -818,6 +818,19 @@ struct CommonOrigin<Head, Tail...> :
                                OriginOf<Head>,
                                CommonOrigin<Tail...>>>> {};
 
+// `UnitOfLowestOrigin<Us...>` is any unit among `Us` whose origin equals `CommonOrigin<Us...>`.
+template <typename... Us>
+struct UnitOfLowestOriginImpl;
+template <typename... Us>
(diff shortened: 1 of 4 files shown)


au/code/au/unit_of_measure.hh
```diff
@@ -818,6 +818,19 @@ struct CommonOrigin<Head, Tail...> :
                                OriginOf<Head>,
                                CommonOrigin<Tail...>>>> {};
 
+// `UnitOfLowestOrigin<Us...>` is any unit among `Us` whose origin equals `CommonOrigin<Us...>`.
+template <typename... Us>
+struct UnitOfLowestOriginImpl;
+template <typename... Us>
+using UnitOfLowestOrigin = typename SortAs<UnitProduct, UnitOfLowestOriginImpl<Us...>>::type;
```
hoffbrinkle, line 825: I see it's a pattern already in the file, so more of a discussion point:  I generally name the type to be what it is (`U`), as opposed to encoding the fact that it's more than one in the type name (`Us`).  Would it always be obvious from context that `U` was 0, 1, or more items (i.e. a pack)?

hoffbrinkle (approved):

Issue #387, opened 2025-02-26
Separate risk checks for overflow and truncation


chiphogg:
Given a conversion from one (unit, rep) to another, we should reason separately about overflow and truncation risks.  This will pave the way for new APIs in #122.

chiphogg:
Note: we should add this kind of argument not just everywhere we currently let users say "coerce", but also for the `as_raw_number` utility.

chiphogg:
Blocked on #349, which is under active development.

PR #421, opened 2025-05-08
Simplify common point unit implementation

chiphogg:
Now, the common point unit is just the _common unit_, except that
instead of just looking at all of the input units, we also consider
their mutual origin displacements (treated as new ad hoc units).  And,
we set an origin that is equal to the lowest origin among those inputs.
That's it!  No more ugly hacks... goodbye `TypeHoldingCommonOrigin`!

This simplifies our logic for conversion checking in `QuantityPoint`,
and maximally expands the set of allowed operations.  This is made
possible because the origin displacement is now always a shapeshifter
type, which has a perfect conversion policy, known at compile time, to
every destination type.  I believe this is a first among units
libraries.

Some people might have been using the `OriginDisplacement` trait, so we
retain it, but it has to move from `unit_of_measure.hh` to
`quantity_point.hh`, because now it depends on `:constant`.  We document
this.

We did need to refine the ordering by adding another tiebreaker based on
the origin displacement unit, too.

Fixes https://github.com/aurora-opensource/au/issues/369.

### DIFF
--- a/au/code/au/quantity_point.hh
+++ b/au/code/au/quantity_point.hh
@@ -77,7 +77,7 @@ template <typename UnitT, typename RepT>
 class QuantityPoint {
     // Q: When should we enable IMPLICIT construction from another QuantityPoint type?
     // A: EXACTLY WHEN our own Diff type can be IMPLICITLY constructed from BOTH the target's Diff
-    //    type, AND the offset between our Units' zero points.
+    //    type AND the offset between our Units' zero points.
     //
     // In other words, there are two ways to fail implicit convertibility.
     //
--- a/au/code/au/quantity_point.hh
+++ b/au/code/au/quantity_point.hh
@@ -159,12 +159,11 @@ class QuantityPoint {
     template <typename NewUnit,
               typename = std::enable_if_t<IsUnit<AssociatedUnitForPointsT<NewUnit>>::value>>
     constexpr Rep in(NewUnit) const {
-        using TargetUnit = AssociatedUnitForPointsT<NewUnit>;
-        // `rep_cast` is needed because if these are integral types, their difference might become a
-        // different type due to integer promotion.
-        return rep_cast<Rep>(x_.as(TargetUnit{}) +
-                             Quantity<TargetUnit, Rep>{origin_displacement(TargetUnit{}, Unit{})})
-            .in(TargetUnit{});
+        constexpr auto target = AssociatedUnitForPointsT<NewUnit>{};
+
+        // The explicit `<Rep>` is needed because if these are integral types, their difference
+        // might become a different type due to integer promotion.
+        return (x_.as(target) + origin_displacement(target, Unit{})).template in<Rep>(target);
     }
 
     // "Forcing" conversions, which explicitly ignore safety checks for overflow and truncation.
--- a/au/code/au/quantity_point.hh
+++ b/au/code/au/quantity_point.hh
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include "au/constant.hh"
 #include "au/fwd.hh"
 #include "au/quantity.hh"
 #include "au/stdx/type_traits.hh"
@@ -46,19 +47,37 @@ template <typename P1, typename P2>
 struct AreQuantityPointTypesEquivalent;
 
 namespace detail {
-template <typename TargetRep, typename U1, typename U2>
-struct OriginDisplacementFitsIn;
+template <typename U, typename R, typename ConstUnit>
+constexpr Quantity<U, R> coerce_as_quantity(Constant<ConstUnit> c) {
+    return c.template coerce_as<R>(U{});
+}
+template <typename U, typename R>
+constexpr Quantity<U, R> coerce_as_quantity(Zero z) {
+    return z;
+}
 
 template <typename FromRep, typename ToRep>
 struct IntermediateRep;
 }  // namespace detail
 
+// Some units have an "origin".  This is not meaningful by itself, but its difference w.r.t. the
+// "origin" of another unit of the same Dimension _is_ meaningful.  This type trait provides access
+// to that difference.
+template <typename U1, typename U2>
+constexpr auto origin_displacement(U1, U2) {
+    return make_constant(detail::ComputeOriginDisplacementUnit<AssociatedUnitForPointsT<U1>,
+                                                               AssociatedUnitForPointsT<U2>>{});
+}
+
+template <typename U1, typename U2>
+using OriginDisplacement = decltype(origin_displacement(U1{}, U2{}));
+
 // QuantityPoint implementation and API elaboration.
 template <typename UnitT, typename RepT>
 class QuantityPoint {
     // Q: When should we enable IMPLICIT construction from another QuantityPoint type?
-    // A: EXACTLY WHEN our own Diff type can be IMPLICITLY constructed from the SUM of the target's
-    //    Diff type, and the offset between our Units' zero points.
+    // A: EXACTLY WHEN our own Diff type can be IMPLICITLY constructed from BOTH the target's Diff
+    //    type, AND the offset between our Units' zero points.
     //
     // In other words, there are two ways to fail implicit convertibility.
     //
@@ -72,10 +91,8 @@ class QuantityPoint {
     //      OK : QuantityPoint<Celsius, int> -> QuantityPoint<Milli<Kelvins>, int>
     template <typename OtherUnit, typename OtherRep>
     static constexpr bool should_enable_implicit_construction_from() {
-        return std::is_convertible<
-            decltype(std::declval<typename QuantityPoint<OtherUnit, OtherRep>::Diff>() +
-                     origin_displacement(UnitT{}, OtherUnit{})),
-            QuantityPoint::Diff>::value;
+        using Com = CommonUnitT<OtherUnit, detail::ComputeOriginDisplacementUnit<Unit, OtherUnit>>;
+        return std::is_convertible<Quantity<Com, OtherRep>, QuantityPoint::Diff>::value;
     }
 
     // This machinery exists to give us a conditionally explicit constructor, using SFINAE to select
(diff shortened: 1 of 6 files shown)


au/code/au/quantity_point.hh
```diff
@@ -46,19 +47,37 @@ template <typename P1, typename P2>
...
+using OriginDisplacement = decltype(origin_displacement(U1{}, U2{}));
+
 // QuantityPoint implementation and API elaboration.
 template <typename UnitT, typename RepT>
 class QuantityPoint {
     // Q: When should we enable IMPLICIT construction from another QuantityPoint type?
-    // A: EXACTLY WHEN our own Diff type can be IMPLICITLY constructed from the SUM of the target's
-    //    Diff type, and the offset between our Units' zero points.
+    // A: EXACTLY WHEN our own Diff type can be IMPLICITLY constructed from BOTH the target's Diff
+    //    type, AND the offset between our Units' zero points.
```
hoffbrinkle, line 80:
```suggestion
    //    type AND the offset between our Units' zero points.
```

au/code/au/quantity_point.hh
```diff
@@ -131,27 +148,23 @@ class QuantityPoint {
...
+    constexpr NewRep in(NewUnit) const {
         using CalcRep = typename detail::IntermediateRep<Rep, NewRep>::type;
-        return (rep_cast<CalcRep>(x_) -
-                rep_cast<CalcRep>(
-                    OriginDisplacement<Unit, AssociatedUnitForPointsT<NewUnit>>::value()))
-            .template in<NewRep>(associated_unit_for_points(u));
+        using Target = AssociatedUnitForPointsT<NewUnit>;
+        return (x_.template as<CalcRep>(Target{}) +
+                detail::coerce_as_quantity<Target, CalcRep>(origin_displacement(Target{}, Unit{})))
+            .template in<NewRep>(Target{});
```
hoffbrinkle, line 156: Much nicer.

au/code/au/quantity_point.hh
```diff
@@ -131,27 +148,23 @@ class QuantityPoint {
...
+        using TargetUnit = AssociatedUnitForPointsT<NewUnit>;
         // `rep_cast` is needed because if these are integral types, their difference might become a
         // different type due to integer promotion.
-        return rep_cast<Rep>(
-                   x_ + rep_cast<Rep>(
-                            OriginDisplacement<AssociatedUnitForPointsT<NewUnit>, Unit>::value()))
-            .in(associated_unit_for_points(u));
+        return rep_cast<Rep>(x_.as(TargetUnit{}) +
+                             Quantity<TargetUnit, Rep>{origin_displacement(TargetUnit{}, Unit{})})
+            .in(TargetUnit{});
```
hoffbrinkle, line 167: Would this work?

```suggestion
        return (x_.as(TargetUnit{}) +
                Quantity<TargetUnit, Rep>{origin_displacement(TargetUnit{}, Unit{})})
            .in<Rep>(TargetUnit{});
```

au/code/au/quantity_point_test.cc
```diff
@@ -437,6 +443,22 @@ TEST(QuantityPoint, PreservesRep) {
...
 
+TEST(OriginDisplacement, IdenticallyZeroForOriginsThatCompareEqual) {
+    ASSERT_THAT(detail::OriginOf<Celsius>::value(),
+                Not(SameTypeAndValue(detail::OriginOf<AlternateCelsius>::value())));
+    EXPECT_THAT(origin_displacement(Celsius{}, AlternateCelsius{}), SameTypeAndValue(ZERO));
+}
+
+TEST(OriginDisplacement, GivesDisplacementFromFirstToSecond) {
+    EXPECT_EQ(origin_displacement(Kelvins{}, Celsius{}), milli(kelvins)(273'150));
+    EXPECT_EQ(origin_displacement(Celsius{}, Kelvins{}), milli(kelvins)(-273'150));
```
hoffbrinkle, line 454:
```suggestion
    EXPECT_THAT(origin_displacement(Kelvins{}, Celsius{}), Eq(milli(kelvins)(273'150)));
    EXPECT_THAT(origin_displacement(Celsius{}, Kelvins{}), Eq(milli(kelvins)(-273'150)));
```

au/code/au/quantity_point_test.cc
```diff
@@ -437,6 +443,22 @@ TEST(QuantityPoint, PreservesRep) {
...
+}
+
+TEST(OriginDisplacement, FunctionalInterfaceHandlesInstancesCorrectly) {
+    EXPECT_EQ(origin_displacement(kelvins_pt, celsius_pt), milli(kelvins)(273'150));
+    EXPECT_EQ(origin_displacement(celsius_pt, kelvins_pt), milli(kelvins)(-273'150));
```
hoffbrinkle, line 459:
```suggestion
    EXPECT_THAT(origin_displacement(kelvins_pt, celsius_pt), Eq(milli(kelvins)(273'150)));
    EXPECT_THAT(origin_displacement(celsius_pt, kelvins_pt), Eq(milli(kelvins)(-273'150)));
```

au/code/au/unit_of_measure_test.cc
```diff
@@ -612,7 +601,8 @@ TEST(MakeCommonPoint, PreservesCategory) {
 
     // The origin of the common point unit is the lowest origin among all input units.
     EXPECT_THAT(celsenheit_pt(0), Eq(fahrenheit_pt(0)));
-    EXPECT_LT(celsenheit_pt(0), celsius_pt(0));
+    EXPECT_LT(celsenheit_pt(0), celsius_pt(0))
+        << "Difference: " << celsius_pt(0) - celsenheit_pt(0);
```
hoffbrinkle, line 605:
```suggestion
    EXPECT_THAT(celsenheit_pt(0), Lt(celsius_pt(0)))
        << "Difference: " << celsius_pt(0) - celsenheit_pt(0);
```

docs/reference/unit.md
```diff
@@ -488,6 +488,26 @@ that is what this trait produces.
...
+!!! note
+    Au 0.4.1 was the last release where the `OriginDisplacement` trait could be found in
+    `"au/unit_of_measure.hh"`, along with all of the other traits documented here.  For all
+    subsequent releases, it can be found in `"au/quantity_point.hh"`.
+
+    The reason we moved it was because `"au/constant.hh"` (which defines `Constant`) depends on
+    `"au/quantity.hh"`, which in turn depends on `"au/unit_of_measure.hh"`.  Therefore, we could
+    never have used `Constant` inside of `"au/unit_of_measure.hh"`.  However,
+    `"au/quantity_point.hh"` _could_ depend on `"au/constant.hh"`.  Origin displacements aren't very
+    useful without `QuantityPoint` anyway, so this new home is acceptable.
```
hoffbrinkle, line 509: Almost seems more than acceptable?  Seems like an inherent trait of a `QuantityPoint`, so maybe makes more sense there?  Should we move this documentation to `quantity_point.md`?

au/code/au/quantity_point.hh
chiphogg: Well, the wise guy answer is "no, because it won't compile".  :grin:  (This is a "dot-template" case.)

But you got me to stop and think about this code that I wrote months ago, and abandoned once I got it working.  First, `origin_displacement` is a shapeshifter type, so let's use it like one!  Now this leaves two possibilities:

```cpp
        return (x_.as(target) + origin_displacement(target, Unit{})).template in<Rep>(target);
        return rep_cast<Rep>(x_.as(target) + origin_displacement(target, Unit{})).in(target);
```

On the one hand, it's generally nice to supply the rep and unit _jointly_ when we want to specify both.  On the other hand, we know this isn't a _unit conversion_ (we're already in `target`), so the `rep_cast` is maybe a little bit clearer in that we're just undoing the mischief of integer promotion.  And, it's one character shorter.

Ultimately, I more or less flipped a coin here.  But in either case, it's a whole lot cleaner than what we had before you made this comment!

au/code/au/quantity_point.hh
```diff
@@ -46,19 +47,37 @@ template <typename P1, typename P2>
...
-template <typename TargetRep, typename U1, typename U2>
-struct OriginDisplacementFitsIn;
+template <typename U, typename R, typename ConstUnit>
+constexpr Quantity<U, R> coerce_as_quantity(Constant<ConstUnit> c) {
+    return c.template coerce_as<R>(U{});
+}
+template <typename U, typename R>
+constexpr Quantity<U, R> coerce_as_quantity(Zero z) {
+    return z;
+}
```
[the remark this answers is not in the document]
chiphogg: I certainly wouldn't have written something as clunky as these functions unless I felt forced to do so.  :slightly_smiling_face:

In this case, the driving factor is that we need to treat `Zero` the same way as `Constant<U>`.  For the "forcing" conversions, we need to overrule the perfect conversion policy of `Constant<U>`.  `detail::coerce_as_quantity` is an OK way to spell that.  This just means we need to provide the same function for `Zero`.

Note, too, that the fact that it's in the `detail` namespace (soon to be `auimpl`) means that it's _not_ part of our API!  I agree with you that we wouldn't want to promote this interface broadly.  This is just a helper for implementing `QuantityPoint`.

-----

Also:

> Basically, it'd be nice to get a `Constant` in a `Rep` in whatever `Unit` the constant is in naturally.

Just for clarity, this is what `c.as<Rep>()` does today; it just so happens that it always stores `Rep{1}` under the hood.

-----

Also also: I hope once we have #387, it could unlock more fluent interfaces for `Constant<U>`.  Maybe we could have a second template parameter that governs the strictness of the conversion policy, and we can provide helper member functions that convert the constant into one with a different policy strictness!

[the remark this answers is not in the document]
hoffbrinkle: > Just for clarity, this is what c.as<Rep>() does today; it just so happens that it always stores Rep{1} under the hood.

I see now that it works.  Maybe it's the other way I'm thinking about.  `c.as(u)`.  In those cases, it's almost never able to infer the `Rep` so I end up having to do `c.as<Rep>(u)`.  Which currently disables our checks, which #387 should help us with.  So that was a roundabout way to get to the same place. 😁

docs/reference/unit.md
hoffbrinkle: To clarify, I'm fine where it is.

Oh, I see.  I think.  Would it be fare to say that `Origin` is a trait for a `QuantityPoint` and `OriginDisplacement` is a trait on the Unit?

hoffbrinkle (approved):

docs/reference/unit.md
chiphogg: I would say that `Origin` is a trait for a unit, and `OriginDisplacement` is a trait on a _pair of_ units.

That's in the current setup of the library.  I definitely don't think it's optimal.  It would be good to enable custom origins that _could_ be associated with the `QuantityPoint`.  Someday I hope we can find the right balance between supporting that kind of use case, and also supporting units with very clear intrinsic non-"zero" origins such as Celsius and Fahrenheit.

PR #424, opened 2025-05-09
Convert EXPECT_STREQ to matcher syntax

hoffbrinkle:
This converts existing uses of `EXPECT_STREQ` to use the `StrEq` and `IsEmpty`
matchers.  Note that we really want to test the `StringContstant`'s ability to
do string things.  If we want to test the behavior of `c_str`, that should be a
specific test.

Partial implementation for #404.

### DIFF
--- a/au/code/au/utility/test/string_constant_test.cc
+++ b/au/code/au/utility/test/string_constant_test.cc
@@ -21,13 +21,14 @@
 namespace au {
 
 using ::testing::Eq;
+using ::testing::IsEmpty;
 using ::testing::StrEq;
 
 namespace detail {
 
 TEST(StringConstant, CanCreateFromStringLiteral) {
     constexpr StringConstant<5> x{"hello"};
-    EXPECT_STREQ(x.c_str(), "hello");
+    EXPECT_THAT(x, StrEq("hello"));
 }
 
 TEST(StringConstant, HasLengthMember) {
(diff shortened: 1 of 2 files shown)


au/code/au/utility/test/string_constant_test.cc
```diff
@@ -92,46 +93,46 @@ TEST(IToA, HasLengthMember) {
 }
 
 TEST(UIToA, CanHandleNumbersBiggerThanIntmaxButWithinUintmax) {
-    EXPECT_STREQ(UIToA<10000000000000000000u>::value.c_str(), "10000000000000000000");
```
chiphogg, line 95: Thanks for adding separators here.  I know it technically makes the "string of source code" not equal to the printed string, but I appreciate making it easier to tell what the actual number is.  (Plus, if the test passes, you _know_ that it's the same number.)

au/code/au/utility/test/string_constant_test.cc
```diff
@@ -21,13 +21,14 @@
...
 
 using ::testing::Eq;
+using ::testing::IsEmpty;
 using ::testing::StrEq;
 
 namespace detail {
 
 TEST(StringConstant, CanCreateFromStringLiteral) {
     constexpr StringConstant<5> x{"hello"};
-    EXPECT_STREQ(x.c_str(), "hello");
```
chiphogg, line 30: I didn't realize we could get rid of these `c_str()` calls.  Neat!  I wonder if we needed them with `EXPECT_STREQ`, and if not, then I wonder why I added them in the first place.

chiphogg (approved):

au/code/au/utility/test/string_constant_test.cc
hoffbrinkle: I'm not sure I tried it.  I do know the matcher certainly provided better ability to match different types, so I just tried it and it worked!

PR #427, opened 2025-05-09
Rename `detail` namespace to `auimpl`

### LINKED ISSUE #412: Rename `detail` namespace to `auimpl`
Motivation: <https://abseil.io/tips/130>.

> Try to give sub-namespaces unique names that are unlikely to collide with future top-levels.

chiphogg:
I think a "bite the bullet megadiff" makes the most sense here, because
the changes should all be the same length, so the PR should be fairly
easy to review.  The reviewer can check the Viewed box for one file at a
time, and thus not be forced to do it all in one pass --- as long as we
do end up landing this relatively soon before we try other changes.

Fixes #412.

### DIFF
--- a/au/code/au/unit_of_measure.hh
+++ b/au/code/au/unit_of_measure.hh
@@ -28,7 +28,7 @@ namespace au {
 // - a member typedef `Dim`, which is a valid Dimension; and,
 // - a member typedef `Mag`, which is a valid Magnitude.
 //
-// These can be accessed by traits `detail::DimT` and `detail::MagT`, respectively.  The detail
+// These can be accessed by traits `auimpl::DimT` and `auimpl::MagT`, respectively.  The auimpl
 // namespace is meant to discourage _end users_ from accessing these concepts directly.  For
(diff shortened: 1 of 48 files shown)


au/code/au/unit_of_measure.hh
```diff
@@ -1150,18 +1150,18 @@ struct UnitAvoidance<CommonUnit<Us...>> : std::integral_constant<int, 6> {};
...
     : LexicographicTotalOrdering<A,
                                  B,
-                                 detail::OrderByUnitAvoidance,
-                                 detail::OrderByDim,
-                                 detail::OrderByMag,
-                                 detail::OrderByScaleFactor,
-                                 detail::OrderByOrigin,
-                                 detail::OrderAsUnitProduct,
-                                 detail::OrderAsOriginDisplacementUnit> {};
+                                 auimpl::OrderByUnitAvoidance,
```
chiphogg, line 1159: I think I'm going to skip doing this for now, because `au::detail::UnitAvoidance` is actually in our public facing documentation.  We'll need to give folks a migration path away from naming the `detail` namespace before we change its name.

PR #430, opened 2025-05-14
Implement `lerp`

### LINKED ISSUE #390: Support std::lerp

chiphogg:
This wraps `std::lerp`.  Includes documentation.

The main implementations, in `math.hh`, convert each input to the common
unit before performing the calculation, and then wrap the result.

The hidden friend implementation, in `Quantity`, is necessary to support
shapeshifter types such as `Zero` and `Constant<U>`.  We would need to
add one in `QuantityPoint` too in order to support shapeshifter
overloads, but `QuantityPoint` doesn't have any shapeshifter types in
the current design.

Fixes #390.

### DIFF
--- a/au/code/au/math.hh
+++ b/au/code/au/math.hh
@@ -312,6 +312,29 @@ constexpr bool isnan(QuantityPoint<U, R> p) {
     return std::isnan(p.in(U{}));
 }
 
+//
+// Linear interpolation between two values of the same dimension, as per `std::lerp`.
+//
+// Note that `std::lerp` is not defined until C++20, so neither is `au::lerp`.
+//
+// Note, too, that the implementation for same-type `Quantity` instances lives inside of the
+// `Quantity` class implementation as a hidden friend, so that we can support shapeshifter types
+// such as `Zero` or `Constant<U>`.
+//
+#if defined(__cpp_lib_interpolate) && __cpp_lib_interpolate >= 201902L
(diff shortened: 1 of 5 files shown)


au/code/au/math.hh
```diff
@@ -312,6 +312,29 @@ constexpr bool isnan(QuantityPoint<U, R> p) {
...
+//
+// Note that `std::lerp` is not defined until C++20, so neither is `au::lerp`.
+//
+// Note, too, that the implementation for same-type `Quantity` instances lives inside of the
+// `Quantity` class implementation as a hidden friend, so that we can support shapeshifter types
+// such as `Zero` or `Constant<U>`.
+//
+#if defined(__cpp_lib_interpolate) && __cpp_lib_interpolate >= 201902L
+template <typename U1, typename R1, typename U2, typename R2, typename T>
+constexpr auto lerp(Quantity<U1, R1> q1, Quantity<U2, R2> q2, T t) {
```
geoffviola, line 326: Is passing in a raw scalar for `T` intended? What about forcing `Unos` so that percentages are supported?

chiphogg: It was intended, because the semantics of the third argument are always dimensionless, and I expect people would use raw numbers almost always.  That said, it's a good suggestion and was something I didn't think of, and I think we could support `percent` without undue difficulty.  Check out the latest commit.

geoffviola (approved):
Looks good. Thanks for supporting percent. 😁

Issue #428, opened 2025-05-12
Comparisons with mixed-rep quantity types can suppress warnings inappropriately


chiphogg:
See <https://godbolt.org/z/MfjEj19Yn>.  Basically, mixed-sign comparisons can produce incorrect results (e.g., `-1 > 1u` is generally `true`).  The problem for Au is that when we compare two `Quantity` types with a different unit **or** rep, we explicitly cast them to their common unit-and-rep, and then perform the comparison for the same types.

What we really should be doing is converting them to their common _units_, and then doing the comparison with respect to the raw numbers.  Implementing this will be tricky and subtle, because we have to consider things like:

- Hidden friends (which are important for supporting shapeshifter types like `Zero`, in a low-effort and symmetric way), which means there will generally be _two_ matching implementations for every mixed-rep comparison.
- Negative units, which means we need to pass the operands in the reversed order.

We'll need to think and explore to figure out the best way to solve this.

The above is all about getting the Au code to produce compiler warnings/errors in the same situations that the raw numeric code does.  But we'll want to go further: we'll also want a solution for _actually performing the comparison correctly_: something equivalent to `std::cmp_less` and friends (from C++20).

PR #432, opened 2025-05-20
Compare underlying values of hetereogeneous types

chiphogg:
Formerly, in `Quantity`, we used a `CompareUnderlyingValues` helper,
and a `UnitSign` enum.  The helper was responsible for taking two
_identical_ `Quantity` types, and extracting their values.  (We built
out this machinery to support negative units.)

Now, we refactor to use a new helper, `SignAwareComparison`.  This
brings two simplifications and one generalization.

- **Simplification:** stop doing the extraction in the helper; let the
  callers extract and pass the values.  (This gives the helper just one
  responsibility: namely, take the unit sign into account when deciding
  which order to pass the parameters.)

- **Simplification:** lose the enum; just use the new `UnitSign` trait.

- **Generalization:** accept _heterogeneous_ parameters.

This last change will allow us to detect when we are passing mixed
signed and unsigned integral arguments.  Of course, this PR is a no-op
refactor, so we're not actually taking advantage of this yet.  That will
come in a follow-on PR.

Helps #428.

### DIFF
--- a/au/code/au/operators.hh
+++ b/au/code/au/operators.hh
@@ -14,14 +14,12 @@
 
 #pragma once
 
-// This file provides drop-in replacements for certain standard library function objects for
-// comparison and arithmetic: `std::less<void>`, `std::plus<void>`, etc.
+// This file provides alternatives to certain standard library function objects for comparison and
+// arithmetic: `std::less<void>`, `std::plus<void>`, etc.
 //
 // These are _not_ intended as _fully general_ replacements.  They are _only_ intended for certain
-// specific use cases in this library, where we can ensure certain preconditions are met before they
-// are called.  For example, these utilities don't handle comparing signed and unsigned integral
-// types, because we only ever use them in places where we've already explicitly cast our quantities
-// to the same Rep.
+// specific use cases in this library.  External user code should not use these utilities: they are
+// subject to change their contract at any time to suit the needs of Au.
 //
 // There are two main reasons we rolled our own versions instead of just using the ones from the
 // standard library (as we had initially done).  First, the `<functional>` header is moderately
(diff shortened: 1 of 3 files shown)


au/code/au/quantity.hh
```diff
@@ -115,13 +115,8 @@ constexpr T as_raw_number(T x) {
...
 namespace detail {
-enum class UnitSign {
-    POSITIVE,
-    NEGATIVE,
-};
-
-template <typename Rep, UnitSign>
-struct CompareUnderlyingValues;
+template <typename SignMag, typename Op>
+struct SignAwareComparison;
```
hoffbrinkle, line 119: Seems like we could use some internal documentation of why we need this?  Something that points out that our comparisons happen in a common type, which looses the signed/unsigned of underlying types, and this effectively allows to implement something equivalent to `std::cmp_less` and friends?

au/code/au/operators.hh
```diff
@@ -37,57 +35,57 @@ namespace detail {
 //
 
 struct Equal {
-    template <typename T>
-    constexpr bool operator()(const T &a, const T &b) const {
+    template <typename T, typename U>
```
hoffbrinkle, line 38: So what necessitated this change?  Are we now comparing two different types?

au/code/au/quantity.hh
chiphogg: Added a doc comment.

au/code/au/operators.hh
chiphogg: For now, we're just _paving the way_ to compare two different types.  We'll need this when we end up passing, say, a signed and unsigned value together to a comparison function.  But in the present PR, this is just a no-op refactor.

hoffbrinkle: Makes sense, just making sure I wasn't missing something.

hoffbrinkle (approved):

PR #433, opened 2025-05-21
Avoid changing signedness of integral rep

chiphogg:
The rep we use with our comparison logic is really tricky and subtle.

At first blush, it would be appealing to simply pass `q.in(U{})` for
both arguments, because `U` is a common unit, so this should just be an
integer multiplication.  This way, we automatically opt in to overflow
checks.

We tried this, though, and it was a disaster: tons of tests failed.  It
turns out that there are many situations where we compare, say, a "large
unit, small rep" quantity to a "small unit, large rep" quantity.  In
these cases, it's natural and good to use something as big as the common
rep when converting the first quantity: we can safely take advantage of
the "extra room" in the bigger type.

The real problem comes when the common rep type is _integral_, **and**
it _changes the sign_ of an input type.  That's how we get things like
`-1 < 1u` being false.

To tackle this problem head-on, we make a new trait with an explicit
name: `CommonTypeButPreserveIntSignedness<R1, R2>` is the common type of
`R1` and `R2`, except that _if_ it is integral, then it _preserves the
signedeness_ of the **first** argument.

By using this in our comparison functions, we are now faithfully
preserving the "category" of the inputs all the way until they are
compared.  This means that `inches(-1) < inches(1u)` _will_ reduce to
`-1 < 1u`, instead of masking the situation with a hidden explicit cast
under the hood.

Helps #428, but does not _quite_ fix it: we'll see why in the next PR,
which does.

### DIFF
--- a/au/code/au/utility/test/type_traits_test.cc
+++ b/au/code/au/utility/test/type_traits_test.cc
@@ -75,5 +75,16 @@ TEST(IncludeInPackIf, MakesPackOfEverythingThatMatches) {
         Pack<uint8_t, uint64_t>>();
 }
 
+TEST(CommonTypeButPreserveIntSignedness, CommonTypeIfItIsNotIntegral) {
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<int, double>, double>();
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<float, int>, float>();
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<double, float>, double>();
+}
+
+TEST(CommonTypeButPreserveIntSignedness, UsesSignOfFirstArgumentIfCommonTypeIsIntegral) {
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<int, uint8_t>, int>();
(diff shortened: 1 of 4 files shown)


au/code/au/utility/test/type_traits_test.cc
```diff
@@ -75,5 +75,16 @@ TEST(IncludeInPackIf, MakesPackOfEverythingThatMatches) {
         Pack<uint8_t, uint64_t>>();
 }
 
+TEST(CommonTypeButPreserveIntSignedness, CommonTypeIfItIsNotIntegral) {
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<int, double>, double>();
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<float, int>, float>();
+    StaticAssertTypeEq<CommonTypeButPreserveIntSignedness<double, float>, double>();
+}
```
hoffbrinkle, line 82: Should we add some unsigned tests to this test case?

hoffbrinkle (approved):

au/code/au/utility/test/type_traits_test.cc
chiphogg: `u` got it

hoffbrinkle: 🤦‍♂️

PR #434, opened 2025-05-21
Fix comparisons for mixed-signedness integral reps

### LINKED ISSUE #428: Comparisons with mixed-rep quantity types can suppress warnings inappropriately

chiphogg:
We upgrade our operator replacements (`Less`, `Equal`, etc.) to detect
the situation where both reps have integral type, and delegate to the
more robust comparators that handle that case directly.

Fixes #428.

### DIFF
--- a/release/common_test_cases.hh
+++ b/release/common_test_cases.hh
@@ -50,5 +50,7 @@ TEST(CommonSingleFile, IncludesMathFunctions) {
     EXPECT_DOUBLE_EQ(sin(radians(get_value<double>(PI / mag<2>()))), 1.0);
 }
 
+TEST(CommonSingleFile, MixedSignQuantityComparisonWorks) { EXPECT_LT(meters(-1), meters(1u)); }
+
 }  // namespace
 }  // namespace au
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -408,6 +408,7 @@ cc_library(
     name = "operators",
     hdrs = ["code/au/operators.hh"],
     includes = ["code"],
+    deps = [":stdx"],
 )
 
 cc_test(
--- a/au/code/au/operators.hh
+++ b/au/code/au/operators.hh
@@ -14,6 +14,9 @@
 
 #pragma once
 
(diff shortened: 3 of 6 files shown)


release/common_test_cases.hh
```diff
@@ -50,5 +50,7 @@ TEST(CommonSingleFile, IncludesMathFunctions) {
     EXPECT_DOUBLE_EQ(sin(radians(get_value<double>(PI / mag<2>()))), 1.0);
 }
 
+TEST(CommonSingleFile, MixedSignQuantityComparisonWorks) { EXPECT_LT(meters(-1), meters(1u)); }
```
hoffbrinkle, line 53: Huh, I totally missed this file when doing the Matcher conversions, didn't realize there was code outside of `au/`.

hoffbrinkle (approved):

release/common_test_cases.hh
chiphogg: Oops... I thought I had a good `EXPECT_THAT` habit going now.  Oh well, we can convert the whole file at once anyway.

PR #437, opened 2025-05-30
Update python, fixing dependabot issues 14, 16, 17

chiphogg:
We're about due for a python package update, and this should also fix
dependabot issues 14, 16, and 17.  These are all related to jinja2, and
all would be fixed by upgrading to at least version 3.1.6, which is
indeed the new version that we have.

I followed the instructions in `REQUIREMENTS_LOCK.md`.

On the version update front, `mike` and `mkdocs` had no update at all.
`mkdocs-material` contained updates 9.5.49, 9.5.50, and 9.6.0 through
9.6.14.  Some of these mentioned mermaid fixes, but mostly it didn't
seem like the changes should affect our rendered webpage.

On the A/B testing front, none of the pages I spot-checked showed any
difference at all.  I made sure to check some mermaid diagrams (using
the monovalue types page), but these looked identical too.

On the mike front: technically, I probably didn't need to test this
since it didn't change, but I did anyway and it seemed to work fine.


timhirsh (approved):
👍

PR #440, opened 2025-07-02
Simplify include paths

matthew-reynolds:
Reorganize the codebase such as the code is located at `//au/*` rather than `//au/code/au/*`.

The primary motivation here is to conform more closely with the canonical header path style that Bazel likes by avoiding [`cc_library.includes`](https://bazel.build/reference/be/c-cpp#cc_library.includes). Previously, we had Bazel targets that looked like
```
cc_library(
    hdrs = ["code/au/au.hh"],
    includes = ["code"],
    ...
)
```

It is unusual and somewhat problematic to do this. The `hdrs` suggests that code should `#include "PACKAGE_PATH/code/au/au.hh"`, but we actually want people to just use the more typical `#include "PACKAGE_PATH/au.hh"`. Using `includes` to permit this essentially sidesteps Bazel's normal include path logic, and typically causes the headers to be added to the system header include path (`-isystem`) rather than the normal include path (`-I`). This can have undesirable side effects such as suppressing warnings in these headers.

One alternative would be to use `strip_include_prefix = "code/au"`. This brings the include path management back into Bazel's domain, but results in a bunch of ugly and unnecessary virtual includes.

We should just simplify our include paths by dropping the extra `code/au` directory entirely.

---

Note that doing this actually revealed a couple additional warnings that were previously suppressed by the `-isystem`. These have been addressed in this PR as well.


au/utility/string_constant.hh
```diff
@@ -103,7 +103,7 @@ constexpr std::make_unsigned_t<S> abs_as_unsigned(S x) {
     constexpr auto UMAX = std::numeric_limits<U>::max();
 
     auto result = static_cast<U>(x);
-    return (result > SMAX) ? (UMAX - result + 1u) : result;
+    return (result > SMAX) ? static_cast<U>(UMAX - result + 1u) : result;
```
matthew-reynolds: Without this change, this function fails with `-Wconversion`.

GCC10:
```
In file included from au/utility/test/string_constant_test.cc:15:
./au/utility/string_constant.hh: In instantiation of 'constexpr std::make_unsigned_t<S> au::detail::abs_as_unsigned(S) [with S = signed char; std::make_unsigned_t<S> = unsigned char]':
au/utility/test/string_constant_test.cc:50:5:   required from here
./au/utility/string_constant.hh:106:28: error: conversion from 'unsigned int' to 'std::make_unsigned_t<signed char>' {aka 'unsigned char'} may change value [-Werror=conversion]
  106 |     return (result > SMAX) ? (UMAX - result + 1u) : result;
      |            ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Clang11: No error

Clang14: No error

au/quantity.hh
```diff
@@ -177,11 +177,18 @@ class Quantity {
...
+                                          std::common_type_t<Rep, detail::RealPart<NewRep>>,
+                                          std::common_type_t<Rep, NewRep>>;
+        using Intermediate = std::conditional_t<REAL_TO_COMPLEX, detail::RealPart<NewRep>, NewRep>;
         using Factor = UnitRatioT<AssociatedUnitT<Unit>, AssociatedUnitT<NewUnit>>;
 
         return make_quantity<AssociatedUnitT<NewUnit>>(
-            static_cast<NewRep>(detail::apply_magnitude(static_cast<Common>(value_), Factor{})));
+            static_cast<NewRep>(static_cast<Intermediate>(
+                detail::apply_magnitude(static_cast<Common>(value_), Factor{}))));
     }
```
matthew-reynolds: Without this change, this function fails with `-Wfloat-conversion`. This is because we're calling `std::complex<int>(int, int)` with as double. The `static_cast<Common>(value_)` aka `static_cast<std::complex<int>>(value_)` doesn't help us here, we're still performing an implicit conversion when invoking the constructor. Not sure my fix is the most elegant, open to feedback, but it does ensure we can still perform explicit `.as<std::complex<int>>()` with non-integral-rep Quantities, which I believe is the design goal.

GCC10:
```
In file included from au/quantity_test.cc:15:
./au/quantity.hh: In instantiation of 'constexpr auto au::Quantity<UnitT, RepT>::as(NewUnit) const [with NewRep = std::complex<int>; NewUnit = au::QuantityMaker<au::Miles>; <template-parameter-2-3> = void; UnitT = au::Feet; RepT = double]':
au/quantity_test.cc:528:49:   required from here
./au/quantity.hh:184:77: error: conversion from 'au::Quantity<au::Feet, double>::Rep' {aka 'double'} to 'int' may change value [-Werror=float-conversion]
  184 |             static_cast<NewRep>(detail::apply_magnitude(static_cast<Common>(value_), Factor{})));
      |                                                                             ^~~~~~
```

Clang11:
```
In file included from au/quantity_test.cc:15:
./au/quantity.hh:184:77: error: implicit conversion turns floating-point number into integer: 'const au::Quantity<au::Feet, double>::Rep' (aka 'const double') to 'const std::__1::complex<int>::value_type' (aka 'const int') [-Werror,-Wfloat-conversion]
            static_cast<NewRep>(detail::apply_magnitude(static_cast<Common>(value_), Factor{})));
                                                        ~~~~~~~~~~~         ^~~~~~
au/quantity_test.cc:528:22: note: in instantiation of function template specialization 'au::Quantity<au::Feet, double>::as<std::__1::complex<int>, au::QuantityMaker<au::Miles>, void>' requested here
    const auto b = a.as<std::complex<int>>(miles);
                     ^
1 error generated.
```

Clang14:
```
In file included from au/quantity_test.cc:15:
./au/quantity.hh:184:77: error: implicit conversion turns floating-point number into integer: 'const au::Quantity<au::Feet, double>::Rep' (aka 'const double') to 'const std::complex<int>::value_type' (aka 'const int') [-Werror,-Wfloat-conversion]
            static_cast<NewRep>(detail::apply_magnitude(static_cast<Common>(value_), Factor{})));
                                                        ~~~~~~~~~~~         ^~~~~~
au/quantity_test.cc:528:22: note: in instantiation of function template specialization 'au::Quantity<au::Feet, double>::as<std::complex<int>, au::QuantityMaker<au::Miles>, void>' requested here
    const auto b = a.as<std::complex<int>>(miles);
                     ^
1 error generated.
```

cmake/HeaderOnlyLibrary.cmake
```diff
@@ -43,7 +43,7 @@ function(header_only_library)
     ${ARG_NAME}
     INTERFACE
     FILE_SET HEADERS
-    BASE_DIRS "${PROJECT_SOURCE_DIR}/au/code"
+    BASE_DIRS "${PROJECT_SOURCE_DIR}"
```
chiphogg, line 46: Maybe for some reason I had thought (in #266) that we shouldn't set `BASE_DIRS` to `${PROJECT_SOURCE_DIR}`?

Maybe because `cmake` would then be a subfolder of the project source, including `cmake/build`, where the build lives?

I don't _think_ that's actually a problem.  I'm just doing my best to try to remember why we didn't simply do this in the first place.

au/utility/string_constant.hh
matthew-reynolds: Split to https://github.com/aurora-opensource/au/pull/444

au/quantity.hh
matthew-reynolds: Split to https://github.com/aurora-opensource/au/pull/445

hoffbrinkle (approved):

au/BUILD.bazel
```diff
@@ -589,20 +566,19 @@ cc_test(
 cc_library(
     name = "stdx",
     srcs = glob([
-        "code/au/stdx/*.cc",
-        "code/au/stdx/experimental/*.cc",
+        "stdx/*.cc",
+        "stdx/experimental/*.cc",
```
chiphogg, line 570: lol... Why does this exist?  Looks like I added it in #8.  Really weird.

I'll remove it in a follow on PR.

chiphogg (approved):
Wow --- thanks for doing this!

PR #441, opened 2025-07-02
Remove MSVC 2019 workflow

matthew-reynolds:
GitHub deprecated the Windows Server 2019 image on 2025-06-01 and removed it on 2025-06-30. See https://github.com/actions/runner-images/issues/12045.

Fortunately, we already have a Windows 2022 workflow, so we can safely drop 2019 without much concern.


chiphogg (changes requested):
I agree with the general strategy of making a PR to remove 2019 first, and then considering whether to add 2025 in a follow on PR.  This should get us unblocked ASAP.

To land this PR, we'll need to update `README.md` as well to remove the 2019 badge, and also `docs/supported-compilers.md` to move this to a lower category of support.  (We can save `docs/troubleshooting.md` for later and do a fuller update.)

Also, we apparently can't land this yet anyway, because for some reason the Checks tab is expecting MSVC 2019 results to be reported.  I've reached out to @timhirsh to see if he knows what to do.  I poked around in the settings for a little bit, but didn't see any obvious moves.

chiphogg:
OK, I removed the requirement for MSVC 2019!  So we should be good to go as soon as you make the doc updates.

chiphogg (approved):
Thanks for getting us unblocked!

Issue #429, opened 2025-05-13
Move UnitAvoidance out of `detail`


chiphogg:
We are telling people to use it in the troubleshooting guide, so we need to make it part of our public interface.

Blocks #412.

PR #443, opened 2025-07-04
Move UnitAvoidance Out Of Detail Issue 429

geoffviola:
# Description

Move `UnitAvoidance` out of the `detail` namespace. This communicates to users that they can use it in their code.

# Test

```
bazel test //au:unit_of_measure_test
```

# Issue

https://github.com/aurora-opensource/au/issues/429

### DIFF
--- a/au/code/au/unit_of_measure.hh
+++ b/au/code/au/unit_of_measure.hh
@@ -1060,6 +1060,40 @@ constexpr const auto &unit_label(Unit) {
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
+// `UnitAvoidance`: Support for customizing unit ordering.
+//
+// "Unit avoidance" is a tiebreaker for quantity-equivalent units.  Anonymous units, such as
+// `UnitImpl<...>`, `ScaledUnit<...>`, and `UnitProduct<...>`, are more "avoidable" than units which
+// are none of these, because the latter are likely explicitly named and thus more user-facing.  The
+// relative ordering among these built-in template types is probably less important than the fact
+// that there _is_ a relative ordering among them (because we need to have a strict total ordering).
+//
+// Users can specialize UnitAvoidance for their own units to customize the ordering behavior.
+template <typename T>
+struct UnitAvoidance : std::integral_constant<int, 0> {};
+
+template <typename... Ts>
+struct UnitAvoidance<UnitProduct<Ts...>> : std::integral_constant<int, 1> {};
+
+template <typename... Ts>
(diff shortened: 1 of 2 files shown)


chiphogg (changes requested):
Thanks for putting up this PR!  But as written, it'll break any code out in the wild that's using `UnitAvoidance`.  I was hoping to provide a smoother upgrade path.  I hadn't worked out all the details yet, but this would probably be some new kind of "unit tiebreaker" trait, maybe `au::UnitOrderTiebreaker` or something like that.  The sequence would be:

1. Tell people to start using `UnitOrderTiebreaker` instead.
2. Move "permanent" (library internal) `UnitAvoidance` use cases to some new trait (I dunno... `au::detail::CoarseUnitOrdering`?), and have `UnitOrderTiebreaker` default to `av::detail::UnitAvoidance` _if it exists_ (which would then be only cases in the wild).
3. In a minor version bump, mark `au::detail::UnitAvoidance` as `[[deprecated]]`.
4. In the next minor version bump, delete `au::detail::UnitAvoidance`.

Let me know if this plan makes sense.  I didn't get a lot of sleep last night. 🙃 And I've been hammering on the overflow/truncation refactor (and making great progress!).

geoffviola:
> this would probably be some new kind of "unit tiebreaker" trait, maybe au::UnitOrderTiebreaker or something like that.
...
> Move "permanent" (library internal) UnitAvoidance use cases to some new trait (I dunno... au::detail::CoarseUnitOrdering?), and have UnitOrderTiebreaker default to av::detail::UnitAvoidance if it exists (which would then be only cases in the wild).

`UnitAvoidance` isn't used outside of unit_of_measure.hh in au. What would be the difference between `CoarseUnitOrdering` and `UnitOrderTiebreaker`? We may be able to use `void_t` to default to `UnitAvoidance` if it exists.

chiphogg:
> `UnitAvoidance` isn't used outside of unit_of_measure.hh in au. What would be the difference between `CoarseUnitOrdering` and `UnitOrderTiebreaker`? We may be able to use `void_t` to default to `UnitAvoidance` if it exists.

This is really non-obvious --- sorry for not elaborating my thoughts here sooner.  It's a lexicographic ordering thing.  `UnitAvoidance` (or `CoarseUnitOrdering`) is the primary sorting key, whereas `UnitOrderTiebreaker` would be the very lowest priority sorting key, a kind of "just prevent compiler errors" ordering.

It's a pretty minor point, but it does show up in user-facing circumstances, e.g., unit labels.

geoffviola:
I don't have the cycles to follow up at the moment. It sounds like there needs to be more thought into making the implementation not break backwards compatibility and make some progress.

PR #450, opened 2025-07-10
Add abstract operations

chiphogg:
We add special types for the kinds of operations that can occur in unit
conversions.  Later on, we'll create traits that judge the risks of
these operations for overflow and truncation, and can even determine
which individual values will be lossy in these ways.

For now, the operations are:

- `StaticCast<T, U>` represents a static cast from type `T` to `U`.
- `MultiplyTypeBy<T, M>` represents multiplying a value of `T` by the
  magnitude `M`.
- `DivideTypeByInteger<T, M>` represents dividing a value of `T` by the
  magnitude `M`, which must be an integer.
- `OpSequence<Ops...>` chains any sequence of operations, and is
  automatically flattened to a single layer.

We also provide the `OpInput<Op>` and `OpOutput<Op>` traits.  For
`OpSequence`, we also explicitly check that the output type of one
operation equals the input type of the next.  This has helped prevent a
few errors already.

Helps #349.

### DIFF
--- a/au/abstract_operations_test.cc
+++ b/au/abstract_operations_test.cc
@@ -1,0 +1,129 @@
+// Copyright 2025 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "au/abstract_operations.hh"
+
+#include "au/testing.hh"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+using ::testing::StaticAssertTypeEq;
+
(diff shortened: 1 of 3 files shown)


au/abstract_operations_test.cc
```diff
@@ -0,0 +1,129 @@
...
+
+TEST(DivideTypeByInteger, IntegerTypeCanBeDividedByIntegerMag) {
+    EXPECT_THAT((DivideTypeByInteger<uint16_t, decltype(mag<3>())>::apply_to(uint16_t{16})),
+                SameTypeAndValue(uint16_t{5}));
+}
+
+TEST(DivideTypeByInteger, IntegerTypeDividedByIntegerTooBigToRepresentGivesZero) {
+    EXPECT_THAT((DivideTypeByInteger<uint8_t, decltype(mag<256>())>::apply_to(uint8_t{1})),
+                SameTypeAndValue(uint8_t{0}));
+}
```
hoffbrinkle, line 81: Would it make sense for an additional case that tests dividing to a result of 0 without overflowing the result type?

```c++
TEST(DivideTypeByInteger, IntegerTypeDividedByIntegerMagGreaterThanDividend) {
    EXPECT_THAT((DivideTypeByInteger<uint8_t, decltype(mag<2>())>::apply_to(uint8_t{1})),
                SameTypeAndValue(uint8_t{0}));
}
```

chiphogg: Done, thanks!

PR #452, opened 2025-07-10
Add lower-bound overflow checker for `StaticCast`

chiphogg:
The new `:overflow_boundary` target will check any operation (from
`:abstract_operations`) and find the minimum and maximum acceptable
value.  Here, we are starting with `StaticCast` only, and `MinGood`
only, to get us up and running.  (The final implementation for all
operations is currently almost 900 lines of code and almost 1600 lines
of tests!)

Strategically, we're thinking about non-arithmetic reps too (#52), even
though we're not ready to actually support them formally.  The strategy
will be to ask any supplier of a non-arithmetic rep to carefully
specialize `std::numeric_limits`.  Then, we'll rely on the answers they
give to judge overflow or truncation risk.  If they don't give answers,
we'll just assume conversions are risky.  `std::numeric_limits` is
probably pretty flawed in general, but I think this will be far more
workable than asking users of a non-arithmetic rep to specialize Au
specific traits.

One other important distinction is that these traits don't return their
values in `OpInput<T>`, but in `RealPart<OpInput<T>>`.  If we think
about, say, a full complex number, then notions of "max" and "min" don't
even make sense, because complex numbers are not ordered!  In fact, we
expect to replace `RealPart<T>` with a more general notion of
`ScalarPart<T>` in order to support #70, because the same considerations
apply to vectors and matrices.  But we will save that for when we are
ready to tackle #70 in earnest.

Helps #349.

### DIFF
--- a/au/overflow_boundary.hh
+++ b/au/overflow_boundary.hh
@@ -1,0 +1,212 @@
+// Copyright 2025 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <limits>
+
+#include "au/abstract_operations.hh"
+#include "au/magnitude.hh"
+#include "au/stdx/type_traits.hh"
+
+// These utilities help assess overflow risk for an operation `Op` by finding the minimum and
+// maximum values in the "scalar type" of `OpInput<Op>` that are guaranteed to not overflow.
+//
+// The "scalar type" of `T` is usually just `T`, but if `T` is something like `std::complex<U>`, or
+// `Eigen::Vector<U, N>`, then it would be `U`.
+
+namespace au {
+namespace detail {
+
+//
+// `MinGood<Op>::value()` is a constexpr constant of the "scalar type" for `OpInput<Op>` that is the
+// minimum value that does not overflow.
+//
+// IMPORTANT: the result must always be non-positive.  The code is structured on this assumption.
+//
+template <typename Op, typename Limits>
+struct MinGoodImpl;
+template <typename Op, typename Limits = void>
+using MinGood = typename MinGoodImpl<Op, Limits>::type;
+
+//
+// `MaxGood<Op>::value()` is a constexpr constant of the "scalar type" for `OpInput<Op>` that is the
+// maximum value that does not overflow.
+//
+// IMPORTANT: the result must always be non-negative.  The code is structured on this assumption.
+//
+template <typename Op, typename Limits = void>
+struct MaxGoodImpl;
+template <typename Op, typename Limits = void>
+using MaxGood = typename MaxGoodImpl<Op, Limits>::type;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// IMPLEMENTATION DETAILS
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+// The implementation strategy will be to decompose to increasingly specific cases, using
+// `std::conditional` constructs that are _at most one layer deep_.  This should keep every
+// individual piece as easy to understand as possible, although it does mean we'll tend to be
+// navigating many layers deep from the top-level API to the ultimate implementation.
+//
+// It's easier to navigate these helpers if we put a shorthand comment at the top of each.  Here's
+// the key:
+//
+// (A) = arithmetic (integral or floating point)
+// (F) = floating point
+// (I) = integral (signed or unsigned)
+// (N) = non-arithmetic
+// (S) = signed integral
+// (U) = unsigned integral
+// (X) = any type
+
+// `LowerLimit<T, Limits>::value()` returns `Limits::lower()` (assumed to be of type `T`), unless
+// `Limits` is `void`, in which case it means "no limit" and we return the lowest possible value.
+template <typename T, typename Limits>
+struct LowerLimit {
+    static constexpr T value() { return Limits::lower(); }
+};
+template <typename T>
+struct LowerLimit<T, void> {
+    static constexpr T value() { return std::numeric_limits<T>::lowest(); }
+};
+
+// Inherit from this struct to produce a compiler error in case we try to use a combination of types
+// that isn't yet supported.
+template <typename T>
+struct OverflowBoundaryNotYetImplemented {
+    struct NotYetImplemented {};
+    static_assert(std::is_same<T, NotYetImplemented>::value,
+                  "Overflow boundary not yet implemented for this type.");
+};
+
+// A type whose `::value()` function returns `0`, expressed in `T`.
+template <typename T>
+struct ValueIsZero {
+    static constexpr T value() { return T{0}; }
+};
+
+// A type whose `::value()` function returns the higher of `std::numeric_limits<T>::lowest()`, or
+// `LowerLimit<U, ULimit>` expressed in `T`.  Assumes that `U` is more expansive than `T`, so that
+// we can cast everything to `U` to do the comparisons.
+template <typename T, typename U, typename ULimit>
+struct ValueIsSourceLowestUnlessDestLimitIsHigher {
+    static constexpr T value() {
+        constexpr auto LOWEST_T_IN_U = static_cast<U>(std::numeric_limits<T>::lowest());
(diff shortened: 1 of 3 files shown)


au/overflow_boundary.hh
```diff
@@ -0,0 +1,212 @@
...
+struct LowerLimit<T, void> {
+    static constexpr T value() { return std::numeric_limits<T>::lowest(); }
+};
+
+// Inherit from this struct to produce a compiler error in case we try to use a combination of types
+// that isn't yet supported.
+template <typename T>
+struct OverflowBoundaryNotYetImplemented {
+    struct NotYetImplemented {};
+    static_assert(std::is_same<T, NotYetImplemented>::value,
```
hoffbrinkle, line 91: Should we be including `<type_traits>` if we're going to use `std::is_same`?

au/overflow_boundary.hh
```diff
@@ -0,0 +1,212 @@
...
+                  "Overflow boundary not yet implemented for this type.");
+};
+
+// A type whose `::value()` function returns `0`, expressed in `T`.
+template <typename T>
+struct ValueIsZero {
+    static constexpr T value() { return T{0}; }
```
hoffbrinkle: For these `Is` structs:

`ValueIsZero` implies that we're answering a question ("true or false, the value is zero?").  Would `ValueZero` be acceptable?

au/overflow_boundary.hh
```diff
@@ -0,0 +1,212 @@
...
+};
+
+// A type whose `::value()` function returns the higher of `std::numeric_limits<T>::lowest()`, or
+// `LowerLimit<U, ULimit>` expressed in `T`.  Assumes that `U` is more expansive than `T`, so that
+// we can cast everything to `U` to do the comparisons.
```
hoffbrinkle, line 97: Can we add a `static_assert` to verify this assumption?

au/overflow_boundary.hh
```diff
@@ -0,0 +1,212 @@
...
+template <typename T, typename U, typename ULimit>
+struct ValueIsSourceLowestUnlessDestLimitIsHigher {
+    static constexpr T value() {
+        constexpr auto LOWEST_T_IN_U = static_cast<U>(std::numeric_limits<T>::lowest());
+        constexpr auto U_LIMIT = LowerLimit<U, ULimit>::value();
+        return (LOWEST_T_IN_U <= U_LIMIT) ? static_cast<T>(U_LIMIT)
+                                          : std::numeric_limits<T>::lowest();
```
hoffbrinkle, line 104: Can we we leverage `max` here?

au/overflow_boundary_test.cc
```diff
@@ -0,0 +1,285 @@
...
+namespace detail {
+namespace {
+
+template <typename T>
+struct NoUpperLimit {
+    static constexpr T upper() { return std::numeric_limits<T>::max(); }
+};
+
+template <typename T>
+struct LowerLimitIsZero : NoUpperLimit<T> {
```
hoffbrinkle: Similar to above, `LowerLimitZero`?

au/overflow_boundary_test.cc
```diff
@@ -0,0 +1,285 @@
...
+using ::testing::IsTrue;
+using ::testing::StaticAssertTypeEq;
+
+namespace au {
+namespace detail {
+namespace {
+
+template <typename T>
+struct NoUpperLimit {
+    static constexpr T upper() { return std::numeric_limits<T>::max(); }
```
hoffbrinkle, line 36: Include `<limits>`.

au/overflow_boundary_test.cc
```diff
@@ -0,0 +1,285 @@
...
+                Eq(MinGood<StaticCast<int64_t, uint32_t>>::value()));
+}
+
+TEST(StaticCast, MinGoodUnchangedWithExplicitLimitLessConstrainingThanExistingResult) {
+    // In these cases, we are applying a non-trivial lower limit (i.e., it is higher than the
+    // `lowest()` value), but it does not constrain the result enough to change it.
+
+    struct DoubleLimitTwiceFloatLowest : NoUpperLimit<double> {
+        static constexpr double lower() {
+            return static_cast<double>(std::numeric_limits<float>::lowest()) * 2.0;
```
hoffbrinkle, line 177: In this case, multiplying by `2` as opposed to `2.0` would give the same result, correct?

```suggestion
            return static_cast<double>(std::numeric_limits<float>::lowest()) * 2;
```

au/overflow_boundary_test.cc
```diff
@@ -0,0 +1,285 @@
...
+                Eq(MinGood<StaticCast<float, double>>::value()));
+
+    EXPECT_THAT((MinGood<StaticCast<int32_t, double>, DoubleLimitTwiceFloatLowest>::value()),
+                Eq(MinGood<StaticCast<int32_t, double>>::value()));
+
+    EXPECT_THAT((MinGood<StaticCast<uint16_t, double>, DoubleLimitTwiceFloatLowest>::value()),
+                Eq(MinGood<StaticCast<uint16_t, double>>::value()));
+
+    struct FloatLimitHalfFloatLowest : NoUpperLimit<float> {
+        static constexpr float lower() { return std::numeric_limits<float>::lowest() / 2.0f; }
```
hoffbrinkle, line 191: Same thing, we could just divide by `2`?

au/overflow_boundary_test.cc
```diff
@@ -0,0 +1,285 @@
...
+    EXPECT_THAT((MinGood<StaticCast<int32_t, uint64_t>, LowerLimitIsZero<uint64_t>>::value()),
+                Eq(MinGood<StaticCast<int32_t, uint64_t>>::value()));
+
+    EXPECT_THAT((MinGood<StaticCast<double, uint32_t>, LowerLimitIsZero<uint32_t>>::value()),
+                Eq(MinGood<StaticCast<double, uint32_t>>::value()));
+}
+
+TEST(StaticCast, MinGoodCappedByExplicitFloatLimit) {
+    struct FloatLowerLimitMinusOne : NoUpperLimit<float> {
+        static constexpr float lower() { return -1.0f; }
```
hoffbrinkle, line 224: I suppose here as well, we could just use `-1`?

```suggestion
        static constexpr float lower() { return -1; }
```

au/overflow_boundary.hh
chiphogg: Yep, done.

au/overflow_boundary.hh
chiphogg: I feel like I still want a word in there.  Maybe `Of` would be better than `Is`?  I tried it out.  Let me know what you think.  If it still feels wrong to you, then I can take it out and do `ValueZero`, `ValueLowestInDestination`, etc.

Incidentally, the `Of` name change forces me to move `ValueOfZero` upstream into `"zero.hh"`, because there is a character-for-character identical utility in `:apply_rational_magnitude_to_integral_test`.  This feels like good support for this naming convention.

au/overflow_boundary.hh
chiphogg: I thought of that, but I struggled with how to express that `static_assert` in a generic and clearly correct manner.  Ultimately, I think it turns out to be a little redundant anyway, because `static_cast<U>(std::numeric_limits<T>::lowest())` should fail if this is ever violated.

au/overflow_boundary.hh
chiphogg: I don't think so, because `U_LIMIT` may not be cleanly representable in `T`.  As long as `static_cast<T>(U_LIMIT)` is not UB in those cases, I'm confident that comparing the values in "U-space" will produce the right answer.  I would be a lot less confident if we did the max in "T-space".

au/overflow_boundary_test.cc
chiphogg: Done.

au/overflow_boundary_test.cc
chiphogg: Done.

au/overflow_boundary_test.cc
chiphogg: Yes, but I preferred to take an explicitly single-type approach, rather than mixing `double` and `int`.  This explains my approach in the other two situations you commented on below.  Is there a compelling reason to switch to the "int-shaped" versions, other than conciseness?

au/overflow_boundary.hh
hoffbrinkle: Ah, makes sense.

au/overflow_boundary_test.cc
hoffbrinkle: Not really, just can't believe you typed all those extra characters (".0"). 😜

au/overflow_boundary.hh
hoffbrinkle: Definitely like "of" over "is."  👍 And "of" is just fine.  As you may have noticed, I just tend to lean towards minimalism, which isn't necessarily the best way... 🙃

hoffbrinkle (approved):

PR #454, opened 2025-07-13
Add upper-bound overflow checker for `StaticCast`

chiphogg:
This completes our overflow checking capabilities for `StaticCast`, in
the obvious way.

Note that we have a different set of subtleties to wrestle with than for
the min.  For the min, we had to deal with unsigned types having a lower
bound of 0, although the signed types' lower bounds were nice powers of
two.  For the max, in this PR, we have to watch out for the upper bounds
being _slightly lower_ than a nice power of two.  This has surprising
consequences, such as the fact that you can cast an int type's max to a
floating point type, but you _can't cast it back_.  We deal with these
as best we can, bearing in mind that many common tools for dealing with
floating point types are not available in `constexpr` for C++14.

Helps #349.

### DIFF
--- a/au/overflow_boundary.hh
+++ b/au/overflow_boundary.hh
@@ -72,6 +72,17 @@ using MaxGood = typename MaxGoodImpl<Op, Limits>::type;
 // (U) = unsigned integral
 // (X) = any type
 
+// `UpperLimit<T, Limits>::value()` returns `Limits::upper()` (assumed to be of type `T`), unless
+// `Limits` is `void`, in which case it means "no limit" and we return the highest possible value.
+template <typename T, typename Limits>
+struct UpperLimit {
+    static constexpr T value() { return Limits::upper(); }
+};
+template <typename T>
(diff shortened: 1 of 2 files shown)


au/overflow_boundary.hh
```diff
@@ -114,6 +138,67 @@ struct ValueOfLowestInDestination {
...
+        }
+
+        Float x = MAX_MANTISSA;
+        while (x + x < LIMIT) {
+            x += x;
+        }
+        return x;
+    }
+
+    // `value()` implementation simply computes the result _once_ (caching it), and then returns it.
```
hoffbrinkle, line 193: Where does the caching come in?  Is it just something that compiler can do when the result is a `constexpr`?

hoffbrinkle (approved):

au/overflow_boundary.hh
chiphogg: Yes, that's what I was going for here.

PR #455, opened 2025-07-14
Implement `MinGood<MultiplyTypeBy<T, M>>`

chiphogg:
If `T` is unsigned, this should always be 0 (because 0 is the minimum
_possible_, and 0 is _always_ "good").  Otherwise, we check whether `M`
is "the sort of thing" that can be represented in `T`.  If not, well, we
just return 0 as a placeholder.  If it _is_ compatible, we _divide_ the
limits by the magnitude, and take whichever is lower --- unless the
result would be lower than the minimum value of the type, in which case
we clamp it.

Helps #349.

### DIFF
--- a/au/overflow_boundary.hh
+++ b/au/overflow_boundary.hh
@@ -56,6 +56,8 @@ using MaxGood = typename MaxGoodImpl<Op, Limits>::type;
 // IMPLEMENTATION DETAILS
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// General note:
+//
 // The implementation strategy will be to decompose to increasingly specific cases, using
 // `std::conditional` constructs that are _at most one layer deep_.  This should keep every
 // individual piece as easy to understand as possible, although it does mean we'll tend to be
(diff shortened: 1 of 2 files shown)


au/overflow_boundary.hh
```diff
@@ -199,6 +238,82 @@ struct ValueOfMaxFloatNotExceedingMaxInt {
...
 
+template <typename T, typename MagT, MagRepresentationOutcome>
+struct MagHelper {
+    static constexpr T div(const T &, const T &) {
+        // We assume the most likely reason to be here is that the magnitude was too large to
+        // represent in the type, so we treat this as a "divide by infinity" and return zero.
+        //
+        // Really, this should probably just not be called, but it definitely needs to exist for the
+        // code to compile.
+        return T{0};
```
hoffbrinkle: Should this just be an assert?

chiphogg: Rereading the comment, I can't account for why I thought this "shouldn't be called".  This seems like a perfectly valid use case.  I've updated it to add the `static_assert`, and fixed up the comment to explain the reasoning.

PR #456, opened 2025-07-16
Implement `MaxGood<MultiplyTypeBy<T, M>>`

chiphogg:
The main hurdle here is that for signed integral types, the minimum
value is more negative than the max value is positive.  To deal with
this, we add a `clamped_negate(T)` function, which negates the input,
but saturates at the limits of the type.  We use it consistently in a
few other places where it makes sense.

We also add an `equal` function to the `MagHelper`, so that we can
seamlessly check whether a value in a given type happens to equal a
`Magnitude`.  This again helps us deal with this hurdle that we have
with signed integral types.

With these helpers in hand, we can implement `MaxGood` using a very
similar strategy as for `MinGood`.  The only substantive difference is
that `MinGood` automatically returns `0` for all unsigned types period,
but `MaxGood` only does this for an unsigned type _with a negative scale
factor_.

Helps #349.

### DIFF
--- a/au/overflow_boundary.hh
+++ b/au/overflow_boundary.hh
@@ -19,6 +19,7 @@
 
 #include "au/abstract_operations.hh"
 #include "au/magnitude.hh"
+#include "au/operators.hh"
 #include "au/stdx/type_traits.hh"
 
 // These utilities help assess overflow risk for an operation `Op` by finding the minimum and
@@ -133,6 +134,17 @@ struct LowerLimit<T, void> {
     static constexpr T value() { return std::numeric_limits<T>::lowest(); }
 };
 
+template <typename T>
+constexpr T clamped_negate(T x) {
+    if (Less{}(x, T{0}) && Less{}(x, -std::numeric_limits<T>::max())) {
+        return std::numeric_limits<T>::max();
+    }
+    if (Greater{}(x, T{0}) && Greater{}(x, clamped_negate(std::numeric_limits<T>::lowest()))) {
+        return std::numeric_limits<T>::lowest();
+    }
+    return -x;
+}
+
 // Inherit from this struct to produce a compiler error in case we try to use a combination of types
 // that isn't yet supported.
 template <typename T>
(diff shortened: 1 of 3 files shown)


au/overflow_boundary.hh
```diff
@@ -304,6 +320,53 @@ struct ClampLowestOfLimitsTimesInverseValue {
...
 };
 
+template <typename T, typename... BPs>
+constexpr bool mag_representation_equals(const T &x, Magnitude<BPs...> m) {
+    constexpr auto result = get_value_result<T>(m);
+    return MagHelper<T, Magnitude<BPs...>, result.outcome>::equal(x, result.value);
+}
+
+// Name reads as "highest of (limits divided by value)".  Remember that the value can be negative,
+// so we just take whichever limit is larger _after_ dividing.  And since `Abs<M>` can be assumed to
```
hoffbrinkle: I was having trouble following this comment.  The "large _after_ dividing" only applies in the negative value case, correct?

```suggestion
// in which case we just take whichever limit is larger _after_ dividing.  And since `Abs<M>` can be assumed to
```

au/overflow_boundary_test.cc
```diff
@@ -915,6 +925,255 @@ TEST(MultiplyTypeBy, MinGoodForComplexOfTProvidesAnswerAsT) {
...
+
+    EXPECT_THAT(max_good_value(multiply_type_by<int8_t>(mag<63>())), SameTypeAndValue(int8_t{2}));
+
+    EXPECT_THAT(max_good_value(multiply_type_by<int8_t>(mag<64>())), SameTypeAndValue(int8_t{1}));
+
+    EXPECT_THAT(max_good_value(multiply_type_by<int8_t>(mag<127>())), SameTypeAndValue(int8_t{1}));
+}
+
+TEST(MultiplyTypeBy, MaxGoodForUnlimitedSignedTimesNegIntIsLowerLimitDivByMag) {
+    EXPECT_THAT(max_good_value(multiply_type_by<int8_t>(-mag<1>())), SameTypeAndValue(int8_t{127}));
```
hoffbrinkle, line 967: Man, this one is a bit of mind bender until I really thought about it.  Having a use case for negative magnitudes really helped me reason about it.

au/overflow_boundary.hh
chiphogg: This reword didn't seem quite right, because we _always_ take the limit that is larger _after_ dividing.  I went back and gave the comment a more thorough rewrite.

au/overflow_boundary_test.cc
chiphogg: This whole effort has been one mind bender after another!  It was such a relief to get to a working proof of concept in my local client after multiple weeks of not knowing how the ideas would work in practice.

Also, I don't even want to think about how many lines of code we need in various places to deal with the fact that a signed integer's most negative value is further from zero than its most positive value.  😅

PR #458, opened 2025-07-16
Implement limits for `OpSequence`

chiphogg:
Now it all comes together.  We carefully designed the overflow checkers
for each individual operation so that they took a `Limits` parameter,
which might be different from the full limits for the type.  This PR
adds a `LimitsFor<Op, Limits>` helper, which takes an "output target"
limits type `Limits`, and an operation `Op`, and produces the _input
limits_ that will stay within that target.  And this object can be used
as the "limits" parameter _for the previous operation in the chain_!
And we're done.

Helps https://github.com/aurora-opensource/au/issues/349.

### DIFF
--- a/au/overflow_boundary.hh
+++ b/au/overflow_boundary.hh
@@ -145,6 +145,13 @@ constexpr T clamped_negate(T x) {
     return -x;
 }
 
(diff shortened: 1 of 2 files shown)


hoffbrinkle (approved):
It's that easy?

PR #460, opened 2025-07-17
Add `:conversion_strategy` target

chiphogg:
This new target is the One True Home for choosing each unit conversion
implementation.  It depends only on the initial rep, final rep, and
conversion factor.  The output is a tailor-made and trimmed sequence of
operations for that conversion.  It will automatically take care of:

- Type promotion
- Using division for pure-inverse-integer factors
- Using multiply-then-divide for applying a rational to an integral type

When all is said and done for #349, each conversion implementation will
be able to simply ask for the right conversion for the reps and factor,
and just apply it blindly.  And when we want to reason about overflow
and/or truncation risk for those conversions, we can do so by checking
this sequence of operations!

### DIFF
--- a/au/conversion_strategy.hh
+++ b/au/conversion_strategy.hh
@@ -1,0 +1,108 @@
+// Copyright 2025 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


au/conversion_strategy.hh
```diff
@@ -0,0 +1,108 @@
...
+#include "au/abstract_operations.hh"
+#include "au/magnitude.hh"
+#include "au/stdx/type_traits.hh"
+
+namespace au {
+namespace detail {
+
+//
+// `ConversionForRepsAndFactor<OldRep, NewRep, Factor>` is the operation that takes a value of
+// `OldRep`, and produces the product of that value with magnitude `Factor` in the type `NewRep`.
```
hoffbrinkle, line 26: "product" could be construed as multiplication?

chiphogg: Yep: that's intentional.  _Logically_, that is always what we are doing.  _Implementation wise_, we may of course choose a different sequence of operations (such as dividing by the inverse when that is an exact integer, or else breaking up a fraction into separate steps for numerator and denominator).

PR #461, opened 2025-07-17
Add fuzz-style explorer for conversion lossiness

chiphogg:
This is a simple tool to randomly generate values in a wide variety of
types, and try applying a bunch of unit conversions.  We make sure that
the round trip value changes if and only if we expect the conversion to
be lossy.

Of course, there are a variety of edge case ways for this to be
mismatched that we deem to be OK.  For example, floating point values
can be "subnormal", and we want to consider those out of scope.
Additionally, we assume that a floating point conversion can differ from
the best value by up to 1 ULP, so a round trip can be off by up to 2
ULP.  There are a few other carveouts of this nature.

If we find a mismatch, we terminate and immediately print out details
about the specific test case.  The expectation is that we will either
add a test case to the library and fix the implementation, or else add a
carveout to this utility.  The goal is to get to a state where we can
run continuously and not find any unaccepted anomalies.

Overall, this tool should greatly increase our confidence that our
runtime overflow and truncation checkers are working correctly.

Usage:

```sh
bazel run //fuzz:quantity_runtime_conversion_check
```

Helps #349.

### DIFF
--- a/fuzz/BUILD.bazel
+++ b/fuzz/BUILD.bazel
@@ -1,0 +1,51 @@
+# Copyright 2025 Aurora Operations, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
+
+cc_binary(
+    name = "quantity_runtime_conversion_check",
+    testonly = True,
+    srcs = ["quantity_runtime_conversion_check.cc"],
+    deps = [
+        ":quantity_runtime_conversion_checkers",
+        "//au:conversion_strategy",
+        "//au:testing",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
+cc_library(
+    name = "quantity_runtime_conversion_checkers",
+    testonly = True,
+    hdrs = ["quantity_runtime_conversion_checkers.hh"],
+    deps = [
+        "//au",
+        "//au:abstract_operations",
+        "@com_google_googletest//:gtest",
+    ],
+)
+
+cc_test(
+    name = "quantity_runtime_conversion_checkers_test",
+    size = "small",
+    srcs = ["quantity_runtime_conversion_checkers_test.cc"],
+    deps = [
+        ":quantity_runtime_conversion_checkers",
+        "//au",
+        "//au:conversion_strategy",
+        "//au:testing",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
--- a/fuzz/quantity_runtime_conversion_check.cc
+++ b/fuzz/quantity_runtime_conversion_check.cc
@@ -184,12 +184,10 @@ constexpr TestCategory categorize_testing_scenario() {
     if (std::is_integral<RepT>::value) {
         return std::is_integral<DestRepT>::value ? TestCategory::INTEGRAL_TO_INTEGRAL
                                                  : TestCategory::INTEGRAL_TO_FLOAT;
-    } else {
-        return std::is_integral<DestRepT>::value ? TestCategory::FLOAT_TO_INTEGRAL
-                                                 : TestCategory::FLOAT_TO_FLOAT;
     }
-
-    return TestCategory::UNSUPPORTED;
+    
+    return std::is_integral<DestRepT>::value ? TestCategory::FLOAT_TO_INTEGRAL
+                                             : TestCategory::FLOAT_TO_FLOAT;
 }
 
 template <typename RepT,
--- a/fuzz/quantity_runtime_conversion_check.cc
+++ b/fuzz/quantity_runtime_conversion_check.cc
@@ -278,11 +278,11 @@ struct LossChecker<RepT, UnitT, DestRepT, DestUnitT, TestCategory::FLOAT_TO_FLOA
 
         std::ostringstream oss;
         oss << "Breakdown:" << std::setprecision(std::numeric_limits<RepT>::digits10 + 1u)
-            << std::endl
-            << "  Initial:    " << value << std::endl
-            << "  Min OK:     " << min_ok << std::endl
-            << "  Round trip: " << round_trip << std::endl
-            << "  Max OK:     " << max_ok << std::endl;
+            << '\n'
+            << "  Initial:    " << value << '\n'
+            << "  Min OK:     " << min_ok << '\n'
+            << "  Round trip: " << round_trip << '\n'
+            << "  Max OK:     " << max_ok << '\n';
 
         const auto result =
             ((round_trip == value)                          ? RoundTripResult::IDENTICAL
--- a/fuzz/quantity_runtime_conversion_check.cc
+++ b/fuzz/quantity_runtime_conversion_check.cc
@@ -56,7 +56,7 @@ std::string print(RoundTripResult result) {
 
 template <typename T>
 T next_higher(T x, std::size_t n = 1u) {
-    static_assert(std::is_floating_point<T>::value, "");
+    static_assert(std::is_floating_point<T>::value, "Utility only for floating point");
     while (n-- > 0u) {
         x = std::nextafter(x, std::numeric_limits<T>::infinity());
     }
@@ -65,7 +65,7 @@ T next_higher(T x, std::size_t n = 1u) {
 
 template <typename T>
 T next_lower(T x, std::size_t n = 1u) {
-    static_assert(std::is_floating_point<T>::value, "");
+    static_assert(std::is_floating_point<T>::value, "Utility only for floating point");
     while (n-- > 0u) {
         x = std::nextafter(x, -std::numeric_limits<T>::infinity());
     }
@@ -74,12 +74,12 @@ T next_lower(T x, std::size_t n = 1u) {
 
 template <typename U, typename R>
 Quantity<U, R> next_higher_quantity(Quantity<U, R> q, std::size_t n = 1u) {
-    return make_quantity<U>(next_higher(q.in(U{}), n));
+    return make_quantity<U>(next_higher(q.data_in(U{}), n));
 }
 
 template <typename U, typename R>
 Quantity<U, R> next_lower_quantity(Quantity<U, R> q, std::size_t n = 1u) {
-    return make_quantity<U>(next_lower(q.in(U{}), n));
+    return make_quantity<U>(next_lower(q.data_in(U{}), n));
 }
 
 constexpr std::size_t MAX_DIST = 1000u;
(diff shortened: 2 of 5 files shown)


fuzz/BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+cc_binary(
+    name = "quantity_runtime_conversion_check",
+    testonly = True,
+    srcs = ["quantity_runtime_conversion_check.cc"],
+    deps = [
+        ":quantity_runtime_conversion_checkers",
+        "//au:conversion_strategy",
+        "//au:testing",
+        "@com_google_googletest//:gtest_main",
+    ],
```
hoffbrinkle, line 26: Should this be a `cc_test`?

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+            return "IGNORE_SUBNORMALS";
+        case RoundTripResult::CAST_NON_INT_TO_INT_IS_DEFINITELY_LOSSY:
+            return "CAST_NON_INT_TO_INT_IS_DEFINITELY_LOSSY";
+    }
+    return "UNKNOWN";
+}
+
+template <typename T>
+T next_higher(T x, std::size_t n = 1u) {
+    static_assert(std::is_floating_point<T>::value, "");
```
hoffbrinkle: Should we have something in the `static_assert` string?

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+    while (n-- > 0u) {
```
hoffbrinkle, line 60: Can you avoid the signed/unsigned danger entirely by dropping the comparison?

```suggestion
    while (n--) {
```

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+        x = std::nextafter(x, -std::numeric_limits<T>::infinity());
+    }
+    return x;
+}
+
+template <typename U, typename R>
+Quantity<U, R> next_higher_quantity(Quantity<U, R> q, std::size_t n = 1u) {
+    return make_quantity<U>(next_higher(q.in(U{}), n));
```
hoffbrinkle: Could you use `.data_in` here?

```suggestion
    next_higher(q.data_in(U{}), n));
    return q;
```

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+    static_assert(std::is_integral<DestRepT>::value || std::is_floating_point<DestRepT>::value, "");
+    if (std::is_integral<RepT>::value) {
+        return std::is_integral<DestRepT>::value ? TestCategory::INTEGRAL_TO_INTEGRAL
+                                                 : TestCategory::INTEGRAL_TO_FLOAT;
+    } else {
+        return std::is_integral<DestRepT>::value ? TestCategory::FLOAT_TO_INTEGRAL
+                                                 : TestCategory::FLOAT_TO_FLOAT;
+    }
+
+    return TestCategory::UNSUPPORTED;
```
hoffbrinkle: The `else` is redundant, which then makes it obvious that the UNSUPPORTED is impossible.

```suggestion
    }
    
    return std::is_integral<DestRepT>::value ? TestCategory::FLOAT_TO_INTEGRAL
                                             : TestCategory::FLOAT_TO_FLOAT;
```

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+struct SignFlipImpl;
+template <>
+struct SignFlipImpl<true, true> {
+    template <typename U1, typename R1, typename U2, typename R2>
+    static constexpr bool assess(const Quantity<U1, R1> &, const Quantity<U2, R2> &) {
+        return false;
+    }
+};
+template <>
+struct SignFlipImpl<true, false> {
```
hoffbrinkle: Not a huge fan of booleans as parameters, as it makes it difficult to comprehend the template.  Can we use an enum for this parameter?

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+        const auto min_ok = next_lower_quantity(value, 2u);
+        const auto max_ok = next_higher_quantity(value, 2u);
+
+        std::ostringstream oss;
+        oss << "Breakdown:" << std::setprecision(std::numeric_limits<RepT>::digits10 + 1u)
+            << std::endl
+            << "  Initial:    " << value << std::endl
+            << "  Min OK:     " << min_ok << std::endl
+            << "  Round trip: " << round_trip << std::endl
+            << "  Max OK:     " << max_ok << std::endl;
```
hoffbrinkle: Some linter or something google thing somewhere recommends to not use `std::endl` as it's pretty heavy weight (forces a flush and everything).

```suggestion
            << '\n'
            << "  Initial:    " << value << '\n'
            << "  Min OK:     " << min_ok << '\n'
            << "  Round trip: " << round_trip << '\n'
            << "  Max OK:     " << max_ok << '\n';
```

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+        return {result, oss.str()};
+    }
+};
+
+template <typename RepT, typename UnitT, typename DestRepT, typename DestUnitT>
+struct LossChecker<RepT, UnitT, DestRepT, DestUnitT, TestCategory::INTEGRAL_TO_FLOAT> {
+    static LossCheck check_for_loss(const Quantity<UnitT, RepT> &value,
+                                    const Quantity<DestUnitT, DestRepT> &destination,
+                                    const Quantity<UnitT, RepT> &round_trip) {
+        auto fraction = value / 100'000;
```
hoffbrinkle, line 312: Calling back to our recent discussions on per cent and how to manage those, this is an example where it'd be slick if we could do:

```suggestion
        auto fraction = per_cent_mille(value);
```

fuzz/quantity_runtime_conversion_check.cc
```diff
@@ -0,0 +1,523 @@
...
+            << "  Min OK:     " << min_ok << std::endl
+            << "  Round trip: " << round_trip << std::endl
+            << "  Max OK:     " << max_ok << std::endl;
+
+        const auto result = ignore_because_subnormal
+                                ? RoundTripResult::IGNORE_SUBNORMALS
+                                : ((round_trip == value) ? RoundTripResult::IDENTICAL
+                                   : (round_trip < min_ok || round_trip > max_ok)
+                                       ? RoundTripResult::SIGNIFICANT_LOSS
+                                       : RoundTripResult::MINOR_LOSS);
```
hoffbrinkle: This is one of those cases where I think the parens may hurt readability.  Ternary is pretty low on the precedence order, you have to really try to get an unexpected result.

Without them:

```c++
        const auto result = ignore_because_subnormal ? RoundTripResult::IGNORE_SUBNORMALS
                            : round_trip == value    ? RoundTripResult::IDENTICAL
                            : round_trip < min_ok || round_trip > max_ok
                                ? RoundTripResult::SIGNIFICANT_LOSS
                                : RoundTripResult::MINOR_LOSS;
```

fuzz/quantity_runtime_conversion_checkers_test.cc
```diff
@@ -0,0 +1,137 @@
...
+template <typename T>
+class QuantityRuntimeConversionChecker : public ::testing::Test {};
+
+TYPED_TEST_SUITE_P(QuantityRuntimeConversionChecker);
+
+TYPED_TEST_P(QuantityRuntimeConversionChecker, RoundTripIsIdentityIffConversionNotLossy) {
+    using Rep = std::tuple_element_t<0, TypeParam>;
+    constexpr auto destination_unit = std::tuple_element_t<1, TypeParam>{};
+    RandomValueGenerator<Rep> generator{9876543210u};
+    for (auto i = 0u; i < 1'000'000u; ++i) {
```
hoffbrinkle: If you're not using `i`, I've got this idiom that I like to use:
```suggestion
    for (auto _ = 1'000'000u; --_;) {
```

fuzz/quantity_runtime_conversion_checkers_test.cc
```diff
@@ -0,0 +1,137 @@
...
+        const auto value = meters(generator.next_value());
+
+        const bool expect_loss = is_conversion_lossy(value, destination_unit);
+
+        const auto round_trip = value.coerce_as(destination_unit).coerce_as(meters);
+        const bool actual_loss = (value != round_trip);
+
+        EXPECT_EQ(expect_loss, actual_loss) << "Value: " << value << " Round trip: " << round_trip;
```
hoffbrinkle:
```suggestion
        EXPECT_THAT(actual_loss, Eq(expect_loss)) << "Value: " << value << " Round trip: " << round_trip;
```

hoffbrinkle (approved):

fuzz/BUILD.bazel
chiphogg: I see this as being (or perhaps growing into) a tool that we can run.  Maybe even as an infinite loop.  The fact that it currently uses googletest is kind of an implementation detail.

fuzz/quantity_runtime_conversion_check.cc
chiphogg: Done.

fuzz/quantity_runtime_conversion_check.cc
chiphogg: I don't think there's a danger here, because we're explicitly using unsigned types.  And, I prefer to use conditions that are already clearly `bool` rather than relying on implicit conversions from numeric types.  Yes, it's guaranteed by the standard, but it's still a convention to remember.

fuzz/quantity_runtime_conversion_check.cc
chiphogg: Nice catch!

(The `suggestion` won't work as written, because `next_higher` doesn't modify its argument, but I did use `data_in`.)

fuzz/quantity_runtime_conversion_check.cc
chiphogg: Done.

fuzz/quantity_runtime_conversion_check.cc
chiphogg: Sounds good; done.

fuzz/quantity_runtime_conversion_checkers_test.cc
chiphogg: Ha!  I dredged up the files from #346 (Dec. 2024), and fixed them up.  Not surprised that I missed one.  Thanks for catching it!

fuzz/quantity_runtime_conversion_check.cc
hoffbrinkle: Thinking more about this, I think don't think this would be quite right with our API fluency.  I think what would follow our fluency would be something like:

```c++
        auto fraction = value * per_cent_mille;
```

which I believe you gave me something like this to try, so I'll circle back to that this week...

fuzz/BUILD.bazel
hoffbrinkle: Makes sense.  If we get a chance, maybe a note in the build file in case someone is wonders that in the future.

fuzz/quantity_runtime_conversion_check.cc
hoffbrinkle: That came out pretty well, thanks!

PR #462, opened 2025-07-17
Assess truncation risk for every simple operation

chiphogg:
This means adding a new trait, `TruncationRiskFor<Op>`, which maps any
operation onto one of the truncation risk types that we added in the
previous PR.

These "simple" operations are reasonably straightforward, so I'm adding
all of them in a single PR.  Handling operation _sequences_ will be more
involved, and I expect it to take two more PRs.

Helps #349.

### DIFF
--- a/au/truncation_risk.hh
+++ b/au/truncation_risk.hh
@@ -19,6 +19,11 @@
 namespace au {
 namespace detail {
 
+template <typename Op>
+struct TruncationRiskForImpl;
+template <typename Op>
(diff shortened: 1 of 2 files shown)


hoffbrinkle (approved):

chiphogg:
This commit looks like a [phone number from a movie](https://en.wikipedia.org/wiki/555_(telephone_number)#Fictional_usage) :sweat_smile: 

image

PR #464, opened 2025-07-17
Remove tests for deprecated utilities

chiphogg:
We do not plan on keeping any of the utilities `IsAbsMagLessThanOne`,
`categorize_magnitude()` or `can_scale_without_overflow()` around, so we
delete the tests for them now.  Helps #349.


hoffbrinkle (approved):
It does feel weird to delete the tests before deleting the code.

hoffbrinkle:
> So, to sum up: small weird feeling now, but meaningfully lower cognitive load soon. 😄

I can appreciate the meaningfully lower cognitive load!

PR #465, opened 2025-07-18
Handle `std::complex` types in conversion strategy

chiphogg:
Fundamentally, this is all about the tests.  We're just adding a few
more test cases, and making them pass.

What those tests showed us is interesting.

The first test is pretty straightforward.  It showed that we weren't
doing the same multiply-then-divide operation for `std::complex<int>`
that we would do for `int`.  That was very easy to fix.

The rest of the tests cover the tricky and subtle case that #445
handled.  In the parlance of our new `:abstract_operations` target, if
you try to `StaticCast<double, std::complex<int>>`, it will _compile_,
but you'll get a warning.  `std::complex<int>` has an `int` constructor,
and the `double` will get _implicitly_ converted to `int` --- even
though the static cast itself is an _explicit_ conversion!

To fix this, we need to add two abstractions into our strategy
generation.  First, instead of calling it the "promoted common" type
(which was always properly an implementation detail), we now call it
what it is: the _"application rep"_ --- that is, the rep we should use
for applying the conversion factor.  Next, we replace `StaticCast` with
`StaticCastSequence`, which allows us to detect problematic single-hop
conversions and replace them with a more manageable sequence.

For now, we're only handling the "real to complex" case, but these
abstractions feel like the right ones.  In general, we need to get our
initial type to the conversion type, perform the conversion, and then
get the result to the final type.

Helps #349.

### DIFF
--- a/au/conversion_strategy.hh
+++ b/au/conversion_strategy.hh
@@ -73,12 +73,12 @@ struct ApplicationStrategyForImpl<T, Mag, MagKind::NONTRIVIAL_RATIONAL>
           MultiplyTypeBy<T, Mag>> {};
 
 //
-// `RepForConversion<OldRep, NewRep>` is the rep we should use when applying the conversion factor.
+// `ConversionRep<OldRep, NewRep>` is the rep we should use when applying the conversion factor.
 //
 template <typename OldRep, typename NewRep>
-struct ApplicationRepImpl;
+struct ConversionRepImpl;
 template <typename OldRep, typename NewRep>
-using RepForConversion = typename ApplicationRepImpl<OldRep, NewRep>::type;
+using ConversionRep = typename ConversionRepImpl<OldRep, NewRep>::type;
 
 template <typename OldRep, typename NewRep>
 struct IsRealToComplex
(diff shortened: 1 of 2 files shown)


au/conversion_strategy.hh
```diff
@@ -68,31 +68,67 @@ struct ApplicationStrategyForImpl<T, Mag, MagKind::INTEGER_DIVIDE>
...
+          std::is_integral<RealPart<T>>::value,
           OpSequence<MultiplyTypeBy<T, NumeratorT<Mag>>, DivideTypeByInteger<T, DenominatorT<Mag>>>,
           MultiplyTypeBy<T, Mag>> {};
 
 //
-// `FullConversionImpl<OldRep, PromotedCommon, NewRep, Factor>` should resolve to the most efficient
+// `RepForConversion<OldRep, NewRep>` is the rep we should use when applying the conversion factor.
+//
+template <typename OldRep, typename NewRep>
+struct ApplicationRepImpl;
```
hoffbrinkle: "Application" could also imply "the application using this library."  Would `ConversionRep` be more clear in this case?  Or maybe `IntermediateRep`?

chiphogg: Good point.  I went with `ConversionRep`.  Seems like a pretty clear improvement.

hoffbrinkle (approved):

PR #472, opened 2025-07-20
Switch Au to new operation-based implementation

### LINKED ISSUE #349: Improve efficiency of explicit-rep overflow checkers

chiphogg:
We update `:quantity` so that it simply asks for the right conversion,
and applies it.  Then, now that we know that `:quantity` is using
operations, we can update `:conversion_policy` to use those operations
to check for overflow and truncation risk.

This lets us get rid of the `:apply_rational_magnitude_to_integral`,
`:apply_magnitude`, and `:static_cast_checkers` libraries, because we no
longer use them anywhere.  However, we retain their _test files_, by
sticking a very short implementation (based on the new libraries) at the
top of each test file.  This way, we still get value from all those old
test cases: they give us confidence that we executed the switch
correctly.

Fixes #349.


au/apply_magnitude_test.cc
```diff
@@ -26,6 +27,38 @@ using ::testing::Not;
 namespace au {
 namespace detail {
 namespace {
+
+// `NewOverflowChecker<Op>::would_product_overflow(x)` checks whether the value `x` would exceed the
```
hoffbrinkle, line 38: Should we add a similar comment about these being from an older library?  Obviously git history can be used to check that out, but that would explain the differing idiom for the unit tests (normally they have a corresponding library that they are testing).

au/static_cast_checkers_test.cc
```diff
@@ -24,6 +24,16 @@ using ::testing::IsTrue;
 
 namespace detail {
 
+template <typename U, typename T>
```
hoffbrinkle, line 34: Comment about the previous library implementation?

chiphogg: Done.

au/apply_magnitude_test.cc
chiphogg: Great point!  These files would look really weird otherwise.  Done.

PR #474, opened 2025-07-21
Support policy argument for `Quantity` `in`, `as`

chiphogg:
This policy will indicate, _at compile time_, which conversion risks we
should guard against.  To implement this, we add a user-facing "risk
set" idea, with constants `OVERFLOW_RISK`, `TRUNCATION_RISK`, and
`ALL_RISKS` (which is `OVERFLOW_RISK | TRUNCATION_RISK`).  Users can
create a _policy_ by passing any of these to either `check()` or
`ignore()`.

So, for example, `inches(36).as(feet, ignore(TRUNCATION_RISK))` will now
produce `feet(3)`.

This gives us a more intent based interface for circumventing safety
checks, as compared to `coerce_as` and `coerce_in`.  We plan to provide
replacements for the other instances of the "coerce" word, which can be
found in `:quantity_point` and `:constant`.  Then, if real-world
experience with these new APIs is encouraging, we plan to formally
deprecate all "coerce" word APIs in 0.6.0, and delete them in 0.7.0.

Incidentally, we're also switching from SFINAE to a hard error if
somebody tries to call `.in` or `.as` without a valid unit, but this was
always a poor use of SFINAE anyway.

Helps #387.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -173,7 +173,7 @@ class Quantity {
         std::enable_if_t<std::is_convertible<CorrespondingQuantityT<T>, Quantity>::value, int> = 0>
     constexpr Quantity(T &&x) : Quantity{as_quantity(std::forward<T>(x))} {}
 
-    // `q.as<R>(new_unit)`, or `q.as<R>(new_unit, risk_policy)`
+    // `q.as<Rep>(new_unit)`, or `q.as<Rep>(new_unit, risk_policy)`
     template <typename NewRep,
               typename NewUnitSlot,
               typename RiskPolicyT = decltype(ignore(ALL_RISKS))>
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -173,58 +173,32 @@ class Quantity {
         std::enable_if_t<std::is_convertible<CorrespondingQuantityT<T>, Quantity>::value, int> = 0>
     constexpr Quantity(T &&x) : Quantity{as_quantity(std::forward<T>(x))} {}
 
+    // q.as<R>(new_unit, risk_policy)
     template <typename NewRep,
-              typename NewUnit,
-              typename = std::enable_if_t<IsUnit<AssociatedUnitT<NewUnit>>::value>>
-    constexpr auto as(NewUnit) const {
-        return make_quantity<AssociatedUnitT<NewUnit>>(
-            detail::ConversionForRepsAndFactor<
-                Rep,
-                NewRep,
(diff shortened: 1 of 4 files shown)


au/quantity.hh
```diff
@@ -173,58 +173,32 @@ class Quantity {
...
-                Rep,
-                Rep,
-                UnitRatioT<AssociatedUnitT<Unit>, AssociatedUnitT<NewUnit>>>::apply_to(value_));
+              typename NewUnitSlot,
+              typename RiskPolicyT = decltype(ignore(ALL_RISKS))>
+    constexpr auto as(NewUnitSlot u, RiskPolicyT policy = RiskPolicyT{}) const {
+        return make_quantity<AssociatedUnitT<NewUnitSlot>>(in_impl<NewRep>(u, policy));
     }
 
+    // `q.as<Rep>(new_unit)`, or `q.as<Rep>(new_unit, risk_policy)`
```
```suggestion
    // `q.as(new_unit)`, or `q.as(new_unit, risk_policy)`
```

au/quantity.hh
```diff
@@ -173,58 +173,32 @@ class Quantity {
...
+    template <typename NewUnitSlot, typename RiskPolicyT = decltype(check(ALL_RISKS))>
+    constexpr auto as(NewUnitSlot u, RiskPolicyT policy = RiskPolicyT{}) const {
+        return make_quantity<AssociatedUnitT<NewUnitSlot>>(in_impl<Rep>(u, policy));
+    }
+
+    // `q.in(new_unit)`, or `q.in(new_unit, risk_policy)`
```
```suggestion
    // `q.in<Rep>(new_unit)`, or `q.in<Rep>(new_unit, risk_policy)`
```

hoffbrinkle (approved):

au/quantity.hh
chiphogg: Nice catch(es), thanks!

au/quantity.hh
```diff
@@ -173,58 +173,32 @@ class Quantity {
         std::enable_if_t<std::is_convertible<CorrespondingQuantityT<T>, Quantity>::value, int> = 0>
     constexpr Quantity(T &&x) : Quantity{as_quantity(std::forward<T>(x))} {}
 
+    // `q.as<R>(new_unit)`, or `q.as<R>(new_unit, risk_policy)`
```
chiphogg:
```suggestion
    // `q.as<Rep>(new_unit)`, or `q.as<Rep>(new_unit, risk_policy)`
```

PR #477, opened 2025-07-21
Support risk policy argument in `as_raw_number()`

chiphogg:
I still haven't fully figured out the updates for `Constant`, but at
least `as_raw_number()` is pretty straightforward and easy to knock out!

Helps #387.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -105,9 +105,9 @@ constexpr auto as_quantity(T &&x) -> CorrespondingQuantityT<T> {
 // Only works for dimensionless `Quantities`; will return a compile-time error otherwise.
 //
 // Identity for non-`Quantity` types.
-template <typename U, typename R>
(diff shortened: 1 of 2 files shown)


au/quantity.hh
```diff
@@ -105,9 +105,9 @@ constexpr auto as_quantity(T &&x) -> CorrespondingQuantityT<T> {
 // Only works for dimensionless `Quantities`; will return a compile-time error otherwise.
 //
 // Identity for non-`Quantity` types.
-template <typename U, typename R>
-constexpr R as_raw_number(Quantity<U, R> q) {
-    return q.as(UnitProductT<>{});
```
chiphogg, line 110: Not sure why this was `as` before.  Amusingly, it still worked because we are converting to the rep, and this conversion is allowed implicitly _only for_ something equivalent to `UnitProductT<>{}`.

But `.in` is clearer.

hoffbrinkle (approved):

PR #479, opened 2025-07-21
Split out base power computation into templates

### LINKED ISSUE #467: Fix Green Hills Compiler warnings in Aurora-internal builds
We can't easily create CI builds for this in this open source repo.  But we'll be blocked on releasing [0.5.0](https://github.com/aurora-opensource/au/milestone/8) until we fix this.

The errors look really dumb to my eye: it feels like a total lack of understanding of how generic programming is supposed to work.  But, we should be able to fix them anyway.

chiphogg:
Those templates are primarily on the sign of `N`, and on whether or not
the type is an integer.

The new implementation of `root` handles 3 common and universally
applicable trivial cases.  Then, it delegates to a so-called "nontrivial
root implementation" type to do the rest.  This implementation splits on
whether the type in question is an integer type (using the simplest yet
most expansive detection that I know), or not.  (We used to split on
`std::is_floating_point`, but that gave us the wrong default behavior
for non-arithmetic types.)

For integers, the "nontrivial root implementation" is simple: we just
return an error condition.  (I added a comment explaining why we don't
have to handle integers whose roots are exact integers.)

For other types, we copy the remaining implementation wholesale: handle
the remaining special cases (negative numbers and numbers in (0, 1)),
and then dive in to the full algorithm.

As for the `base_power_value` implementation itself (which is,
incidentally, the only caller of `root()` in the codebase), we use a new
`enum class` to split based on the sign in a readable way.  Amusingly,
this saves us one `static_cast<std::uintmax_t>` in one branch's
implementation, because now we do it upstream.

Fixes #467.  This was tested on the Aurora-internal repo: we got a fully
clean build when this was combined with #475.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -489,13 +489,13 @@ struct GeneralNontrivialRoot {
         if (x < 0) {
             if (n % 2 == 0) {
                 return {MagRepresentationOutcome::ERR_INVALID_ROOT};
-            } else {
-                const auto negative_result = root(-x, n);
-                if (negative_result.outcome != MagRepresentationOutcome::OK) {
-                    return {negative_result.outcome};
-                }
-                return {MagRepresentationOutcome::OK, static_cast<T>(-negative_result.value)};
+
+            const auto negative_result = root(-x, n);
+            if (negative_result.outcome != MagRepresentationOutcome::OK) {
+                return {negative_result.outcome};
             }
+
+            return {MagRepresentationOutcome::OK, static_cast<T>(-negative_result.value)};
         }
 
         // Handle numbers bewtween 0 and 1.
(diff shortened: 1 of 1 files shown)


au/magnitude.hh
```diff
@@ -448,89 +474,108 @@ constexpr MagRepresentationOrError<T> root(T x, std::uintmax_t n) {
...
+        if (x < 0) {
+            if (n % 2 == 0) {
+                return {MagRepresentationOutcome::ERR_INVALID_ROOT};
+            } else {
+                const auto negative_result = root(-x, n);
+                if (negative_result.outcome != MagRepresentationOutcome::OK) {
+                    return {negative_result.outcome};
+                }
+                return {MagRepresentationOutcome::OK, static_cast<T>(-negative_result.value)};
+            }
```
hoffbrinkle: I tend to drop the `else` when it's made redundant by other control mechanisms.  I think it makes this more clear that this `if (x < 0) {` is going to `return` in any case:

```suggestion

            const auto negative_result = root(-x, n);
            if (negative_result.outcome != MagRepresentationOutcome::OK) {
                return {negative_result.outcome};
            }

            return {MagRepresentationOutcome::OK, static_cast<T>(-negative_result.value)};
```

PR #488, opened 2025-07-30
Fix up the fuzz helper tool

chiphogg:
The most pressing issue is that right now, it makes `//...:all`
nightmarishly slow.  It basically adds about a minute or two _after_ all
the tests finish running, just to build (and not even run) this target.
(This utility is extremely compiler intensive!)  The easy fix is to add
a `manual` tag.

While I'm at it, I might as well stop using gtest as a shortcut to get
an executable.  This was never really that much of a shortcut to begin
with --- it's just a lazy decision I made N months ago when I started
working on this, and I never felt like writing a `main` function.  I was
going to add the comments as [suggested] in #461 to explain why we're
using gtest, but it was easier to just stop using gtest.

[suggested]: https://github.com/aurora-opensource/au/pull/461#discussion_r2217955375

### DIFF
--- a/fuzz/BUILD.bazel
+++ b/fuzz/BUILD.bazel
@@ -18,11 +18,11 @@ cc_binary(
     name = "quantity_runtime_conversion_check",
     testonly = True,
     srcs = ["quantity_runtime_conversion_check.cc"],
+    tags = ["manual"],
     deps = [
         ":quantity_runtime_conversion_checkers",
         "//au:conversion_strategy",
         "//au:testing",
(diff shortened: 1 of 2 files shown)


geoffviola:
Thanks for optimizing the workflow.

Do you see the test as more of an integration test? More specifically, when should fuzz/quantity_runtime_conversion_check.cc be run?

I can see that there are assertions, but it doesn't use the GTest framework. Why is that?

hoffbrinkle (approved):

Issue #149, opened 2023-07-12
Add support for fmtlib


chiphogg:
We have a formatter internal to Aurora, which was pretty easy to write.  However, I don't think it was that smart about taking the unit label length into account.  Maybe we can write some test cases for an improved implementation.

chiphogg:
Yes, this issue tracker is only for things that we intend for the public library.  🙂  We have another internal tracker for Aurora-internal issues.

It's a very busy time at Aurora right now with our product launch imminent, so we can't work on Au much at all right now.  That said, I think we could fit it in for the next release we _do_ get a chance to do, so I'll bump it up to the 0.4.2 milestone.  Thanks for asking!

chiphogg:
Update: a flurry of activity, with a frustrating result.  I've already spent at least 3x as much time as I expected when I put this on the 0.4.2 (now [0.5.0](https://github.com/aurora-opensource/au/milestone/8)) milestone.  We even got to the point of landing what we thought was a working implementation!  But now I don't even know if this is feasible.  At the very least, I'll probably revert what we have for the 0.5.0 release, and hope that we can figure something out in time for 0.6.0.

The tricky part is that we need to maintain Au's zero-dependency status, and we need a solution that works for a wide variety of `{fmt}` versions, as well as `std::format`.  Those are some wickedly hard constraints, as far as I can (now) tell.

It's not all bad news.  I feel pretty satisfied with the format string syntax we came up with (see #491).  And I have a much better sense of the shape of the problem and its constraints.

I think I'll write up a detailed retrospective on what happened.  The main audience is "future me, trying to solve this for 0.6.0", but it may be of interest to others yearning for this feature.  Maybe I'll follow up by taking stock of where we're at.

## Strategies and attempts

### Initial idea: forward declare and specialize

If this had worked, it would have been the best end user experience: anyone who uses `fmt::format` or `std::format` would get Au's format string syntax for free, without any extra work.  The idea was to forward declare the `fmt::formatter` class, and provide a specialization for `au::Quantity<U, R>`.

Note how this avoids the _physical_ dependency on `{fmt}`.  Without `{fmt}`, it'd just be some random template specialization.  But as soon as a user includes `"fmt/format.h"`, the specialization "comes to life", and they automatically get Au's format string syntax.

[...]

PR #492, opened 2025-07-30
Make `parse_ctx` constructor call more precise

chiphogg:
Here is the first attempt to test `std::format` support:
<https://godbolt.org/z/fYbarPdvG>.

The error complains about our `static_cast<int>` needing to _then_
become a `size_t` in the `std::format` implementation.  So I decided to
look into the constructors in more detail.  It turns out that for
`{fmt}`, the constructor takes a `fmt::string_view` and optional `int`
argument count, while the `std::format` takes a `std::string_view` and a
`size_t` argument count.

So this makes it seem like the constructor was spelled wrongly, and we
were actually providing an argument count, not a second argument to the
`xyz::string_view`.  Oops!

The fix is to construct it with a _nested_ braced list whose second
argument is the `size_t` difference.  To disambiguate the constructor,
we also need to pass something for the argument count.  `0` should work
in all cases.  It's pretty rare for the format parse context argument to
actually _need_ the argument count, and is mostly used for, e.g.,
getting part of a formatting string from a _different numbered
argument_.

Helps #149.  After this, we will try again with the godbolt link ---
unfortunately, we can't easily test the `std::format` directly just yet.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -941,7 +941,7 @@ struct QuantityFormatter {
         if (is_unit_label) {
             ++it;
         }
-        FormatParseContext parse_ctx({it, static_cast<std::size_t>(next_end - it)});
+        FormatParseContext parse_ctx{{it, static_cast<std::size_t>(next_end - it)}, 0};
 
         if (is_unit_label) {
             unit_label_format.parse(parse_ctx);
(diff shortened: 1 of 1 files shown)


chiphogg:
Taking into draft: I found some issues with the context constructor on slightly older versions of fmt.

hoffbrinkle (approved):

chiphogg:
Closing (see [comment](https://github.com/aurora-opensource/au/issues/149#issuecomment-3144905376) for more details).

PR #493, opened 2025-07-31
Add `-Wsign-conversion` coverage to `:fmt_test`

chiphogg:
We fix the problem by downgrading to 11.0.2, which was the last version
before the `-Wsign-conversion` errors were introduced (fmtlib/fmt#4268).

Helps #149.


chiphogg:
I think this depends on #492.  I'll cherry pick that fix just to test.

PR #497, opened 2025-08-04
Update requirements lock

chiphogg:
Fixes dependabot issues 18, 19, 21.

The only named package update is for mkdocs-material, from 9.6.14 to
9.6.16.  There are no significant-seeming changes to the layout in
[9.6.15] or [9.6.16].  There is, however, a bugfix for
squidfunk/mkdocs-material#8334.  I was able to reproduce this bug on our
current website:

1. Click search bar.
2. Type `qu`.
3. Click outside the search bar to de-focus it.
4. `Ctrl-F` for `qu`:

You get an inordinately high number of results, most of which are from
the search bar (and thus hidden).  On the new version, you get only 8
results, all visible.

[9.6.15]: https://github.com/squidfunk/mkdocs-material/releases/tag/9.6.15
[9.6.16]: https://github.com/squidfunk/mkdocs-material/releases/tag/9.6.16


geoffviola (approved):
Thanks for updating these dependencies and noting the bug fix.

PR #499, opened 2025-08-07
Support risk policy parameter in constructors

chiphogg:
Sometimes, end users have aliases for `Quantity` or `QuantityPoint`
types.  In these cases, sticking exclusively to `.as` can be more
awkward than just using a constructor.  Therefore, we add new Q-to-Q
(and QP-to-QP) constructors to support a second policy argument.

Follow-on for #387.


geoffviola (approved):
The changes look good. Thanks for cleaning that up.

> sticking exclusively to .as can be more
awkward than just using a constructor.

What's the guidance there? Use fewer characters? 😁

Issue #470, opened 2025-07-20
Take a pass through docs before cutting release


chiphogg:
I'm guessing we'll have missed a bunch of stuff, and we definitely want to reflect it in the released version of the doc website.  Including, but not limited to:

- [ ] Library comparison
- [ ] Troubleshooting guide
- [ ] `coerce_as`: dis-recommend

PR #500, opened 2025-08-08
Update docs to reflect conversion risk policies

chiphogg:
We add three new documents: two discussion docs on truncation risk and
conversion risks generally, and one reference doc for the conversion
risk policy objects.

Then we go through and update all of the affected docs.  The main types
of changes are:

- Update new APIs (constructors, conversion functions) that can take a
  policy argument.
- Mark the "coerce" APIs as "to be deprecated", and tell people to use
  the policy arguments instead.
- If we find outdated language around permitted/forbidden conversions,
  or around "explicit Rep" APIs, clean it up

This should get the Au docs into a good state with respect to our modern
approach to conversion safety.

Helps #470.

### DIFF
--- a/docs/discussion/concepts/conversion_risks.md
+++ b/docs/discussion/concepts/conversion_risks.md
@@ -66,8 +66,12 @@ from determining whether _a specific input value_ will be lossy.  For example:
 | Conversion | Policy | Result (if allowed) | Actually lossy? |
 |------------|--------|---------------------|-----------------|
 | `inches(24).as(feet)` | :x: Forbidden (truncation risk) | `feet(2)` | :thumbs_up: No |
-| `seconds(4u).as(nano(seconds))` | :x: Forbidden (overflow risk) | `nano(seconds(4'000'000'000u))` | :thumbs_up: No |
-| `seconds(9000u).as(micro(seconds))` | :thumbs_up: Allowed | `micro(seconds(410'065'408u))` | :x: Yes (overflow) |
+| `seconds(4u).as(nano(seconds))`[^1] | :x: Forbidden (overflow risk) | `nano(seconds(4'000'000'000u))` | :thumbs_up: No |
+| `seconds(9000u).as(micro(seconds))`[^1] | :thumbs_up: Allowed | `micro(seconds(410'065'408u))` | :x: Yes (overflow) |
+
+[^1]: In these docs, we assume that `unsigned int` values (such as `4u` and `9000u`) are 32 bits
+wide.  On architectures where this is not true, you could pattern-replace items such as `4u` with
+`uint32_t{4}` to reproduce the results in the tables.
 
 This shows that our default policy has two kinds of failure modes.
 
@@ -96,7 +100,7 @@ Let's look at some more examples to make this clear.
 | Conversion | Policy | Result | Actually lossy? |
 |------------|--------|--------|-----------------|
 | `inches(24).as(feet, ignore(TRUNCATION_RISK))` | :thumbs_up: Allowed (by policy override) | `feet(2)` | :thumbs_up: No |
-| `seconds(4u).as(nano(seconds), ignore(OVERFLOW_RISK))` | :thumbs_up: Allowed (by policy override) | `nano(seconds(4'000'000'000))` | :thumbs_up: No |
+| `seconds(4u).as(nano(seconds), ignore(OVERFLOW_RISK))`[^1] | :thumbs_up: Allowed (by policy override) | `nano(seconds(4'000'000'000))` | :thumbs_up: No |
 
 Here, we have revisited the examples from the previous table, but thanks to our policy override,
 they are no longer forbidden.  We see that they do produce the correct results.  However, we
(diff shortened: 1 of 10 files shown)


docs/discussion/concepts/conversion_risks.md
```diff
@@ -0,0 +1,204 @@
...
+
+It's critically important to understand that these policies are based on _risk_, which takes into
+account _all possible_ input values (and, crudely, their relative likelihoods).  This is different
+from determining whether _a specific input value_ will be lossy.  For example:
+
+| Conversion | Policy | Result (if allowed) | Actually lossy? |
+|------------|--------|---------------------|-----------------|
+| `inches(24).as(feet)` | :x: Forbidden (truncation risk) | `feet(2)` | :thumbs_up: No |
+| `seconds(4u).as(nano(seconds))` | :x: Forbidden (overflow risk) | `nano(seconds(4'000'000'000u))` | :thumbs_up: No |
+| `seconds(9000u).as(micro(seconds))` | :thumbs_up: Allowed | `micro(seconds(410'065'408u))` | :x: Yes (overflow) |
```
hoffbrinkle: I think this is a clarity vs correctness question, but `seconds(uint32_t{4}).as(nano(seconds))`.  Maybe we just put a note to the effect of "for a 32 bit representation"?  I'm on the fence if we really need to, as I think it's probably assumed/implied.

chiphogg: Good catch.  I agree it's assumed/implied, and the clarity is more important than full technical correctness.  But we can still do better!  I ended up adding a footnote, and linking to it in every place where we used literals like `4u` for their brevity.

docs/discussion/concepts/truncation.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Integral types
+
+Integral types are considered _exact_.  To see the implications for truncation, we need to consider
+various [types of conversion factors](../implementation/applying_magnitudes.md).
+
+First, we have multiplying by an integer.  This is easy: it stays within the domain of the integers,
+so it can never truncate.
+
+Next, we have multiplication by a non-integer rational (including reciprocal integers).  This will
```
hoffbrinkle, line 48: "First ... multiplying" and "Next ... multiplication" aren't consistent.  Not sure if they need to be, just pointing that out.

docs/discussion/concepts/index.md
```diff
@@ -26,6 +31,10 @@ and help you use units libraries more effectively.
 - **[Quantity Point](./quantity_point.md)**.  An abstraction for "point types" that have units.
   Most use cases don't need this, but for a few --- including temperatures --- it's indispensable.
 
+- **[Truncation](./truncation.md)**.  Unit conversions truncate when the destination type can't hold
+  the result with acceptable precision.  This page will explain Au's approach to managing this risk,
```
hoffbrinkle, line 35:
```suggestion
  the result with acceptable precision.  This page explains Au's approach to managing this risk,
```

docs/reference/conversion_risk_policies.md
```diff
@@ -0,0 +1,36 @@
+# Conversion Risk Policies
+
+A conversion risk policy is a [monovalue type](./detail/monovalue_types.md) that indicates how
+a unit conversion function should handle [conversion
+risks](../discussion/concepts/conversion_risks.md).  Au tracks two kinds of conversion risk:
+overflow, and truncation.  Each conversion risk policy will either enable or disable checking for
```
hoffbrinkle, line 6:
```suggestion
overflow and truncation.  Each conversion risk policy will either enable or disable checking for
```

docs/reference/quantity_point.md
```diff
@@ -254,29 +265,83 @@ are forbidden.  Additionally, the `Rep` of the output is identical to the `Rep`
...
+off the corresponding safety check, by passing a [policy argument](./conversion_risk_policies.md).
+Every variant of `as` and `in` mentioned above supports such an argument.  If you pass a policy, it
+will control the set of conversion risks that we check for.
+
+To be concrete, here are the signatures of the functions that support the policy argument:
+
+| Usual form | With policy argument |
+|------------|----------------------|
+| `.as(unit)` | `.as(unit, policy)` |
+| `.as<T>(unit)` | `.as<T>(unit, policy)` |
```
hoffbrinkle, line 293: Feels like there's a bunch of duplication with the docs for `Quantity`, though off hand I don't have a great idea on how to consolidate.

hoffbrinkle (commented):
Sorry, I didn't notice the rest of the docs in this PR after I reviewed the first file.  Nothing major to note in the remainder.

docs/discussion/concepts/truncation.md
chiphogg: They certainly should be consistent.  Thanks!

docs/discussion/concepts/index.md
chiphogg: This is pretty funny.  When I wrote this entry, I hadn't yet written the truncation page.  So I was thinking of this in my head as future work, and it accidentally slipped into my phrasing!

docs/reference/conversion_risk_policies.md
chiphogg: Done (#503).

docs/reference/quantity_point.md
chiphogg: Yes, that was mostly an intentional decision to keep each page more self-contained.  It does create maintenance risk, though.  I don't love the situation either, but I also don't have a better idea.

PR #501, opened 2025-08-09
Update docs for common point unit

chiphogg:
The common point unit is simply the common _unit_ of both the input
units, and of the implicitly created "origin displacement" units as
appropriate.  Its origin is the lowest origin of all input units.

This PR fleshes out the description in the discussion page, and provides
a full definition in the reference page.  We also include a worked
example of how the GCD algorithm lets you compute what the common point
unit is.

Finally, I added some links for some usages I saw of "quantity
equivalent" and "point equivalent", because obviously we can't expect
users to just know what these mean.

Helps #470.

### DIFF
--- a/docs/discussion/concepts/common_unit.md
+++ b/docs/discussion/concepts/common_unit.md
@@ -148,16 +148,17 @@ unspecified.
 ## Changes for `QuantityPoint` {#common-quantity-point}
 
 The common unit for [`QuantityPoint`](./quantity_point.md) operations is different from the common
-unit for `Quantity`.  To see why a single notion of "common unit" isn't enough, consider `Celsius`
+unit for `Quantity`.  To see why we can't just reuse the same "common unit", consider `Celsius`
 and `Kelvins`.
 
 - For a **`Quantity`**, these two units are identical.  The "common unit" will be
-  (quantity-)equivalent to both of them.
+  [(quantity-)equivalent](../../reference/unit.md#quantity-equivalent) to both of
+  them.
 
 - For a **`QuantityPoint`**, these units are very different.  A "temperature point" of 0 degrees
-  `Celsius` is (point-)equivalent to a temperature point of 273.15 `Kelvins`.  This additive offset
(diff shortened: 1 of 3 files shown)


docs/discussion/concepts/common_unit.md
```diff
@@ -175,6 +176,34 @@ unit," but the origin adds a new complication.  We'll need to choose a conventio
...
+    | 273.15 K | 1 K   | 0.15 K    |
+    | 1 K    | 0.15 K | 0.10 K    |
+    | 0.15 K | 0.10 K | 0.05 K    |
+    | 0.10 K | 0.05 K | 0 K       |
+
+    Therefore, the greatest common divisor has a magnitude of 0.05 K.  This means that our common
+    point unit is a unit with a magnitude of 0.05 K, whose origin equals the origin of Kelvins
+    (because it is lower than the origin of degrees Celsius).
+
+    Au will automatically generate the label
```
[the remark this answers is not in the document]
chiphogg: "Au" is correct.  It is the way that people are supposed to refer to the library in prose.  I believe we use this consistently throughout the docs.

docs/reference/unit.md
```diff
@@ -653,6 +653,16 @@ A specialization will only exist if the inputs are all units, and will exist but
 error if any two input units have different Dimensions.  We also strive to keep the result
 associative, and symmetric under interchange of any inputs.
 
+To understand the precise definition of the common point unit, first assume that the "origin
+displacement" for any two units with different origins is itself a new unit, which we can create
+implicitly.  Then, the _common point unit_ is defined as the unit such that:
+
+- It is quantity-equivalent to the _common unit_ of **both** all input units, **and** of all "origin
```
hoffbrinkle: I'm not the Grammarian by any stretch, but is this an Oxford comma++?

```suggestion
- It is quantity-equivalent to the _common unit_ of **both** all input units **and** of all "origin
```

chiphogg: I liked the pause when I read it in my head, but I think your suggestion technically improves the grammar.  :slightly_smiling_face:

PR #505, opened 2025-08-13
Support `{fmt}` and `std::format`

### LINKED ISSUE #149: Add support for fmtlib

chiphogg:
We tried this earlier (#491), but then reverted it (#495).  That was
just a few short weeks ago: what changed?

We learned that our old format string syntax wasn't just _hard_ to
implement; it was actually [impossible].  This led naturally to a [new
idea] that may not be as flexible, but actually still supports the most
important use cases.  Happily, this new idea is very easy to implement
without using any internal implementation details of the formatting
libraries.  It even gives us a little (_very_ little) room for evolving
more options for formatting the unit label, such as setting the fill
character, although we don't currently expect to ever need or want them.

I was also able to verify support for `std::format` by generating a
single file, and uploading it to [godbolt].  This shows that on a
compiler version and setup that includes `std::format`, Au supports it
via the `"au/std_format.hh"` header.  I also updated the single-file
script to have `--no-std-format` (which is the default) and
`--std-format` options.  A future PR will add versions with/without
`std::format` support to our automatically generated single-file
options.

On the CMake side, we need to skip linting for `std_format.hh`.  It's
too hard and too complicated to set up reliable rules for when we should
and shouldn't make it available.  Therefore, we _always_ make it
available, but we trust that users will only include it if they know
they can support it.  That's the point of the opt-in header strategy.
Note, though, that this required us to bump the minimum CMake version
from 3.24 to 3.27.

Fixes #149.

[impossible]: fmtlib/fmt#4508 (comment)
[new idea]: fmtlib/fmt#4508 (comment)
[godbolt]: https://godbolt.org/z/f6Pvz9Mzb

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -957,16 +957,6 @@ struct QuantityFormatter {
 
 }  // namespace au
 
-// We want the `std::formatter` specialization to be defined if and only if `std::format` is
-// supported.  Putting it in the same file helps avoid ODR violations.
-#if defined(__cpp_lib_format) && __cpp_lib_format >= 201907L
-#include <format>
-namespace std {
-template <typename U, typename R>
-struct formatter<au::Quantity<U, R>> : ::au::QuantityFormatter<U, R, ::std::formatter> {};
-}  // namespace std
-#endif
-
 namespace std {
 // Note: we would prefer not to reopen `namespace std` [1].  However, some older compilers (which we
 // still want to support) incorrectly treat the preferred syntax recommended in [1] as an error.
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -879,7 +879,91 @@ struct CommonQuantity<Quantity<U1, R1>,
                       Quantity<U2, R2>,
                       std::enable_if_t<HasSameDimension<U1, U2>::value>>
     : stdx::type_identity<Quantity<CommonUnitT<U1, U2>, std::common_type_t<R1, R2>>> {};
+
+//
+// Formatter implementation for fmtlib or `std::format`.
+//
+// To use with fmtlib, add this template specialization to a file that includes both
+// `"au/quantity.hh"`, and `"fmt/format.h"`:
+//
+//    namespace fmt {
+//    template <typename U, typename R>
+//    struct formatter<::au::Quantity<U, R>> : ::au::QuantityFormatter<U, R, ::fmt::formatter> {};
+//    }  // namespace fmt
+//
+// Then, include that file any time you want to format a `Quantity`.
+//
+template <typename U, typename R, template <class...> class Formatter>
+struct QuantityFormatter {
+    template <typename FormatParseContext>
+    constexpr auto parse_unit_label_part(FormatParseContext &ctx) {
+        auto it = ctx.begin();
+
+        if (it != ctx.end() && *it == 'U') {
+            // Consume the 'U'.
+            ++it;
+
+            // Parse the total width.
+            while (it != ctx.end() && *it >= '0' && *it <= '9') {
+                min_label_width_ = (min_label_width_ * 10) + static_cast<std::size_t>(*it++ - '0');
+            }
+
+            if (it != ctx.end()) {
+                if (*it == '}') {
+                    return it;
+                }
+
+                if (*it++ != ';') {
+                    // Cause an error condition in further parsing.
+                    it = ctx.end();
+                }
+            }
+        }
+
+        return it;
+    }
+
+    template <typename FormatParseContext>
+    constexpr auto parse(FormatParseContext &ctx) {
+        ctx.advance_to(parse_unit_label_part(ctx));
+        return value_format.parse(ctx);
+    }
+
+    template <typename FormatContext>
+    constexpr auto format(const au::Quantity<U, R> &q, FormatContext &ctx) const {
+        value_format.format(q.data_in(U{}), ctx);
+        *ctx.out()++ = ' ';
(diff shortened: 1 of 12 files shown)


au/quantity.hh
```diff
@@ -879,8 +879,92 @@ struct CommonQuantity<Quantity<U1, R1>,
...
+//
+// Then, include that file any time you want to format a `Quantity`.
+//
+template <typename U, typename R, template <class...> class Formatter>
+struct QuantityFormatter {
+    template <typename FormatParseContext>
+    constexpr auto parse_unit_label_part(FormatParseContext &ctx) {
+        auto it = ctx.begin();
+
+        if (it != ctx.end() && *it == 'U') {
```
hoffbrinkle: Optional:

I'm more and more a fan of returning like I vote: "early and often." 😁.  The advantage being that the business logic can be outdented:

```suggestion
        if (it == ctx.end()) {
            return it;
        }
        
        if (*it != 'U') {
            return it;
        }
```

au/quantity.hh
```diff
@@ -879,8 +879,92 @@ struct CommonQuantity<Quantity<U1, R1>,
...
+            if (it != ctx.end()) {
+                if (*it == '}') {
+                    return it;
+                }
+
+                if (*it++ != ';') {
+                    // Cause an error condition in further parsing.
+                    it = ctx.end();
+                }
+            }
```
```suggestion
            if (it == ctx.end()) {
                return it;
            }

            if (*it == '}') {
                return it;
            }

            if (*it++ != ';') {
                // Cause an error condition in further parsing.
                it = ctx.end();
            }
```

docs/reference/format.md
```diff
@@ -0,0 +1,138 @@
...
+- If both are present, then the unit label format string must come first, and they must be separated
+  by a `;`.
+
+It may be easier to understand with several examples.
+
+| Specialize format for | Pattern | Example | Output |
+|-----------------------|---------|---------|--------|
+| Neither | `{}` | `fmt::format("{}", meters(123.456))` | `"123.456 m"` |
+| Numeric value | `{:~^10.2f}` | `fmt::format("{:~^10.2f}", meters(123.456))` | `"~~123.46~~ m"` |
+| Unit label | `{:U5}` | `fmt::format("{:U5}", meters(123.456))` | `"123.456 m____"`<br>(Each `'_'` char represents a space `' '`; see footnote[^1]) |
```
hoffbrinkle: Mind if I push up a commit to fix this?

docs/reference/index.md
```diff
@@ -31,6 +31,11 @@ Here's a guide to the main reference pages.
   unit with a specified _relative size_.  For example, `centi` is a prefix whose relative magnitude
   is $1/100$, and a `centi(meter)` is a new unit which is one one-hundredth of a `meter`.
 
-- **[`Math functions`](./math.md).**  We provide many common mathematical functions out of the box.
+- **[Math functions](./math.md).**  We provide many common mathematical functions out of the box.
```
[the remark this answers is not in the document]
chiphogg: This is nonsense.  No line has been duplicated; line 35 is a blank line.  The change to this line was to remove some backticks that never should have been there in the first place.

au/quantity.hh
```diff
@@ -879,8 +879,92 @@ struct CommonQuantity<Quantity<U1, R1>,
...
+        *ctx.out()++ = ' ';
+        return write_and_pad(unit_label(U{}), sizeof(unit_label(U{})), ctx);
+    }
+
+    template <typename FormatContext>
+    constexpr auto write_and_pad(const char *data,
+                                 std::size_t data_size,
+                                 FormatContext &ctx) const {
+        Formatter<const char *> unit_label_formatter{};
+        unit_label_formatter.format(data, ctx);
```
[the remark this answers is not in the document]
chiphogg: I preferred to go with a high level command, for clearer intent, rather than a loop.  Additionally, I thought the formatter might have better performance --- certainly, at the very least, it should be no worse.

au/quantity.hh
```diff
@@ -879,8 +879,92 @@ struct CommonQuantity<Quantity<U1, R1>,
...
+        while (data_size <= min_label_width_) {
```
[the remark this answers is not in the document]
chiphogg: This statement makes no sense.  It's just token prediction gone off the rails.

The reason for `<=` instead of `<`, incidentally, is that the `'\0'` null terminator was included in the `data_size` computation, and this is just a way to cancel out that mistake.

chiphogg (commented):
The "zero setup" version of `std::format` support turns out to be impossible.  There will always be some build systems that scan includes such as `#include <format>` in order to build an include graph, regardless of any surrounding `#ifdef`-type preprocessor macros.  If we make this design choice, we will break current or future users.

The way forward is to add a new header, `"au/std_format.hh"`, that users can opt in and include.  Only users who can support `#include <format>` would ever do this, so this would be a robust fix.

We also need to upgrade the single file tool to give an option for `std::format` support, off by default.

docs/reference/format.md
chiphogg: If you know how to, then that would be great! A commit or a PR; either would be fine.

hoffbrinkle: Pushed.  You can use the NBSP character.  Unicode code point `U+00A0`.  With MacOS, you can use `option` + `space`.  In Vim, if you're in insert mode: `CTRL` + `V`, `U`, `00a0`.

chiphogg: This looks much better --- thanks!

PR #513, opened 2025-08-17
Update troubleshooting guide for 0.5.0 release

chiphogg:
First, this is the right time to make a few _code_ changes.  The error
messages for unit conversions and inversions now link directly to the
corresponding section of the troubleshooting guide.  An HTML page can
discuss the necessary subtleties in much better depth than any
reasonable `static_assert` message.  Especially when what we want to say
is some variant of "here's how to do this, _but you probably don't want
to..._".

Now, for the troubleshooting guide itself.  First, I removed the clang
11 section entirely.  The marginal value over and above having clang 14
alone is very small, but the maintenance burden is significant.

Then, I tweaked some newly updated sections:

- "Dangerous conversion" became "Conversion risk too high".  I added a
  bunch of links to our brand new documentation pages, and reworked the
  text to explain things in terms of our modern thinking.
- "Integer division forbidden" now explains more carefully _precisely
  which cases_ are forbidden, and warns much more strongly that
  `unblock_int_div` is risky.
    - It made sense to update the reference docs as well, so I did.
      They now also warn more strongly, and they encourage users to read
      our detailed warning in the troubleshooting page.
- "Dangerous inversion" was just plain outdated: we didn't update the
  text when we tightened up our policy.  This is fixed.
- "Broken strict total ordering" now directs users to specialize
  `au::UnitOrderTiebreaker`, instead of `au::detail::UnitAvoidance`.

Finally, I made sure the section titles in `error_examples.cc` were all
present and correct, and I moved some CSS upstream so that we could use
"checkmark-bulleted list items" more broadly.

Helps #470.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,7 +9,7 @@ search the text of this page using your browser's Find function.
 
 !!! tip
     To improve your chances of finding what you're looking for, we include full compiler errors from
-    both gcc and clang, inline with the text.  Naturally, this makes this page very long, so it's
+    gcc, clang, and MSVC, inline with the text.  Naturally, this makes this page very long, so it's
     not meant to be read straight through.  Stick with your browser's Find function.
 
 Each section below lists one category of compiler error you might encounter when using the library.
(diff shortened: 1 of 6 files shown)


docs/troubleshooting.md
```diff
@@ -79,69 +79,53 @@ It's the "classic" error the units library aims to prevent.
 
     **Compiler error (clang 14)**
     ```
-    au/error_examples.cc:33:17: error: calling a private constructor of class 'au::Quantity<au::Seconds, double>'
+    au/error_examples.cc:34:17: error: calling a private constructor of class 'au::Quantity<au::Seconds, double>'
```
hoffbrinkle, line 82: For these compiler error output updates, do you have a script that updates these?  I could see that being finicky to set up, but the tradeoff would be not having to manually maintain all these minor changes.

chiphogg: Not so much a script. More like vim macros to paste the compiler error output. :)

It would be good to write down the instructions, though.

hoffbrinkle: I wonder if we could build these as artifacts and then incorporate them into the generated documentation?

chiphogg: That would be ideal! But I'm not sure how to go about it.

Also, MSVC is pretty labor intensive. I have a long lived branch for it, and every time I need to update it, I push one commit for each compiler error, and get the errors from the remote build logs. 

Still... Maybe someday. 🤞

PR #514, opened 2025-08-18
Update Au's release instructions

chiphogg:
First, we add a section that contains everything we expect to be
relevant for upgrading from the last "minor or larger" release, which we
will name.

Next, we add a "release branch" strategy.  Why?  Well, the direct
motivation is that we have a lot of URLs in our C++ code that point to
`/main/`, whereas this content can change in ways that we can't predict.
We will want it to point to `/0.5.0/` instead.  But I don't want our
history to contain a bunch of PRs that go updating tons of URLs.  It
makes sense for `main` to always point to `/main/`.  So I came up with a
release branch strategy.

The other thing this ties into is the "future proofing" releases.  I am
currently planning to have each future-proofing commit be a PR against
the release **tag** on the release branch.  When the PR lands, its final
commit will get tagged with `0.5.0-future-122`, `0.5.0-future-185`, etc.
When all PRs have landed, the HEAD _of the release branch_ gets tagged
with `0.5.0-future`.  This will let us download tarballs in the exact
same way as for the main release, so everything will be consistent.

We still need a "how-to guide" for upgrading in general, and using
future-proof releases in particular.  But the point of this PR is to
give the team a vehicle to discuss updating our release policies.

Helps #470.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -126,7 +126,7 @@ and push to GitHub.
 
 ```sh
 # Remember to update the version number!
-git switch --checkout release-0.3.1
+git switch --create release-0.3.1
 git push origin release-0.3.1
 ```
 


RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
+
+### Download all artifacts
 
-Manually uploading releases helps future-proof us against previous breakages.  See:
+Manually uploading releases helps future-proof us against known failure modes.  See:
 https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/
 
 On the [tags page](https://github.com/aurora-opensource/au/tags), click the `.tar.gz` link to
-download the tarball.
+download every tarball.  This will always include the release tarball, and may also include one or
```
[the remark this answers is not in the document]
chiphogg: Boo!

RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
+
+Make a PR against the release branch that is, essentially, "our best guess for how we will implement
+this feature".  Strive for something that we could simply cherry pick when it's time to land to the
+main branch.  This doesn't mean we _commit_ to doing it _exactly_ this way, but it's a good guide.
+In any case, the code must be high enough quality to pass code review.
+
+**Important:** every PR must use the release **tag** as its base commit, not the HEAD of the release
+branch.  Do **not** pull in any changes from the release branch later.
+
+When each PR lands, tag the HEAD commit _of the PR's branch_ with a tag of the format
```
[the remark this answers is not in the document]
chiphogg: Nope: the original instructions are precise and correct.

RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
-First, make sure the "final commit" (which updates the CMake variables) has already landed.
+First, make sure the "final commit" (which updates the CMake variables) has already landed, and is
+currently checked out.  This will be the "base" commit for the release.
 
 Then, use the command below, replacing `0.3.1` with the version to create.
 
 ```sh
-# Remember to update the tag number!
+# Remember to update the version number!
+git tag -a 0.3.1-base
```
hoffbrinkle: Do we need this tag in the main repo?  Is it used for anything else other than creating a branch at this commit?  This information is otherwise easily obtainable from git and it won't get stale.

RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
+```
+
+For the message, write "Base commit for the 0.3.1 release." (again, updating the version number as
+needed).  Now, we're going to create the release branch, and push both tag and branch to GitHub.
+
+```sh
+# Remember to update the version number!
+git push origin 0.3.1-base
+git checkout -b release-0.3.1
```
hoffbrinkle: `git` has updated the UI for this:

```suggestion
git switch --checkout release-0.3.1
```

RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
+
+We do this in a PR on the release branch in order to avoid churn commits on the main branch.
+
+### Create the tag for the release
+
+Once the above PR has landed, that commit _is_ the release, so it's time to tag it as such:
+
+```sh
+# Remember to update the version number!
 git tag -a 0.3.1
```
hoffbrinkle: For documentation and the like, I tend to use the long opts.  Users can always use `-a` if they remember the short opt.

```suggestion
# Remember to update the version number!
git tag --annotate 0.3.1
```

RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
+
+At this point, the release branch consists of the release tag commit itself, followed by a series of
+commits that implement future-proofing changes.  We now need to tag the HEAD of the release branch
+as the final future-proofing release.
+
+```sh
+# Remember to update the version number!
+git fetch origin release-0.3.1:release-0.3.1
+git switch release-0.3.1
+git tag -a 0.3.1-future
```
hoffbrinkle: If we go with the above suggestion:

```suggestion
git tag --annotate 0.3.1-future
```

RELEASE.md
```diff
@@ -111,40 +118,140 @@ Issues!  Alphabetically:
...
+# Remember to update the version number!
+Future-proofing release for 0.3.1, comprising:
+
+# Add one line for every future-proofed issue:
+- #NNN
+- #MMM
+# etc.
+```
+
+(Lines starting with `#` are instructions, and should not be included in the actual message.)
```
hoffbrinkle, line 219: Wouldn't they be ignored anyway?

RELEASE.md
chiphogg: Hmm... good point.  We probably don't need it.  Nice catch!

RELEASE.md
chiphogg: I looked it up because `--checkout` sounded redundant.  Looks like it's `--create`.

RELEASE.md
chiphogg: Probably.  I can't remember which situations do and don't ignore the `#`.  For example, I know we use `###` section headers in our PR summaries in AV, and I see those lines still show up in the commit messages that land.

So I just wanted to add this instruction to be super clear.

chiphogg (commented):
Apparently I never actually submitted my responses!

RELEASE.md
hoffbrinkle: 🤦‍♂️  I've been using my zsh aliases (`gswc` in this case) for too long...

RELEASE.md
hoffbrinkle: Made me look it up.  This is driven by the `--cleanup` option to commands like `git commit` or `git tag`.  The `default` is to `strip`, which does things like remove comments and extra whitespace.  `verbatim` allows that to be kept.

PR #517, opened 2025-08-19
Re-enable safety checks in convert-and-compare

### LINKED ISSUE #516: Common unit conversions ignore safety checks
This has disabled a key safety feature of the library.  The only way we can test for it is with one of the "uncomment" tests that we run manually, and only at release time.  Looks like it was broken in #432 with the `convert_and_compare()` utilities.  See here: we are [using the explicit-rep versions](https://github.com/aurora-opensource/au/blob/3ea10cae344c8adce6f68bb41ee08f509b8a4225/au/quantity_point.hh#L334-L335), which doesn't opt in to the safety checks until #122 is done in [0.6.0](https://github.com/aurora-opensource/au/milestone/9)!

Thankfully, our new policy based overloads should make this a trivial fix.  Looks like we'll get an invocation of the explicit rep APIs that _does_ use safety checks a little sooner than we thought!

chiphogg:
We did not mean to disable these in #432.  Maybe my excuse is that my
brain was already living in the glorious future where #122 is landed.
At any rate, thank goodness for those policy argument overloads!  This
would have been a much more complicated fix without them.

I tested this by uncommenting the "feeters" test case, and verifying
that it now produces a compiler error, as it should.  I also uncommented
the quantity point test case where we check inheriting the overflow
safety surface.

Fixes #516.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -732,8 +732,9 @@ constexpr auto convert_and_compare(Quantity<U1, R1> q1, Quantity<U2, R2> q2) {
     using U = CommonUnitT<U1, U2>;
     using ComRep1 = detail::CommonTypeButPreserveIntSignedness<R1, R2>;
     using ComRep2 = detail::CommonTypeButPreserveIntSignedness<R2, R1>;
-    return detail::SignAwareComparison<UnitSign<U>, Op>{}(q1.template in<ComRep1>(U{}),
-                                                          q2.template in<ComRep2>(U{}));
+    return detail::SignAwareComparison<UnitSign<U>, Op>{}(
+        q1.template in<ComRep1>(U{}, check_for(ALL_RISKS)),
+        q2.template in<ComRep2>(U{}, check_for(ALL_RISKS)));
 }
 }  // namespace detail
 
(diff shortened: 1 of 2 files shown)


au/quantity.hh
```diff
@@ -732,8 +732,9 @@ constexpr auto convert_and_compare(Quantity<U1, R1> q1, Quantity<U2, R2> q2) {
     using U = CommonUnitT<U1, U2>;
     using ComRep1 = detail::CommonTypeButPreserveIntSignedness<R1, R2>;
     using ComRep2 = detail::CommonTypeButPreserveIntSignedness<R2, R1>;
-    return detail::SignAwareComparison<UnitSign<U>, Op>{}(q1.template in<ComRep1>(U{}),
-                                                          q2.template in<ComRep2>(U{}));
+    return detail::SignAwareComparison<UnitSign<U>, Op>{}(
+        q1.template in<ComRep1>(U{}, check_for(ALL_RISKS)),
+        q2.template in<ComRep2>(U{}, check_for(ALL_RISKS)));
```
hoffbrinkle, line 737: When checking becomes available on the explicit Rep APIs, do we then remove these policies?

chiphogg: We can.  And it would be better in some ways if we do, because it would remove some clutter.  But there's also no harm in them being there, so the main answer is that it just shouldn't be a big deal either way!

chiphogg: Put another way: if I got a PR at that time to remove them, I'd approve it.  I might even make a PR myself to remove them.  But I'm not going to put any effort at all into reminding myself to make said PR, because it just matters so little to me.

PR #524, opened 2025-08-20
Apply safety checks to explicit rep overloads

chiphogg:
This is the future proof release for #122 for 0.5.0, paving the way for
the 0.6.0 release.

### DIFF
--- a/au/quantity_point_test.cc
+++ b/au/quantity_point_test.cc
@@ -134,7 +134,7 @@ TEST(QuantityPoint, CanGetValueInDifferentUnits) {
 }
 
 TEST(QuantityPoint, IntermediateTypeIsSignedIfExplicitRepIsSigned) {
-    EXPECT_THAT(milli(kelvins_pt)(0u).coerce_as<int>(celsius_pt),
+    EXPECT_THAT(milli(kelvins_pt)(0u).as<int>(celsius_pt, ignore(TRUNCATION_RISK)),
                 SameTypeAndValue(celsius_pt(-273)));
 }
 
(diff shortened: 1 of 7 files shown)


au/quantity_point_test.cc
```diff
@@ -134,7 +134,7 @@ TEST(QuantityPoint, CanGetValueInDifferentUnits) {
 }
 
 TEST(QuantityPoint, IntermediateTypeIsSignedIfExplicitRepIsSigned) {
-    EXPECT_THAT(milli(kelvins_pt)(0u).coerce_as<int>(celsius_pt),
+    EXPECT_THAT(milli(kelvins_pt)(0u).as<int>(celsius_pt, ignore(TRUNCATION_RISK)),
                 SameTypeAndValue(celsius_pt(-273)));
 }
```
hoffbrinkle: I don't think we need this as part of the future proofing PR?

Or do we intend to pull this in to main for 0.6.0 as the basis of that work?

chiphogg: My default approach for most/all of these was to make a future cherry-pick as easy as possible.  That said, more details on this particular one...

Turning on these policies in a naive way breaks `coerce_as`: it no longer coerces!  I saw this test failing, so I just cherry picked the line from the `coerce` future-proof PR.  Later on, I realized what I actually needed to do was to fix the `coerce_as` implementation (in the way that you can see in this PR).

I forgot to revert this change, but it doesn't matter: it's what we will ultimately want it to look like anyway.

hoffbrinkle: Fair enough.  I could mention in passing that we could also clean up the stuff on line 738 in `quantity.hh` (and similar in `quantity_point.hh`) if we were so inclined... 😁

Issue #485, opened 2025-07-30
Upgrade repo to bzlmod


chiphogg:
Our bazel version gets more and more out of date all the time, as does our way of interacting with bazel via a WORKSPACE file.  We will need to bite the bullet and upgrade to a modern bazel version that uses bzlmod.

PR #534, opened 2025-09-08
Migrate to bazel 7

hoffbrinkle:
This gives us the ability to work on the transition to bzlmod.

Helps #485.


MODULE.bazel.lock
```diff
@@ -0,0 +1,110 @@
+{
```
chiphogg, line 1: How does this file get generated, and how will we update it?

chiphogg (approved):
Very nice!  Thank you for taking this up. :pray:  I checked out this branch, and ran all the tests on several different toolchains, and everything looks good.

What work will need to be done in follow on PRs before we can call #485 done?

MODULE.bazel.lock
hoffbrinkle: The default behavior of [`--lockfile_mode`](https://bazel.build/reference/command-line-reference#flag--lockfile_mode) is update, so it will be updated automatically as part of the build process (so nothing special required, other than to check it in when it gets updated).

We could consider setting the mode to `error` for CI.

PR #536, opened 2025-09-09
Update buildifier deps

hoffbrinkle:
This gets buildifier building with the current setup.


hoffbrinkle:
> I see we're updating the WORKSPACE file, rather than migrating this to bzlmod. Is there a reason this doesn't work on bzlmod, or was this just an easier step in that direction?

At least a step in that direction.  The versions we were using a) didn't work for me (I was running my own buildifier) and b) were too old to be supported in the Bazel Central Registry.  This at least gets us a bit closer.

The WORKSPACE file is going to be around until I figure out what to do with the toolchain.  We currently pass config settings to the toolchain (via `@//build:clang14_requested`).  With bzlmod, the module can't see anything in the root workspace, so I'll have to figure that out.

For `buildifier` it is not directly supported with BCR, so I'll have to figure out if we want to keep building `buildifier` from source or just use a [prebuilt](https://registry.bazel.build/modules/buildifier_prebuilt).

chiphogg:
> For `buildifier` it is not directly supported with BCR, so I'll have to figure out if we want to keep building `buildifier` from source or just use a [prebuilt](https://registry.bazel.build/modules/buildifier_prebuilt).

Prebuilt might be nice, especially in that it would avoid randomly waiting for rebuilds that get triggered from time to time.  This manifests as a save command (for format-on-save setups) randomly taking tens of seconds, or even some minutes.

PR #541, opened 2025-09-19
Add Alternatives section for ongoing maintenance

### LINKED ISSUE #532: Document ease of upgrading in Alternatives page
This is a particular strength of the library.  We should advertise it!

chiphogg:
This is a key strength and competitive advantage of Au, so it's time we
started giving ourselves credit for it!

The new section speaks for itself: a good library both _has_ bugfixes
and improvements, and makes it easy to _get_ them.  As for the
assessments, here's my reasoning:

- On "active maintenance":
    - Boost units should get the lowest rating here, and it should be
      "poor": no releases since 2010, and no substantive releases since
      2007 (!).  There are issues on GitHub, and even some commits, but
      they seem to be scattered and sporadic.
    - nholthaus and SI are "fair" because they're a mixed bag.
      nholthaus had a new release this year, but gets dinged for having
      a lot of unresponded issues.  SI has responded to _most_ issues,
      but has also now gone several years without a release.
    - IMO both mp-units and Au are doing a great job here.  Very active
      development, and solid responsiveness to issues.

- On "smooth upgrades":
    - mp-units is an interesting case...
        - The 0.8.0-to-2.x breakage was _necessary_ in order to get all
          the improved interfaces, because they were so radically
          different from what came before.  But it was a huge deal for
          users, many of whom were caught off guard by this, and had
          been treating it as "production" software: I saw a lot of
          complaints.
        - Still, their release notes are excellent, and I do believe
          this was a one-time thing, so this should be "fair" and not
          "poor".
    - SI is "good by default": I don't see any reason to ding them for
      anything here.
    - On nholthaus, I got direct clarification from the author in the
      linked issue.  I give them credit as "good" because even though
      2.x to 3.x requires a megadiff, 2.x still gets backports.
    - No sense assessing boost when they don't have any upgrades.
    - Au is outstanding here.  We always provide "bridging" syntax that
      works for both old and new versions, and our future proof releases
      are something I haven't seen anywhere else.  Basically, we bend
      over backwards to enable every upgrade to be as incremental as
      possible; yet, we still avoid stagnation, and even routinely make
      major interface changes.

Ultimately, the reason Au shines here is because it is very widely used
in production in a large codebase (Aurora's), and the lead "units
maintainer" in that codebase (i.e., me) is fanatical about avoiding
megadiffs.

These rankings pass the smell test for me.  On active maintenance: boost
is clearly the most stale by a wide margin; Au and mp-units are both
exemplary; and nholthaus and SI are more current than boost but have
significant issues.  And on ease of upgrade, Au is a clear "best", boost
a clear "not applicable", and mp-units caused the most significant
breakage of the remaining 3 options.

Fixes #532.

### DIFF
--- a/docs/alternatives/index.md
+++ b/docs/alternatives/index.md
@@ -165,7 +165,7 @@ experience.
 
 2. You will get **more compiler errors** that developers will need to understand and fix.
 
-These costs can bring significant benefits, but we still want them to be as small as possible.
+These costs purchase significant benefits, but we still want them to be as small as possible.
 
 <table>
     <tr>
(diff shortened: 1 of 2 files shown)


docs/alternatives/index.md
```diff
@@ -244,13 +244,148 @@ These costs can bring significant benefits, but we still want them to be as smal
...
+                </li>
+            </ul>
+        </td>
+        <td class="good">
+            <ul>
+                <li class="check">Regular commits and releases</li>
+                <li class="check">
+                    <a
+                    href="https://github.com/aurora-opensource/au/issues?q=is%3Aissue%20state%3Aopen%20-author%3Achiphogg%20-author%3Ahoffbrinkle%20-author%3Ageoffviola%20-author%3Atobin">External
+                    issues</a> always promptly responded, usually closed
```
hoffbrinkle:
```suggestion
                    issues</a> always promptly responded to, usually closed
```

PR #544, opened 2025-09-28
Use `uintmax_t`, not `size_t`, in `mag`

### LINKED ISSUE #542: mag() helper is limited to std::size_t inputs
The [mag helper](https://aurora-opensource.github.io/au/main/reference/magnitude/#magn) is currently defined as taking a `std::size_t` parameter:

```c++
// A helper function to create a Magnitude from an integer constant.
template <std::size_t N>
constexpr auto mag();
```

This usually works, but on a platform where `std::size_t` is 32 bits, it limits the size of a magnitude that can be created even if the platform `std::uintmax_t` supports larger integers.  The machinery that `mag` interacts with (namely `PrimeFactorizationT`) takes a `std::uintmax_t` parameter, so it would seem reasonable for `mag` to also take a `std::uintmax_t`.

chiphogg:
The point of `mag` is to plug into the prime factorization machinery,
all of which uses `uintmax_t`.  Apparently, we were able to get away
with the existing mismatch for quite some time, but it gets exposed on
architctures where `size_t` is 32 bits.

Fixes #542.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -102,7 +102,7 @@ template <typename MagT>
 constexpr const auto &mag_label(MagT = MagT{});
 
 // A helper function to create a Magnitude from an integer constant.
-template <std::size_t N>
+template <std::uintmax_t N>
 constexpr auto mag();
 
 // A base type for prime numbers.
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -14,6 +14,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


hoffbrinkle (approved):
There's a few other uses of `std::size_t` in this file, should we address those while we're at it?

chiphogg (commented):
Err... whoops.  😅  Nice catch.

Issue #555, opened 2025-11-06
Create and document robust idiom for checking integer division


chiphogg:
It seems reasonable to expect that if `a % b == ZERO`, then `a / unblock_int_div(b)` is safe.  However, this is not true in general.  Consider <https://godbolt.org/z/8vE3M8bf5>, where the condition holds, but `a / unblock_int_div(b)` catastrophically truncates to zero, instead of the real solution of `3`.

One takeaway for me is that `unblock_int_div` is a big hammer, and every usage of it deserves extra scrutiny.  (Au is still far ahead of all other units libraries here, none of whom provide any protection against the dangers of integer quantity division.)

Another takeaway is that we don't have a crisp story for what to tell users to do instead.  What it feels like we _should_ do is convert the inputs to their common unit before performing the division.  (This only makes sense if they have the same dimension, but the same is already true for `a % b` anyway.)  Unfortunately, I don't know a concise and intent-forward way to express this right now.

Maybe we bring back `integer_quotient`, but for same-dimension types only?

I want to give this some thought, and open it up for discussion as well.

PR #558, opened 2025-11-20
Add `divide_in_common_units()` utility

chiphogg:
This is probably what people want when they divide two same-dimension
quantities.  It is also a safe, idiomatic alternative to many if not
most use cases for `unblock_int_div`.

We update the troubleshooting guide while we're at it.

Helps #555, but I think we should add a `quotient_and_remainder` helper
too, to really nail down this idiom.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -16,11 +16,27 @@ module(
     name = "au",
 )
 
-bazel_dep(name = "googletest", version = "1.12.1")  # NOTE: if updating this version, also update the version numbers in `CMakelists.txt`.
+# GoogleTest is required for our publically available `//au:testing` target.  We
+# specify the current version we use here so this will participate in any users
+# BCR resolution, but it won't actually be pulled down to the users workspace if
+# they don't actually depend on a target that depends on `googletest`.
+bazel_dep(
+    name = "googletest",
+    # NOTE: if updating this version, also update the version numbers in
+    # `CMakelists.txt`.
+    version = "1.12.1",
+)
 
+# Development dependencies are below.
+
+# Primary deps that we use directly.
 bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
 bazel_dep(name = "fmt", version = "11.0.2", dev_dependency = True)
 bazel_dep(name = "rules_python", version = "1.6.0", dev_dependency = True)
+
+# Configuration of above deps.
+
+# Python configuration and pips
 
 python = use_extension(
     "@rules_python//python/extensions:python.bzl",
@@ -30,13 +46,17 @@ python = use_extension(
 python.defaults(python_version = "3.10")
 python.toolchain(python_version = "3.10")
 
-pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip", dev_dependency = True)
 pip.parse(
     hub_name = "au_pip_deps",
     python_version = "3.10",
     requirements_lock = "//:requirements_lock.txt",
 )
 use_repo(pip, "au_pip_deps")
+
+# Dependencies that don't have BCR equivalents.
+
+# nholthaus_units
 
 http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
@@ -50,12 +70,12 @@ http_archive(
     url = "https://github.com/nholthaus/units/archive/refs/tags/v2.3.3.tar.gz",
 )
 
-# Pull in buildifier
+# buildifier
 
 bazel_dep(name = "rules_go", version = "0.57.0", dev_dependency = True, repo_name = "io_bazel_rules_go")
 bazel_dep(name = "gazelle", version = "0.45.0", dev_dependency = True)
 
-go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
+go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps", dev_dependency = True)
 go_deps.module(
     path = "golang.org/x/tools",
     sum = "h1:DVSRzp7FwePZW356yEAChSdNcQo6Nsp+fex1SUW09lE=",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -15,23 +15,23 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("//build:copts.bzl", "BASE_CLANG_COPTS", "EXTRA_COPTS")
 
-BAZEL_TOOLCHAIN_RELEASE = "0.10.3"
+BAZEL_TOOLCHAIN_RELEASE = "v1.2.0"
 
-BAZEL_TOOLCHAIN_SHA = "b7cd301ef7b0ece28d20d3e778697a5e3b81828393150bed04838c0c52963a01"
+BAZEL_TOOLCHAIN_SHA = "e3fb6dc6b77eaf167cb2b0c410df95d09127cbe20547e5a329c771808a816ab4"
 
 http_archive(
-    name = "com_grail_bazel_toolchain",
+    name = "toolchains_llvm",
     canonical_id = BAZEL_TOOLCHAIN_RELEASE,
     sha256 = BAZEL_TOOLCHAIN_SHA,
     strip_prefix = "toolchains_llvm-{ref}".format(ref = BAZEL_TOOLCHAIN_RELEASE),
     url = "https://github.com/bazel-contrib/toolchains_llvm/releases/download/{ref}/toolchains_llvm-{ref}.tar.gz".format(ref = BAZEL_TOOLCHAIN_RELEASE),
 )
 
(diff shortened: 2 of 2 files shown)


docs/reference/quantity.md
```diff
@@ -789,6 +805,22 @@ the denominator.
...
+    --- their _common unit_ --- before dividing.  This removes the need for `unblock_int_div` in the
+    most common cases, and is generally much safer.  See
+    [`divide_in_common_units()`](#divide-in-common-units) below for more details.
+
+#### `divide_in_common_units()` {#divide-in-common-units}
+
+The `divide_in_common_units()` utility takes two `Quantity` inputs, `a`, and `b`, whose dimension is
+the same.  It first converts each input to their [common
+unit](../discussion/concepts/common_unit.md), and then performs regular division.  The result
+will be a dimensionless and unitless number (or, after [#185], a `Quantity`).
```
hoffbrinkle, line 819: Can we return the value in a `Quantity` of the common unit?

1. This is a new API so nobody will break.
2. We don't have to fix anyone that uses this API prior to #185.
3. If we return a raw number, we then have to infer what the common unit is.

As an example, if I add this to the test cases we have above:

```c++
    EXPECT_THAT(divide_in_common_units(hours(8), minutes(60)), SameTypeAndValue(8));
    EXPECT_THAT(divide_in_common_units(hours(8), hours(1)), SameTypeAndValue(8));
```

Those are two very different `8`s!  This is the kind of thing that a units library should be trying to prevent... 😁

chiphogg: It was tempting to return a `Quantity`, for sure, since that's where the library's headed.  I ultimately decided to keep the definition simple so that it would be easy to reason about.  `divide_in_common_units` should just convert to common units, and then divide.  While we could add a small amount of complexity to "future proof" this right out of the gate, I think it's a little moot because it'll _automatically_ pick up that behavior with the next release anyway.

That said, I think you might have had a little misconception in this response, because the resulting `Quantity` would almost never be expressible in that common unit.  After all, the result will always be dimensionless (in fact, unitless too).  So it turns out that those `8`s aren't actually different at all!

chiphogg: > ...because it'll _automatically_ pick up that behavior with the next release anyway.

As in: there will never be a release where this function would return a `Quantity` while raw division returns a raw number.

hoffbrinkle: Oh, right.  I think it was this comment that threw me off:

```c++
        // (FIXED): 1. Using floating point, we get ~= (hours / minute)(0.2)
        //
        // Despite the unfamiliar units, this is the same as (60 * 0.2) == 12.
        const auto n = hours(8.0) / minutes(40.0);
```

I read that, and for some reason figured `12` was `12` minutes.  Which it would be if you multiplied `0.2 * minutes(60)`.  But that's not what the comment says, I just read it that way.  The comment says `60 * 0.2`.  Not `minutes(60) * 0.2` and not `60 * n`, both of which yield a different answer than `60 * 0.2`.

Would it make sense to update the comment to something like:

```c++
        // Despite the unfamiliar units, it's easily converted using .as(unos) which is 12 Unos.
```

So that's how I got off track.

Then, what I realized is that the new API is really answering a "how many" question.  As in "how many `y` are in `x`?".  Or, more specifically in the case of integral values, "how many whole `y` are in `x`?"[^1].

[^1]: Note that this is slightly different than the question of "how many `y`s would it take to contain `x`", which would be `(x + (y - 1) / y` and is a similarly often asked question in code.

At this point, I was re-reading the diff, and this now jumps out at me:

```markdown
    And how many 40-minute periods are in 8 hours?
```

It turns out you're asking the same question. 😁

The question then is "what's the correct answer?"  One might be tempted to say `12 [40 minute periods]` (i.e. `12 y`).  But really it's `12 of y`.  So then `12 Unos * y` makes sense (and is `<= y`).

As an aside, this concept is something that comes up quite a bit in the code I work on, and is something I've occasionally addressed by using `Unos`.  Given these three forms of a constant:

```c++
constexpr auto RECORD_SIZE = make_constant(Bytes * mag<12>());
constexpr auto RECORD_SIZE = 12;
constexpr auto RECORD_SIZE = make_constant(Unos * mag<12>());
```

The first and second forms being relatively common in our code base, the third is almost non-existent.  The first version is fairly straight forward as to what it is (it's `12 Bytes`).  The second is quite ambiguous: Is that 12 bytes or twelve entries?  The third we can see that it's "not `Bytes`" so we can infer that it's a "count" (e.g. 12 entries).

chiphogg: Good catch: the comment was confusing.  I went back and forth for a while on the wording.  I really felt like I was struggling, but I'm reasonably happy with where I ended up.  Thanks for pointing this out --- especially in public facing content like the troubleshooting guide.

I think what really threw me was `(hours / minute)`, which I naturally read in my head as "hours per minute".  This makes me want to say that it's 1/60, even though I can put on my "units glasses" and clearly see that one hour divided by one minute is 60, not 1/60.  Where I ended up is that when we say something sloppy like "hours per minute is 1/60", we don't mean `(hours / minute) == 1/60.0`; we mean that 1/60 is the number of hours per minute.  That is, implicitly, the number of hours per minute _that gives an identity_: since, after all, unit conversions all boil down to multiplying by 1, cleverly expressed.

Still pretty confusing.  Thank goodness we have units libraries!

Anyway, LMK what you think of the new comment.

au/quantity.hh
```diff
@@ -553,6 +553,17 @@ class AlwaysDivisibleQuantity {
     Quantity<U, R> q_;
 };
 
+// Perform division in the common unit of two inputs.
+//
+// When two quantities have the same dimension, this is what most people probably expect when
+// dividing them.  When they have different dimension, the operation is undefined, and we'll get a
+// compiler error.
+template <typename U1, typename R1, typename U2, typename R2>
+constexpr auto divide_in_common_units(Quantity<U1, R1> q1, Quantity<U2, R2> q2) {
```
hoffbrinkle: Now for the bike shed:

We use `in` and `as` in all sorts of APIs.  When #185 lands, would this become `divide_as_common_units`?  Or maybe it should be `as` now since the internals use `.as`?  Should we use a different name that doesn't include either?

Also, should this be singular, as there's only one common unit?  Or is it plural because we use that common unit for both divided and divisor?

The best I've got so for is `divide_using_common_unit` but it's wordy.

docs/reference/quantity.md
hoffbrinkle: New comment is good.  I probably still would've made the error because I was really thinking the new API was doing something else (I was thinking it was more of a replacement for `unblock_int_div` but it definitely is a more narrow use case).

au/quantity.hh
chiphogg: Oops --- `in` has special meaning in Au!  Good catch.

I think `using` most closely matches what I wanted to express.

docs/troubleshooting.md
```diff
@@ -776,23 +781,49 @@ approach: read the warning below first.
...
+        const auto n = hours(8.0) / minutes(40.0);
+        ```
+
+    === "Fixed (2. Divide in common units)"
+        ```cpp
+        // (FIX NOT APPLICABLE): 2. Divide-in-common-units needs same-dimension inputs
+        // QuantityD<Seconds> t = divide_using_common_unit(meters(60), (miles / hour)(65));
+        // (Will produce a compiler error.)
+
+        // (FIXED): 2. Divide-in-common-units produces 480 min / 40 min == 12
```
hoffbrinkle:
```suggestion
        // (FIXED): 2. Divide-using-common-units produces 480 min / 40 min == 12
```

docs/troubleshooting.md
```diff
@@ -776,23 +781,49 @@ approach: read the warning below first.
...
+        // (FIXED): 1. Using floating point, we get ~= (hours / minute)(0.2)
+        //
+        // To get more familiar units, we can call `.as(unos)`, obtaining `unos(12)`.
+        // This corresponds to the expected result of 12.
+        const auto n = hours(8.0) / minutes(40.0);
+        ```
+
+    === "Fixed (2. Divide in common units)"
+        ```cpp
+        // (FIX NOT APPLICABLE): 2. Divide-in-common-units needs same-dimension inputs
```
hoffbrinkle:
```suggestion
        // (FIX NOT APPLICABLE): 2. Divide-using-common-units needs same-dimension inputs
```

docs/troubleshooting.md
```diff
@@ -776,23 +781,49 @@ approach: read the warning below first.
...
         QuantityD<Seconds> t = meters(60.0) / (miles / hour)(65.0);
+
+        // (FIXED): 1. Using floating point, we get ~= (hours / minute)(0.2)
+        //
+        // To get more familiar units, we can call `.as(unos)`, obtaining `unos(12)`.
+        // This corresponds to the expected result of 12.
+        const auto n = hours(8.0) / minutes(40.0);
+        ```
+
+    === "Fixed (2. Divide in common units)"
```
hoffbrinkle:
```suggestion
    === "Fixed (2. Divide using common units)"
```

PR #563, opened 2025-12-03
Use platforms for managing compiler versions

hoffbrinkle:
This comment in `.bazelrc`

```
# Use platforms for compiling C/C++ code.
```

is now accurate.  This allows us to remove the build flag and instead
specify the target platform to select the compiler.

We can drop the `--platform_suffix` option as output for a `platform` will
be handled by Bazel.

This also allows us to use `target_compatible_with` and drop part of the
patch we've been using on the llvm toolchain.

Helps #485.


chiphogg (approved):
Hey, this looks a lot cleaner!  Thanks 😎

PR #564, opened 2025-12-03
Migrate the LLVM toolchain to MODULE.bazel

### LINKED ISSUE #486: Upgrade `com_grail_bazel_toolchain` to v1.4.0
This is probably blocked on #485.  I tried updating the version, and it started complaining about protobuf packages:

Image

I googled around a little bit, and it seemed like this might be something that is fixed for bzlmod users, but broken for WORKSPACE users.  That's just a hunch, though.

hoffbrinkle:
This also updates us to v1.5.0 of the llvm_toolchain.  No other
differences, just using bzlmod for it.

Helps #485.
Resolves #486.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -34,26 +34,19 @@ bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
 bazel_dep(name = "fmt", version = "11.0.2", dev_dependency = True)
 bazel_dep(name = "rules_python", version = "1.6.0", dev_dependency = True)
 bazel_dep(name = "toolchains_llvm", version = "1.5.0", dev_dependency = True)
+single_version_override(
+    module_name = "toolchains_llvm",
+    patch_strip = 1,
+    patches = [
+        # This is currently on main, should be available in the next release.
+        # https://github.com/bazel-contrib/toolchains_llvm/commit/1d0c58ff030fe232a8c6bb75dc8fb8fb4f2ddb00.patch
+        "//third_party/toolchains_llvm:0001-Support-extra-toolchain-flags-554.patch",
+    ],
+)
 
 # Configuration of above deps.
 
 # Configure and register the LLVM toolchain.
-
-# Since the clang toolchain we're using doesn't let us extract the default flags, we have to
-# manually specify the default flags here. These are copied from:
-# https://github.com/bazel-contrib/toolchains_llvm/blob/069ee4e20ec605a6c76c1798658e17175b2eb35e/toolchain/cc_toolchain_config.bzl#L118
-BASE_CLANG_COPTS = [
-    # Security
-    "-U_FORTIFY_SOURCE",  # https://github.com/google/sanitizers/issues/247
-    "-fstack-protector",
-    "-fno-omit-frame-pointer",
-    # Diagnostics
-    "-fcolor-diagnostics",
-    "-Wall",
-    "-Wshadow",
-    "-Wthread-safety",
-    "-Wself-assign",
-]
 
 # Extra flags we want to pass to the compilers.
 EXTRA_COPTS = [
@@ -64,12 +57,8 @@ EXTRA_COPTS = [
 llvm_11 = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency = True)
 llvm_11.toolchain(
     name = "llvm_11",
-    compile_flags = {
-        "": BASE_CLANG_COPTS + EXTRA_COPTS,
-    },
-    cxx_standard = {
-        "": "c++14",
-    },
+    cxx_standard = {"": "c++14"},
+    extra_compile_flags = {"": EXTRA_COPTS},
     llvm_version = "11.1.0",
 )
 llvm_11.extra_target_compatible_with(
@@ -86,12 +75,8 @@ register_toolchains(
 llvm_14 = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency = True)
 llvm_14.toolchain(
     name = "llvm_14",
-    compile_flags = {
-        "": BASE_CLANG_COPTS + EXTRA_COPTS,
-    },
-    cxx_standard = {
-        "": "c++14",
-    },
+    cxx_standard = {"": "c++14"},
+    extra_compile_flags = {"": EXTRA_COPTS},
     llvm_version = "14.0.0",
 )
 llvm_14.extra_target_compatible_with(
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -195,6 +195,133 @@
         ]
       }
     },
+    "@@pybind11_bazel~//:python_configure.bzl%extension": {
+      "general": {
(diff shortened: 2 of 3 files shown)


build/copts.bzl
```diff
@@ -18,20 +18,3 @@ EXTRA_COPTS = [
     "-Wextra",
     "-pedantic",
 ]
-
-# Since the clang toolchain we're using doesn't let us extract the default flags, we have to
-# manually specify the default flags here. These are copied from:
-# https://github.com/bazel-contrib/toolchains_llvm/blob/069ee4e20ec605a6c76c1798658e17175b2eb35e/toolchain/cc_toolchain_config.bzl#L118
-BASE_CLANG_COPTS = [
-    "--target=x86_64-unknown-linux-gnu",
```
chiphogg, line 26: Why did this flag not get copied?

I assume this was intentional, perhaps related to the new platform changes.  We need to update this comment (in its new home in `MODULE.bazel`) to explain why we departed from [the upstream source](https://github.com/bazel-contrib/toolchains_llvm/blob/069ee4e20ec605a6c76c1798658e17175b2eb35e/toolchain/cc_toolchain_config.bzl#L118).  (I checked, and it's [still on the latest master](https://github.com/bazel-contrib/toolchains_llvm/blob/ace6215bbfe8a76884646b2458b42380e2201607/toolchain/cc_toolchain_config.bzl#L166).)

MODULE.bazel
```diff
@@ -33,9 +33,78 @@ bazel_dep(
...
+    "-Wshadow",
+    "-Wthread-safety",
+    "-Wself-assign",
+]
+
+# Extra flags we want to pass to the compilers.
+EXTRA_COPTS = [
+    "-Wextra",
+    "-pedantic",
+]
```
chiphogg, line 55: This is now duplicated with `copts.bzl`.  Can we de-duplicate?

Alternatively, if `copts.bzl` is about to be deleted in one of your upcoming PRs anyway, then I'm fine with leaving the duplication in place.

MODULE.bazel
```diff
@@ -33,9 +33,78 @@ bazel_dep(
...
+
+llvm_11 = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency = True)
```
chiphogg, line 57: For everything new below, I diffed with the deleted contents in `WORKSPACE`.  It all looks like a reasonable outcome of "this is how you spell this in the new world".  Given the fact that all remote builds pass ✨, this section LGTM.

MODULE.bazel
```diff
@@ -33,9 +33,78 @@ bazel_dep(
 bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
 bazel_dep(name = "fmt", version = "11.0.2", dev_dependency = True)
 bazel_dep(name = "rules_python", version = "1.6.0", dev_dependency = True)
+bazel_dep(name = "toolchains_llvm", version = "1.5.0", dev_dependency = True)
```
chiphogg, line 36: Formerly (see deleted `WORKSPACE` code), we specified the `sha256`.  It seems like it would be nice if we could continue to do that.  Is that possible?

If it is, it's fine (and even preferred) to add all the `sha256` values for all of these `bazel_dep` invocations in some follow-on PR.

build/copts.bzl
hoffbrinkle: Fantastic question.  Maybe I was thinking it was coming from the platform?  The upshot is I did some digging, and there's much better solution available in [main](https://github.com/bazel-contrib/toolchains_llvm/commit/1d0c58ff030fe232a8c6bb75dc8fb8fb4f2ddb00).  I pulled that patch forward so we can just set the extra flags we need and mess with the default flags.

MODULE.bazel
hoffbrinkle: The reproducibility hashes are maintained by bzlmod now.  The starting point for the source of truth is the `MODULE.bazel.lock` file.  In there, we can find these reproducibility hashes:

```
    "https://bcr.bazel.build/modules/toolchains_llvm/1.5.0/MODULE.bazel": "31c7077ef64bafdf2dfb46d4bca321b4e8f143b00ac68b2c31f5ff0c91044b60",
    "https://bcr.bazel.build/modules/toolchains_llvm/1.5.0/source.json": "aecbd0eea924f27bf8d33d3823f1427e1eddc826ff96425e4304b0d7ad6d7ffa",
```

The `source.json` file is what has the archive hash:

```json
{
    "integrity": "sha256-SeacARvKpMmnJGooerH7T37T/efL1zADdMEDD0DSu5U=",
    "strip_prefix": "toolchains_llvm-v1.5.0",
    "url": "https://github.com/bazel-contrib/toolchains_llvm/releases/download/v1.5.0/toolchains_llvm-v1.5.0.tar.gz",
    "patches": {
        "module_dot_bazel_version.patch": "sha256-KyXi0JApxA6Qjpm2xKr4JTFWoefUbl/CtrP2gNsBq0o="
    },
    "patch_strip": 1
}
```

When using bzlmod, Bazel provides a bunch of other tooling for managing integrity.  You can manage which repositories you get modules from, you can run in vendor mode where external dependencies come from a local store, you can check to see what is being used with the `bazel mod` command:

```console
❯ bazel mod show_repo toolchains_llvm
INFO: Invocation ID: 51e8bd81-330a-4958-b4c7-001514b1c220
WARNING: For repository 'googletest', the root module requires module version googletest@1.12.1, but got googletest@1.14.0.bcr.1 in the resolved dependency graph.
## toolchains_llvm@1.5.0:
# <builtin>
http_archive(
  name = "toolchains_llvm~",
  urls = ["https://github.com/bazel-contrib/toolchains_llvm/releases/download/v1.5.0/toolchains_llvm-v1.5.0.tar.gz"],
  integrity = "sha256-SeacARvKpMmnJGooerH7T37T/efL1zADdMEDD0DSu5U=",
  strip_prefix = "toolchains_llvm-v1.5.0",
  patches = ["//third_party/toolchains_llvm:0001-Support-extra-toolchain-flags-554.patch"],
  remote_file_urls = {},
  remote_file_integrity = {},
  remote_patches = {"https://bcr.bazel.build/modules/toolchains_llvm/1.5.0/patches/module_dot_bazel_version.patch": "sha256-KyXi0JApxA6Qjpm2xKr4JTFWoefUbl/CtrP2gNsBq0o="},
  remote_patch_strip = 1,
  patch_args = ["-p1"],
  patch_cmds = [],
)
# Rule toolchains_llvm~ instantiated at (most recent call last):
#   <builtin> in <toplevel>
# Rule http_archive defined at (most recent call last):
#   /ephemeral/cache/bazel_hordijk/13e61321db9cdfceb1da8181639c9087/external/bazel_tools/tools/build_defs/repo/http.bzl:387:31 in <toplevel>
```

MODULE.bazel
hoffbrinkle: This will be deduplicated when the GCC toolchain comes in.  You can't do a `load` in the `MODULE.bazel`, so no easy way to share this with `WORKSPACE`.

MODULE.bazel
chiphogg: Terrific, this seems much better all around!

build/copts.bzl
chiphogg: Even better!

chiphogg (approved):

PR #573, opened 2025-12-10
Support newest and oldest available GCC version

hoffbrinkle:
The toolchain rules we're currently using allow support GCC 12 and 15
natively. We could add additional versions using this toolchain as
needed.


chiphogg (approved):
Nice to be getting some options in gcc for the first time!

Issue #578, opened 2025-12-12
Cannot build with bazel 8 after WORKSPACE deleted


chiphogg:
Example error:

<img width="819" height="547" alt="Image" src="https://github.com/user-attachments/assets/814a817e-e87f-44a9-9f3d-934d12fdfb27" />

Text:

```
$ bazel test //...:all
WARNING: option '--platforms' was expanded to from both option '--config=clang' (source /home/chogg/au/.bazelrc) and option '--config=clang' (source /home/chogg/au/.bazelrc)
WARNING: For repository 'rules_cc', the root module requires module version rules_cc@0.0.4, but got rules_cc@0.2.2 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
WARNING: For repository 'googletest', the root module requires module version googletest@1.12.1, but got googletest@1.14.0.bcr.1 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off
WARNING: option '--platforms' was expanded to from both option '--config=clang' (source /home/chogg/au/.bazelrc) and option '--config=clang' (source /home/chogg/au/.bazelrc)
ERROR: /home/chogg/.cache/bazel/_bazel_chogg/f4320e9077711029fb47fb5166db48f7/external/rules_go+/go/private/extensions.bzl:430:28: Traceback (most recent call last):
        File "/home/chogg/.cache/bazel/_bazel_chogg/f4320e9077711029fb47fb5166db48f7/external/rules_go+/go/private/extensions.bzl", line 268, column 26, in _go_sdk_impl
                _download_sdk(
        File "/home/chogg/.cache/bazel/_bazel_chogg/f4320e9077711029fb47fb5166db48f7/external/rules_go+/go/private/extensions.bzl", line 430, column 28, in _download_sdk
                sdks_by_version.clear()
Error: 'Facts' value has no field or method 'clear'
ERROR: Analysis of target '//au:truncation_risk_test' failed; build aborted: error evaluating module extension @@rules_go+//go:extensions.bzl%go_sdk
INFO: Elapsed time: 1.335s, Critical Path: 0.10s
INFO: 1 process: 8 action cache hit, 1 internal.
ERROR: Build did NOT complete successfully
ERROR: No test targets were found, yet testing was requested
FAILED: 
    Fetching repository @@gcc_toolchain++gcc_toolchains+gcc_15; starting
[...]

PR #580, opened 2025-12-12
Delete stray clang=clang14 line in `.bazelrc`

### LINKED ISSUE #487: Add support for clang 17
Clang 17 isn't the newest version, but it's supported starting in [version 1.0.0 of `toolchains_llvm`](https://github.com/bazel-contrib/toolchains_llvm/releases/tag/1.0.0).  However, that has a minimum version of bazel of 7.0.0, and we're still stuck on 6.x because of #485.  So we might as well upgrade to the latest release, 1.4.0, which means that this is blocked on #486 which is blocked on #485.

chiphogg:
This appears to have been added by mistake (probably resolving a merge
conflict) in #573, and it causes us to still use `clang14` by default.

Fixes #487, again.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -191,7 +191,7 @@ http_archive(
 
 # buildifier
 
-bazel_dep(name = "rules_go", version = "0.57.0", dev_dependency = True, repo_name = "io_bazel_rules_go")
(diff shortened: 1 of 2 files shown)


chiphogg:
Blocked on #579.  Anyway, it's nice to see that #578 (originally discovered locally) is finally reproducing in CI!  :upside_down_face:

PR #584, opened 2025-12-18
Get clang format working with the updated toolchains

### LINKED ISSUE #585: Bazel-based tooling is broken
@hoffbrinkle pointed out that running `clang-format` on main will break.

hoffbrinkle:
With the toolchain updates, we have a new repository name for the llvm
toolchain.  Look for the "latest" llvm toolchain available when running
clang-format.

Update our tool wrapper to run under the current directory so paths
passed in on the command line resolve correctly.

Update the bazel output to suppress more output.

Fixes #585.

### DIFF
--- a/tools/lib/command_from_bazel.sh
+++ b/tools/lib/command_from_bazel.sh
@@ -47,7 +47,7 @@ function make_command_from_bazel_run() {
   TARGET="$1"
   shift
 
-  USER_MESSAGE="Building tool.  If curios to what's taking so long, you can Ctrl-C and run: bazel --nohome_rc $BUILD_CMD $TARGET"
+  USER_MESSAGE="Building tool.  If curious as to what's taking so long, you can Ctrl-C and run: bazel --nohome_rc $BUILD_CMD $TARGET"
 
   # Write message, then run _building_ command.
   # When done: back up; then, write spaces; then, back up again.


chiphogg (changes requested):
Thanks for noticing and fixing this!  I filed #585.

This looks good directionally, with two possible tweaks.

### llvm_latest

I love the idea of creating this toolchain name.  Is it possible to make it an _alias_ of some kind?  I'd rather have the `llvm_17` toolchain named `llvm_17` if possible.

If this is hard --- or even if you don't really know how to do it yet --- feel free to leave it as-is.  The main downside would be that it makes adding new toolchains less "cookie cutter".

I also wonder whether it's worthwhile to update the `.bazelrc` so that `--config=clang` defaults to a `clang_latest`, by name, and same for gcc.  That would be out of scope for this PR, but it's an idea for a potential future change that I got from reading your code here.

### bazel output

I checked how it is when I run the tool automatically within nvim, and was pleased to see that the output doesn't show up.  And it's a great improvement to actually see the progress when I run in the command line!

The downside is that I want the commands to just "feel like" running a binary _once it's already built_.  If no changes are needed, I want it to add no output.  Is that possible?  I tried fiddling with re-adding some of the `ui_event_filters` and/or `noshow_progress`, but I can't seem to retain the baby when I throw out the bathwater.

If it turns out that we really have to take both as a package deal, then I'd really be on the fence about it.  Maybe I'd be ready for us to show the explicit build progress each time?  But let's see if we can split them first.

hoffbrinkle:
> I love the idea of creating this toolchain name. Is it possible to make it an _alias_ of some kind? I'd rather have the `llvm_17` toolchain named `llvm_17` if possible.

The way we have our tooling done right now, it really needs to know the repository name, and with bzlmod, I don't think aliases for repository names are possible (thank goodness).  I had thought about going down that rabbit hole and doing newest/oldest for both clang and gcc, but we can leave that for another day.  In the meantime, I've figured out a way to find the latest llvm toolchain for the clang-format tool to use.

> I checked how it is when I run the tool automatically within nvim, and was pleased to see that the output doesn't show up. And it's a great improvement to actually see the progress when I run in the command line!

Okay, I've managed to get everything back to "super clean, no output."  The problem is we don't really know how long it's going to take before we try to get a tool up to date, so no easy way to print output if it's going to take longer.

tools/lib/command_from_bazel.sh
```diff
@@ -18,24 +18,36 @@ set -eou pipefail
...
-  TARGET="$1"; shift
-  USER_MESSAGE="Building tool.  If slow, you can Ctrl-C and run: bazel --nohome_rc $BUILD_CMD $TARGET"
+function make_command_from_bazel_run() {
+  BUILD_CMD="$1"
+  shift
+
+  TARGET="$1"
+  shift
+
+  USER_MESSAGE="Building tool.  If curios to what's taking so long, you can Ctrl-C and run: bazel --nohome_rc $BUILD_CMD $TARGET"
```
chiphogg:
```suggestion
  USER_MESSAGE="Building tool.  If curious as to what's taking so long, you can Ctrl-C and run: bazel --nohome_rc $BUILD_CMD $TARGET"
```

PR #586, opened 2025-12-19
Update test suite names to follow style guide

hoffbrinkle:
The GoogleTest docs indicate that test suite names should follow class
naming guidelines, and specifically that an underscore in the test suite
or test name is undefined behavior.


chiphogg (approved):
Thank you!  I've been meaning to fix this up for ages.

PR #587, opened 2025-12-19
Update numeric_limit tests to check against std::numeric_limits

hoffbrinkle:
Specifically, the `traps` test was failing on MacOS for `int`.  Update
the tests to verify that `numeric_limits` is indeed specialized for the
type, and verify that the behavior matches the behavior for the
underlying type on the system.

### DIFF
--- a/au/math_test.cc
+++ b/au/math_test.cc
@@ -804,9 +804,15 @@ struct NumericLimits {
     }
 };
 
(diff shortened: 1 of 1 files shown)


chiphogg (approved):
Nice de-duplication --- and the main case better expresses intent for how `Q` relates to `R`.  Thanks!

PR #588, opened 2025-12-19
Use matrix for job configuration

hoffbrinkle:
This allows us to have all the build-and-test configurations specified
in a single location (and will make it easier to matrix other
configurations).


chiphogg (commented):
I think we used to use a matrix, but switched to this approach.  I think the reason was so that we could have separate badges?  Indeed, if you click on [clang17-ubuntu](
https://github.com/aurora-opensource/au/actions/workflows/clang17-ubuntu.yml), you see that it links to a **workflow**.

hoffbrinkle:
Well that is a shame.  The matrix stuff also makes nice when running the workflows locally (e.g. `--matrix=config:gcc1{2,5}`).  There's some [third party stuff](https://github.com/40ants/github-actions?tab=readme-ov-file) out there, but we'd have to stand up a service or something.  There is this [discussion](https://github.com/orgs/community/discussions/52616) but not a lot of traction.

The reusable workflow thing is gives us the badges, but does result in a fair bit of boilerplate being copied around.  I suppose it's the best option for now though.

PR #599, opened 2025-12-24
Add coverage for `//au:std:format`

### LINKED ISSUE #508: Add CI tests for `std::format` support
This is blocked on clang 17 (#487).

chiphogg:
We write some new simple tests for `<format>`, by adapting the test
cases that we already had for `{fmt}`.  Interestingly, these revealed
that we needed some tweaks in the case of clang 17.

We make this new test target `"manual"`, so that it won't get picked
up in our workflows by default.  Unfortunately, many toolchains that
"support C++20" don't actually support `<format>`!  So, we give it
another tag, `"requires_std_format"`, which individual workflows can use
to _explicitly opt out_ if necessary.  It turns out that clang 17 is the
only toolchain we currently have that appears to support `<format>`, so
we add the explicit opt-out everywhere else.

Fixes #508.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -953,21 +953,23 @@ struct QuantityFormatter {
     template <typename FormatContext>
     constexpr auto format(const au::Quantity<U, R> &q, FormatContext &ctx) const {
         value_format.format(q.data_in(U{}), ctx);
-        *ctx.out()++ = ' ';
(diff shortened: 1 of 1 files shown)


timhirsh (approved):

geoffviola:
Instead of introducing a new bazel tag, could we use [__cpp_lib_format](https://en.cppreference.com/w/cpp/experimental/feature_test.html#cpp_lib_format)?

PR #606, opened 2026-01-07
Make every unit definition independent

### LINKED ISSUE #451: Make every unit include file singular
By this, I mean that `#include "au/units/foo.hh"` should provide only `foo`, and no other units, for every file in `au'units`.

This will help keep users IWYU-clean, and in principle should give everyone the fastest achievable compile times.

Each unit will be _defined_ independently of the others, but what really matters is the _coherent web of relationships_ they have with each other.  The unit tests will be sufficient to guarantee those relationships are correct.

### LINKED ISSUE #560: GCC 10.2.0 ICE (Internal Compiler Error) on complex `decltype` in Newtons unit definition
### Problem

RISCV64 GCC 10.2.0 crashes with an internal compiler error when compiling the Newtons unit definition in `au/units/newtons.hh`.

### Error Message

```
au/units/newtons.hh:34:71: internal compiler error: in build_over_call, at cp/call.c:8982
   34 | struct Newtons : decltype(Kilo<Grams>{} * Meters{} / squared(Seconds{})), NewtonsLabel<void> {
      |                                                                       ^
```

The complex `decltype` expression involving template instantiation, operator overloading, and the `squared()` function triggers a bug in GCC 10.2.0's template instantiation or overload resolution code.

### Reproduction

**Toolchain:** `riscv64-unknown-elf-gcc 10.2.0` (SiFive Freedom Tools)

Simply including `au/units/newtons.hh` triggers the ICE.

### Workaround

Replace the `decltype` expression with explicit type construction using `UnitProduct` and related type aliases:

```cpp
// Instead of:
struct Newtons : decltype(Kilo<Grams>{} * Meters{} / squared(Seconds{})), NewtonsLabel<void> {

// Use:
struct Newtons : UnitProduct<UnitProduct<Kilo<Grams>, Meters>, UnitInverseT<UnitPowerT<Seconds, 2>>>, NewtonsLabel<void> {
```

This avoids the complex template instantiation that triggers the compiler bug.
### Environment

- **Toolchain:** riscv64-unknown-elf-gcc 10.2.0-2020.12.8
- **Target:** riscv64-unknown-elf (bare-metal)
- **GCC Bug:** Likely related to template instantiation in `cp/call.c:8982`

### Notes

This appears to be a GCC bug that may be fixed in newer versions. However, bare-metal toolchains often lag behind in compiler versions, so a workaround in Au would improve compatibility.

chiphogg:
The goal of this commit is that each unit should be defined purely by
its dimension, and an essentially "magic" number, with no explicit
reference to any other unit.  They all enter on equal footing.  We rely
on the _unit tests_ --- which we do not change --- to make sure that
every unit has the correct relationship with every other unit.

This simplifies the logical structure of the library, and reduces
entanglements.  It may also slightly improve compile time performance.

The downside is that any code that was relying on _transitive includes_
to bring in some of the units that it needed will break.  The fix is
very simple: include the file directly for every unit that you use.

Finally, we completely abandon user-friendly definitions and aliases in
our unit definitions.  Since these units are included directly with the
library, we have enough test coverage to be perfectly confident that we
got them right. Therefore, we spell out the concrete `Dimension<...>`
and `Magnitude<...>` types directly.  End users must never do this in
their code, so we include a stark extra warning with every instance.
This direct approach should be the easiest possible for compilers to
handle, which we assume fixes #560.

Fixes #451.

### DIFF
--- a/au/units/fahrenheit.hh
+++ b/au/units/fahrenheit.hh
@@ -21,7 +21,6 @@
 #include "au/quantity.hh"
 #include "au/quantity_point.hh"
 #include "au/unit_symbol.hh"
-#include "au/units/kelvins.hh"
 
 namespace au {
 
@@ -33,11 +32,19 @@ struct FahrenheitLabel {
 };
 template <typename T>
 constexpr const char FahrenheitLabel<T>::label[];
-struct Rankines : decltype(Kelvins{} * mag<5>() / mag<9>()) {};
+struct Rankines
+    // In particular, do NOT manually specify `Dimension<...>` and `Magnitude<...>` types.  The
+    // ordering of the arguments is very particular, and could change out from under you in future
+    // versions, making the program ill-formed.  Only units defined within the Au library itself can
+    // safely use this pattern.
+    : UnitImpl<Temperature, Magnitude<Pow<Prime<3>, -2>, Prime<5>>> {};
 constexpr auto rankines = QuantityMaker<Rankines>{};
 struct Fahrenheit : Rankines, FahrenheitLabel<void> {
     using FahrenheitLabel<void>::label;
-    static constexpr auto origin() { return centi(rankines)(459'67); }
+    static constexpr auto origin() {
+        // 459.67 Rankines = 45967 centi-rankines
+        return make_quantity<Centi<Rankines>>(45967);
+    }
 };
 constexpr auto fahrenheit_qty = QuantityMaker<Fahrenheit>{};
 constexpr auto fahrenheit_pt = QuantityPointMaker<Fahrenheit>{};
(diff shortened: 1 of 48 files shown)


au/units/fahrenheit.hh
```diff
@@ -33,11 +32,19 @@ struct FahrenheitLabel {
 };
 template <typename T>
 constexpr const char FahrenheitLabel<T>::label[];
-struct Rankines : decltype(Kelvins{} * mag<5>() / mag<9>()) {};
+struct Rankines
```
geoffviola, line 35: Should this unit be split out to separate file?

chiphogg: Ahhh, you found that.  :sweat_smile:  I noticed that when I was going through this, but I decided to just plow ahead and worry about it later.  I'll put up a PR to give it its own unit file straight away.

chiphogg: #608

Issue #86, opened 2023-02-06
Avoid names which differ only by `T`


chiphogg:
We strive to be "easy to use correctly; hard to use incorrectly".

Right now, one significant way the library is "easy to use incorrectly" is to do with very different type names that differ only by `T`.  Examples:

- `UnitProductT` and `UnitProduct`.
    - The former accepts any units in any order.  The latter has strict rules for ordering and should only ever be auto-generated.
- `CommonUnitT` and `CommonUnit`
    - Same considerations as above.

For `UnitProduct`, I could see using something like `CompoundUnit`.  I don't have any good ideas for an alternative name for `CommonUnit`.

PR #614, opened 2026-01-09
Rename internal alias for corresponding quantity

chiphogg:
This is just internal to the library, but it was still a "names differ
by only `T` suffix" situtaion, so we wanted to fix it.  To emphasize its
internal-only nature, we also move it into the `detail` sub-namespace.

In the unlikely event that any external users are using this, they can
simply find-and-replace and update to the new name.  The number of uses
should be very small at most.

Helps #86.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -179,9 +179,10 @@ class Quantity {
     constexpr Quantity() noexcept = default;
 
     // Implicit construction from any exactly-equivalent type.
-    template <typename T,
-              std::enable_if_t<std::is_convertible<detail::CorrespondingQuantityType<T>, Quantity>::value,
-                               int> = 0>
+    template <
+        typename T,
+        std::enable_if_t<std::is_convertible<detail::CorrespondingQuantityType<T>, Quantity>::value,
+                         int> = 0>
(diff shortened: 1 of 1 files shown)


au/quantity.hh
```diff
@@ -66,7 +66,7 @@ struct AreQuantityTypesEquivalent;
 template <typename T>
 struct CorrespondingQuantity {};
 template <typename T>
-using CorrespondingQuantityT =
+using CorrespondingQuantityType =
```
hoffbrinkle, line 71: This is a different idiom than what I remember seeing in the previous PRs?

chiphogg: Yes, sorry for not calling that out.  `CorrespondingQuantity` is an external-facing struct already, which many users have probably specialized.  We probably would give it _some_ slightly different name today, but a smooth migration path wasn't clear to me.  Since the `T` version is only ever used by us, and only in this one file, I felt that spelling out `Type` was an acceptable compromise for this type.

hoffbrinkle: Would it make sense to put it into the `detail` namespace?

chiphogg: Good call --- done.

Issue #601, opened 2026-01-03
Bazel Au clients broken by `ignore_directories` if bazel is too old


chiphogg:
A subset of the error is `REPO.bazel:15:1: name 'ignore_directories' is not defined`.  This is what I get if I'm trying to use Au in a project using an older version of bazel.

PR #615, opened 2026-01-10
Add CI coverage for bzlmod projects

chiphogg:
After bugs such as #578, #585, and especially #601, we could really
benefit from some CI coverage.  This PR adds a test client for bzlmod,
making sure that both bazel 7 and 8 are covered.  I verified that if we
revert the fix #603, this test would have failed, so the coverage seems
good.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -185,10 +185,19 @@ class Quantity {
                          int> = 0>
     constexpr Quantity(T &&x) : Quantity{as_quantity(std::forward<T>(x))} {}
 
+    // `q.as<Rep>()`, or `q.as<Rep>(risk_policy)`
+    template <typename NewRep,
+              typename RiskPolicyT = decltype(check_for(ALL_RISKS)),
+              std::enable_if_t<IsConversionRiskPolicy<RiskPolicyT>::value, int> = 0>
+    constexpr auto as(RiskPolicyT policy = RiskPolicyT{}) const {
+        return make_quantity<Unit>(in_impl<NewRep>(Unit{}, policy));
+    }
+
     // `q.as<Rep>(new_unit)`, or `q.as<Rep>(new_unit, risk_policy)`
     template <typename NewRep,
               typename NewUnitSlot,
-              typename RiskPolicyT = decltype(ignore(ALL_RISKS))>
+              typename RiskPolicyT = decltype(ignore(ALL_RISKS)),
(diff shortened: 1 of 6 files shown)


MODULE.bazel
```diff
@@ -196,7 +196,7 @@ bazel_dep(name = "gazelle", version = "0.45.0", dev_dependency = True)
 
 # The Go SDK has an issue with `-strict-prototypes` and newer versions of clang.
 # Patch it until the fix is upstream.
-go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
+go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk", dev_dependency = True)
```
chiphogg, line 199: See this [build failure](https://github.com/aurora-opensource/au/actions/runs/20882938917/job/60001946557?pr=615) for why the new CI jobs force this change.

.github/workflows/bzlmod-client-build-and-test.yml
```diff
@@ -0,0 +1,49 @@
...
+# limitations under the License.
+
+# Test that external Bazel projects can consume Au as a dependency.
+# This workflow tests with multiple Bazel versions to ensure backward
+# compatibility.
+#
+# This workflow was added in response to issue #601, where `REPO.bazel`
+# using `ignore_directories` forced users to upgrade to Bazel 8+.
+
+name: bzlmod-client-build-and-test
```
hoffbrinkle:
```suggestion
name: bazel-client-build-and-test
```

MODULE.bazel
hoffbrinkle: Ah, yeah, missed this when doing the original migration.

.github/workflows/bzlmod-client-build-and-test.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b #v4.1.5
+
+      - name: Build client with Bazel ${{ matrix.bazel_version }}
+        run: bazelisk build //:test_au_client
+        working-directory: bzlmod_test_client
+        env:
+          USE_BAZEL_VERSION: ${{ matrix.bazel_version }}
+
+      - name: Run client executable
+        run: ./bazel-bin/test_au_client
```
hoffbrinkle:
```suggestion
        run: bazelisk run //:test_au_client
```

timhirsh (approved):

PR #616, opened 2026-01-10
Avoid `T` suffix in arithmetic helpers

chiphogg:
The helpers for products and powers now no longer need the `T` suffix.
We updated the docs accordingly.

`UnitProduct` is a special case.  It's trickier.  We'll save it for a
future PR.

Helps #86.

### DIFF
--- a/au/dimension.hh
+++ b/au/dimension.hh
@@ -41,19 +41,19 @@ using DimProduct = PackProductT<Dimension, BPs...>;
 template <typename... BPs>
 using DimProductT = DimProduct<BPs...>;
 template <typename T, std::intmax_t ExpNum, std::intmax_t ExpDen = 1>
-using DimPowerT = PackPowerT<Dimension, T, ExpNum, ExpDen>;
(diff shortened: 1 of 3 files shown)


au/dimension.hh
```diff
@@ -38,12 +38,23 @@ struct Dimension {
 // Define readable operations for product, quotient, power, inverse on Dimensions.
 template <typename... BPs>
 using DimProductT = PackProductT<Dimension, BPs...>;
+template <typename... BPs>
+using DimProduct = DimProductT<BPs...>;
+
```
hoffbrinkle: Should we update the new version now and then just delete the alias when when we follow through on the removal?

```suggestion
template <typename... BPs>
using DimProduct = PackProductT<Dimension, BPs...>;
template <typename... BPs>
using DimProductT = DimProduct<BPs...>;
```

chiphogg: Yep, that's the ideal approach.  I didn't worry too much about it when I initially made the PR, but I think this approach will make that later update PR cleaner and easier to review.

PR #626, opened 2026-01-11
Add `.as<T>()` syntax for quantities and points

### LINKED ISSUE #256: Allow `.as<Rep>()` syntax on a Quantity
With the introduction of `Constant`, there was an API introduced for those that allows the `Rep` to be indicated without changing the `Unit` of the constant.  Specifically, one can use `.as<Rep>()` to return a `Quantity` with the desired `Rep` from a `Constant`.

I find this to be a natural API, and often find myself trying to do the same thing on a `Quantity` when I want a `Quantity` with a specific `Rep`.  We have several other options when this needs to be done:

1. `.coerce_as<Rep>(Unit)`
2. `.as<Rep>(Unit)`
3. `rep_cast<Rep>(Quantity)`

The first one is scary.  The second option works, but forces me know the `Unit` even if I don't want the `Unit` to change.  The third option is does a `static_cast` on the underlying value.  I think the third option is best suited for my use cases.

What I'm thinking is an API that is `.as<Rep>()` if I don't want the unit to change, with as much of the "don't accidentally shoot yourself in the foot" checking as the library can provide.

chiphogg:
Fixes #256.

### DIFF
--- a/docs/reference/quantity.md
+++ b/docs/reference/quantity.md
@@ -880,19 +880,11 @@ common unit, and then computing the remainder from performing integer division o
 
 There are two ways to change the representation type of a `Quantity`, `q`, to some target type `T`.
 
-- `rep_cast<T>(q)` is forcing, like a `static_cast`
 - `q.as<T>()` guards against all conversion risks by default, and can take a [conversion risk
   policy](./conversion_risk_policies.md) parameter to customize this behavior.
+- `rep_cast<T>(q)` is forcing, like a `static_cast`
 
 We describe these in more detail below.
-
-### `rep_cast`
-
-`rep_cast` performs a `static_cast` on the underlying value of a `Quantity`.  It is used to change
-the rep.
-
-Given any `Quantity<U, R> q` whose rep is `R`, then `rep_cast<T>(q)` gives a `Quantity<U, T>`, whose
-underlying value is `static_cast<T>(q.in(U{}))`.
 
 ### `.as<T>()` {#as-rep}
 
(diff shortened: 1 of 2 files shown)


docs/reference/quantity.md
```diff
@@ -876,14 +876,71 @@ common unit, and then computing the remainder from performing integer division o
...
     the operation of repeated [subtraction](#subtraction) as well, so this choice is consistent.)
 
-## `rep_cast`
+## Changing the representation type
+
+There are two ways to change the representation type of a `Quantity`, `q`, to some target type `T`.
+
+- `rep_cast<T>(q)` is forcing, like a `static_cast`
+- `q.as<T>()` guards against all conversion risks by default, and can take a [conversion risk
+  policy](./conversion_risk_policies.md) parameter to customize this behavior.
```
hoffbrinkle: Shall we swap the order of these (with the idea that `as<Rep>()` would be preferable)?

docs/reference/quantity.md
```diff
@@ -876,14 +876,71 @@ common unit, and then computing the remainder from performing integer division o
...
+
+    ```cpp
+    auto length = feet(3);
+    ```
+
+    Then `length.as<float>()` produces a value equivalent to `feet(3.0f)`: that is, it converts
+    a `Quantity<Feet, int>` to a `Quantity<Feet, float>`.
+
+The behavior of `q.as<T>()` is similar to `rep_cast<T>(q)`, except that it guards against
+[conversion risks](../discussion/concepts/conversion_risks.md) by default.  To customize this
```
hoffbrinkle: If so, then re-word this stuff to say `rep_cast` is similar, except that no conversion checking is performed.

docs/reference/quantity.md
chiphogg: Makes sense, thanks!  Done.

I did find a use case in some work I was doing on top of this PR where I started out with `.as<T>(ignore(ALL_RISKS))` (wanting to use the new `.as<T>()` functionality), but I ended up replacing it with `rep_cast<T>()`.  It felt like a cleaner option when what I really just wanted was "the quantity version of `static_cast`".  This made me feel like even though `.as<T>()` is more flexible and has safer defaults, `rep_cast` won't be going away ever.  I also updated the docs to give this guidance: `rep_cast` is a good choice for `static_cast` replacement when you really do want the forcing.

hoffbrinkle (approved):

Issue #528, opened 2025-08-26
feature request - allow disabling floating point implicit conversion


ghorn:
`au` permits implicit floating point conversions [by design](https://aurora-opensource.github.io/au/main/tutorial/103-unit-conversions/#implicit-conversions), including conversions which reduce precision.

GCC and Clang have a `-Wfloat-conversion` flag (implied by `-Wconversion`) which prevents implicit conversions from `double` to `float`. Clang-tidy has the `cppcoreguidelines-narrowing-conversions` check (aliased by `bugprone-narrowing-conversions`) which also enforces this.

In a codebase with these warnings/checks enable, it was surprising to me when a `au::Quantity<Unit, double>` was implicitly cast to `au::Quantity<Unit, float>`. It led to a subtle bug.

I'm guessing that there is no desire to remove the implicit conversions feature. On the other hand, in a codebase with `-Wfloat-conversion` / `cppcoreguidelines-narrowing-conversions`, the `au::Quantity` implicit conversions are essentially a "leak". What are your thoughts on this situation? How could it be possible for a codebase to disallow these implicit conversions? What would you think of something like a `-DAU_NO_IMPLICIT_FLOAT_CONVERSIONS` flag?

ghorn:
> The ideal end state would be that for users with `-Wconversion` enabled, we would raise compiler warnings/errors in exactly the same situations for Quantity as for the underlying raw numeric types.

I agree, that would be idea. I assume the approach would be to have the conversion happen implicitly in `au` as opposed to with `static_cast`, is that right? I'm unclear exactly how it this works, but third party libraries are often included with `-isystem` and  `#include <>`, which masks compiler warnings in those headers. Would `-isystem` mask the warning in this case? Additionally, I've noticed that some conversion operations in template functions are treated differently by static analysis, I'm not sure if this case is affected.

If this approach wouldn't work for these or other reasons, I don't know of another approach to get this behavior, do you?

Question - if we took the approach in (2), would it be forcing stricter `-Wconversion` behavior on users who are not using that warning flag?

> (2) would incur a tremendous upgrade burden, and probably also an ongoing usability burden: it's very common to initialize a float from a literal without an f suffix, for example.

I agree, though anecdotally the big codebase I work on prevents this by linting.

> As to a flag based approach: that's something we strongly want to avoid

[...]

PR #629, opened 2026-01-14
Add `ImplicitConversion` operation

chiphogg:
This is very similar to `StaticCast`, and will play a similar role in
operation sequences, but it uses implicit conversion rather than
`static_cast`.  Since overflow and truncation risks are orthogonal to
the explicitness of the cast, we simply inherit from the implementations
for `StaticCast`.

Helps #528.  Future PRs will both make the choice of conversion
operation customizable, and then change which conversion operation we
choose in various functions.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -797,12 +797,12 @@ constexpr MagLabelCategory categorize_mag_label(Magnitude<BPs...> m) {
 
 template <typename MagT, MagLabelCategory Category>
 struct MagnitudeLabelImplementation {
-    static constexpr const char value[] = "(UNLABELED SCALE FACTOR)";
+    static constexpr const char value[25] = "(UNLABELED SCALE FACTOR)";
 
     static constexpr const bool has_exposed_slash = false;
 };
 template <typename MagT, MagLabelCategory Category>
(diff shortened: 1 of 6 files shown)


au/overflow_boundary_test.cc
```diff
@@ -99,6 +99,32 @@ bool can_overflow_above(Op) {
     return CanOverflowAbove<Op>::value;
 }
 
+// Type-parameterized test infrastructure for cast-like operations.
+//
+// Both `StaticCast` and `ImplicitConversion` have identical overflow boundary behavior (in fact,
+// `ImplicitConversion` delegates its implementation to `StaticCast`).  This wrapper lets us test
+// both with a single set of type-parameterized tests.
```
hoffbrinkle, line 106: Any way we can just do a static assert that this is indeed the case (that `ImplicitConversion` implementation inherits from the `StaticCast` implementation?

hoffbrinkle (approved):

au/overflow_boundary_test.cc
chiphogg: I thought about it a bit, but I don't think so.  `ImplicitConversion` doesn't _inherit_ from `StaticCast`; they're independent implementations that happen to have the same structure.

I'm comfortable that the unit tests give us enough coverage.

Issue #243, opened 2024-06-11
Add integer-only versions of rounding functions


chiphogg:
Right now, all of our rounding functions (the `round`, `floor`, and `ceil` families) are designed as "unit-aware" analogues of their standard library counterparts.  This has the galling consequence that for people working with pure integer types, even though the final result will also be an integer, we need to take a detour through the floating point domain, because that is what the standard library functions do!

It feels like we could do way better, with an approach that Au makes possible which I've never seen before.  We should be able to generate rounding functions that live purely inside of the integral domain.  For example, we could generate a unit that is half the target unit, add "1" in this unit, and truncate the result.  We'd need a bunch of test cases, but this could be really useful (especially for embedded code).

chiphogg:
Another major motivator for this is that these functions could be made `constexpr`, whereas wrappers for the STL functions could not.

PR #637, opened 2026-01-17
Add int-based rounding functions

chiphogg:
We introduce functions such as `int_round_as` as alternatives for every
function in the `round_as` family.  The new functions use bespoke logic,
while the old ones wrap STL functions such as `std::round`.  This gives
the new functions two advantages:

1. They stay purely in the integer domain.
2. They can support `constexpr`.

It feels goofy for an integer input to have to go through floating
point, and all of its weirdness at large values.

I've never seen anything like this before, probably because the notion
of "rounding" an integer type makes little sense in general: it's
_already_ an integer.  In the realm of unit conversions, though, we do
have applications for this.

Helps #243.  A future PR will add "explicit Rep" versions of these.

### DIFF
--- a/bzlmod_test_client/BUILD.bazel
+++ b/bzlmod_test_client/BUILD.bazel
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+
 cc_binary(
     name = "test_au_client",
     srcs = ["test_au_client.cc"],


au/math_test.cc
```diff
@@ -1208,4 +1208,295 @@ TEST(InverseAs, HandlesConversionsBetweenOverflowSafetySurfaceAndRepresentableLi
...
+
+TEST(IntRoundIn, SupportsUnsignedTypes) {
+    EXPECT_THAT(int_round_in(meters, milli(meters)(1'499u)), SameTypeAndValue(1u));
+    EXPECT_THAT(int_round_in(meters, milli(meters)(1'500u)), SameTypeAndValue(2u));
+}
+
+TEST(IntRoundIn, SupportsInt64) {
+    constexpr int64_t large_value = 9'000'000'000'000'000'500LL;
+    EXPECT_THAT(int_round_in(kilo(meters), meters(large_value)),
+                SameTypeAndValue(int64_t{9'000'000'000'000'001LL}));
```
hoffbrinkle, line 1254: Do we need the `int64_t` here if we've got `LL`?

docs/reference/math.md
```diff
@@ -672,6 +682,155 @@ The policy for the rep is consistent with
 [`std::floor`](https://en.cppreference.com/w/cpp/numeric/math/floor).  The output rep is the same as
 the return type of applying `std::floor` to the input rep.
 
+!!! tip
+    If you need to floor a `Quantity` with integral rep, and you want to stay in the integer domain
+    or need `constexpr` compatibility, consider using [`int_floor_in` and
+    `int_floor_as`](#int_floor) instead.
+
+#### `int_round_in`, `int_round_as` {#int_round}
```
hoffbrinkle, line 690: Just looking at how the standard library uses prefixes and suffixes, should the API be `int_round_as` or `round_int_as` or `round_as_int`, or `int_round_int_as`?

hoffbrinkle (approved):

docs/reference/math.md
chiphogg: You forgot `int_round_int_as_int`. 😅 

I wouldn't say I _love_ the name, but I had a hard time coming up with a better one.  Really, it's a shorthand for "integer-domain rounding".  I wanted to keep the function name as short as I could while still making the distinction.

Now that you bring it up, I think it makes sense to explain this in the docs, so I added a sentence to each section.

au/math_test.cc
chiphogg: I didn't think so, until the test failed.  Turns out that `LL` does not necessarily mean "precisely `int64_t`".

PR #638, opened 2026-01-21
Fix bzlmod CI failure by loading `cc_binary`

chiphogg:
Very surprising that this didn't fail before, as apparently this load
statement was always needed for bzlmod.  It started failing in #637 ---
and, interestingly, only on a _second_ commit.  Still, the fix is
straightforward.

### DIFF
--- a/bzlmod_test_client/BUILD.bazel
+++ b/bzlmod_test_client/BUILD.bazel
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+
 cc_binary(
     name = "test_au_client",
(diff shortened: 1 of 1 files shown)


chiphogg:
> > Very surprising that this didn't fail before
> 
> Should we make a change to the github actions workflow to catch this class of errors?

That's the weird thing.  It _did_ catch this error!  But for whatever reason, it didn't seem to catch it until several commits after the original error had landed.

We saw something similar with several of the bazel changes in the migration to bzlmod.  I think `rules_go` was the wrong version or something.  It never triggered while @hoffbrinkle was landing PRs, and only triggered later when I tried to run the code.  These failures have only seemed to happen with bzlmod-related code.  It's a little unsettling, but at least we do still seem to get the coverage _eventually_.  :sweat_smile:

PR #639, opened 2026-01-22
Add explicit-rep forms of int rounding functions

### LINKED ISSUE #243: Add integer-only versions of rounding functions

chiphogg:
Includes doc updates.

Fixes #243.

### DIFF
--- a/au/math.hh
+++ b/au/math.hh
@@ -682,6 +682,33 @@ constexpr auto int_round_as(RoundingUnits rounding_units, QuantityPoint<U, R> p)
 }
 
 //
+// Rounding function that does not leave the integral domain.  Does not use `std::round`.
+//
+// This is the "Explicit-Rep" format (e.g., `int_round_as<int>(rounding_units, q)`).
+//
+// (a) Version for Quantity.
+template <typename OutputRep, typename RoundingUnits, typename U, typename R>
+constexpr auto int_round_as(RoundingUnits rounding_units, Quantity<U, R> q) {
+    static_assert(std::is_integral<OutputRep>::value, "int_round_as output must be integral");
+
+    auto trunced = q.template as<OutputRep>(rounding_units, ignore(TRUNCATION_RISK));
+    trunced.data_in(rounding_units) +=
+        (q - trunced).template in<OutputRep>(rounding_units / mag<2>(), ignore(TRUNCATION_RISK));
+    return trunced;
+}
+// (b) Version for QuantityPoint.
+template <typename OutputRep, typename RoundingUnits, typename U, typename R>
+constexpr auto int_round_as(RoundingUnits rounding_units, QuantityPoint<U, R> p) {
+    static_assert(std::is_integral<OutputRep>::value, "int_round_as output must be integral");
+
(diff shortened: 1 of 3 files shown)


au/math.hh
```diff
@@ -681,6 +681,33 @@ constexpr auto int_round_as(RoundingUnits rounding_units, QuantityPoint<U, R> p)
...
 }
 
+//
+// Rounding function that does not leave the integral domain.  Does not use `std::round`.
+//
+// This is the "Explicit-Rep" format (e.g., `int_round_as<int>(rounding_units, q)`).
+//
+// (a) Version for Quantity.
+template <typename OutputRep, typename RoundingUnits, typename U, typename R>
+constexpr auto int_round_as(RoundingUnits rounding_units, Quantity<U, R> q) {
```
hoffbrinkle, line 691: Probably a bit late to bring this up, but should we provide just a single API:

```suggestion
constexpr auto int_round(RoundingUnits rounding_units, Quantity<U, R> q) {
```

Then let the user do an `in` if that's what they need?

It would reduce the API surface we need to maintain and test, encourages users to stay in the library, and allows the user to be explicit about `Unit` and/or `Rep` they need the result in.

Same with the explicit `Rep` stuff, can we just use `int_round` then have the user do a `.as<Rep>()` if they need it?

chiphogg: Staying in the library, and reducing API surface, are appealing benefits.  I think the downside is the grammar/readability.  The `_as` or `_in` suffix is a signal to the reader to expect a unit slot.  It's good that you retained the unit slot (we are very explicit in our docs about why we need it), but I think the new name would make it a little harder for the reader to guess.

Additionally, we need the `int_` versions to be equivalent to the non-`int_` versions to reduce cognitive load, and those APIs are pretty well established by this point.

That said: I'm not convinced these APIs are the ultimate best possible form.  I'd be up for discussing ideas offline, maybe in our team meeting next week.

One interesting example (in upcoming PRs) is that the `Constant` versions of these functions return a _`Constant`_, not a `Quantity` as I had originally assumed.  I think this is more powerful and flexible overall.  But it does lead to surprising consequences: for instance, with `Constant`, the `_in` functions _require_ explicit rep, and the `_as` functions _forbid_ it!  My upcoming PRs document this clearly and explain why (`Constant` has no rep, and everything flows from there).  But it would be good to review these design decisions overall, together, before the release.  (As another example of an open question: should we add an explicit rep `_as` function for `Constant`, which would return a _`Quantity`_?  Food for thought.)

PR #641, opened 2026-01-23
Consolidate `int_` rounding helpers

chiphogg:
We had a decent amount of repetition, which we can reduce by pulling out
the common part: parsing the unit slot, truncating, and doing something
with the leftover diff.

The "parsing the unit slot" part was hard to generalize, actually, so
we added a new helper: `AppropriateAssociatedUnit<Q, U>` resolves to
`AssociatedUnit<U>` when `Q` is `Quantity`, and resolves to
`AssociatedUnitForPoints<U>` when `Q` is `QuantityPoint`.  We add docs
for this, as this is the only public facing part of this change.

Follow-up to #243.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -116,6 +116,10 @@ constexpr const auto &mag_label(MagT = MagT{});
 template <std::uintmax_t N>
 constexpr auto mag();
 
+// Check whether a Magnitude is representable in type T.
+template <typename T, typename... BPs>
+constexpr bool representable_in(Magnitude<BPs...> m);
+
 // A base type for prime numbers.
 template <std::uintmax_t N>
 struct Prime {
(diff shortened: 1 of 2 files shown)


docs/reference/unit.md
```diff
@@ -621,6 +621,26 @@ associated with quantity points.
     be removed in a future release.  Please migrate all current instances of
     `AssociatedUnitForPointsT` to `AssociatedUnitForPoints`.
 
+### Appropriate associated unit {#appropriate-associated-unit}
```
hoffbrinkle, line 624: Not a huge fan of the name.  "Approriate" for what?  I don't know that it's possible to tell what "appropriate" means from the API.  I don't know that I have better recommendation right now.

"AssoicatedUnit" would be nice if we had "AssoicatedUnitQuantity" and "AssociatedUnitQuantityPoint".

"AssociatedUnitForQuantityOrPoint" doesn't really flow either...

chiphogg: Yeah... this name is really mediocre.  It's "appropriate" in the sense of selecting the "appropriate" container, I guess.  I couldn't figure out a better alternative either.  Perhaps its saving grace is that it's very, very niche.

How about this: I'll land this change, but if any of us thinks of a better name _before the release_, we can change it, and it won't count as a "breaking" change.  That gives us several weeks at least.  If we don't come up with anything, at least I think the name isn't _so_ bad that we can't live with it.

hoffbrinkle: Sounds good.

Issue #607, opened 2026-01-08
Support arithmetic and comparison operators on Constants when safe


matthew-reynolds:
Currently, `Constant`s do not support all arithmetic and comparison operations, as the goal is to ensure perfect representation.

`Constant`s support the following mixins (All mixins below are commutative):
 - `MakesQuantityFromNumber`, which permits
   - `Constant * raw num -> Quantity`
   - `Constant / raw num -> Quantity`
 - `ComposesWith`, which permits
   - `Constant * Constant -> Constant`
   - `Constant / Constant -> Constant`
   - `Constant * QuantityMaker -> QuantityMaker`
   - `Constant / QuantityMaker -> QuantityMaker`
   - `Constant * SingularNameFor -> SingularNameFor`
   - `Constant / SingularNameFor -> SingularNameFor`
 - `SupportsRationalPowers`, which permits
   - `au::pow<integer>(Constant) -> Constant`
   - `au::root<integer>(Constant) -> Constant`
 - `CanScaleByMagnitude`, which permits
   - `Constant * Magnitude -> Constant`
   - `Constant / Magnitude -> Constant`

Plus implicit conversion to `Quantity` and therefore all the `op(Constant, Quantity)` -> `op(Quantity, Quantity)` operations supported on Quantities.

This leaves a few gaps when working purely in the `Constant` domain (Or, equivalently, in the `Constant` and `Magnitude` domain). See below:

### Arithmetic Operators

| Operator    | LHS Operand | RHS Operand | Status                 |
|-------------|-------------|-------------|------------------------|
| `+` (Unary) | `Constant`  | N/A         | Prohibited but desired |
| `-` (Unary) | `Constant`  | N/A         | Permitted              |
| `+`         | `Constant`  | `Constant`  | Prohibited but desired |
| `-`         | `Constant`  | `Constant`  | Prohibited but desired |
| `*`         | `Constant`  | `Constant`  | Permitted              |
| `/`         | `Constant`  | `Constant`  | Permitted              |
| `%`         | `Constant`  | `Constant`  | Prohibited but desired |
| `~` (Unary) | `Constant`  | N/A         | Prohibited             |
| `&`         | `Constant`  | `Constant`  | Prohibited             |
| `\|`        | `Constant`  | `Constant`  | Prohibited             |
| `^`         | `Constant`  | `Constant`  | Prohibited             |
| `<<`        | `Constant`  | `Constant`  | Prohibited             |
[...]

PR #642, opened 2026-01-23
Support comparing a suitable subset of Magnitudes

chiphogg:
We only support the subset of magnitude pairs such that we can compute
an exact answer for the comparison using only 64-bit integer arithmetic.
We document this limitation, and document the new operations as well.

Helps #607.

### DIFF
--- a/au/magnitude_test.cc
+++ b/au/magnitude_test.cc
@@ -25,10 +25,14 @@ namespace au {
 using ::testing::DoubleEq;
 using ::testing::Eq;
 using ::testing::FloatEq;
+using ::testing::Ge;
+using ::testing::Gt;
 using ::testing::IsFalse;
 using ::testing::IsTrue;
 using ::testing::Le;
+using ::testing::Lt;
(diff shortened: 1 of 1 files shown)


au/magnitude_test.cc
```diff
@@ -51,6 +51,57 @@ TEST(Magnitude, SupportsEqualityComparison) {
     EXPECT_THAT(mag_1, Ne(mag_2));
 }
 
+TEST(Magnitude, SupportsOrderingComparison) {
+    // Basic ordering.
+    EXPECT_TRUE(mag<1>() < mag<2>());
```
hoffbrinkle: Do we not want to use the usual tests/syntax?  I would think this would still exercise the operators.

```suggestion
    EXPECT_THAT(mag<1>(), Lt(mag<2>()));
```

chiphogg: Sorry about this --- totally slipped through the cracks!  I will fix them all up.  Nice catch.

PR #643, opened 2026-01-23
Support adding/subtracting some `Magnitude`s

chiphogg:
Again, we're only supporting the easy stuff.  Here, "easy" means that we
first factor out a common denominator, and then use 64-bit integer
arithmetic for the numerators.  If this can't get us an exact answer
with trivially simple logic, then we produce a compile time error.
Still, this is pretty capable: it even handles cases with irrational
common factors, and can even add huge numbers if they have a big enough
common factor.

Includes doc updates.

Helps #607.

### DIFF
--- a/au/constant.hh
+++ b/au/constant.hh
@@ -135,4 +135,36 @@ constexpr Zero make_constant(Zero) { return {}; }
 template <typename Unit>
 struct AssociatedUnitImpl<Constant<Unit>> : stdx::type_identity<Unit> {};
 
+// Relational operators.
+//
+// Note that these inherit the limitations of the Magnitude comparisons: they will not work for
+// every combination of Constant.  We decided that supporting many common use cases was worth this
+// tradeoff.
+template <typename U1, typename U2>
+constexpr bool operator==(Constant<U1>, Constant<U2>) {
+    return UnitRatio<U1, U2>{} == mag<1>();
+}
+template <typename U1, typename U2>
+constexpr bool operator<(Constant<U1>, Constant<U2>) {
+    using SignU2 = Sign<detail::MagT<U2>>;
+    using AbsU2 = decltype(U2{} * SignU2{});
+    return UnitRatio<U1, AbsU2>{} < SignU2{};
+}
+template <typename U1, typename U2>
+constexpr bool operator!=(Constant<U1> lhs, Constant<U2> rhs) {
+    return !(lhs == rhs);
+}
+template <typename U1, typename U2>
+constexpr bool operator<=(Constant<U1> lhs, Constant<U2> rhs) {
+    return (lhs < rhs) || (lhs == rhs);
+}
+template <typename U1, typename U2>
+constexpr bool operator>(Constant<U1> lhs, Constant<U2> rhs) {
+    return !(lhs <= rhs);
+}
+template <typename U1, typename U2>
+constexpr bool operator>=(Constant<U1> lhs, Constant<U2> rhs) {
+    return !(lhs < rhs);
+}
+
 }  // namespace au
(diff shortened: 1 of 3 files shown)


au/magnitude_test.cc
```diff
@@ -116,6 +116,209 @@ TEST(Magnitude, QuotientBehavesCorrectly) {
...
+}
+
+TEST(Magnitude, AdditionOfLargePositives) {
+    constexpr auto two_to_63 = pow<63>(mag<2>());
+    constexpr auto two_to_62 = pow<62>(mag<2>());
+    EXPECT_THAT(two_to_63 + two_to_62, Eq(mag<3>() * two_to_62));
+
+    EXPECT_THAT(two_to_63 + two_to_63, Eq(pow<64>(mag<2>())));
+
+    EXPECT_THAT(two_to_63 + mag<1>(), Eq(two_to_63 + mag<1>()));
```
hoffbrinkle: This one appears to be a tautology.  Should it be something like this?

```suggestion
    EXPECT_THAT(two_to_63 + mag<1>(), Eq(mag<9'223'372'036'854'775'809u>()));
```

au/magnitude_test.cc
```diff
@@ -116,6 +116,209 @@ TEST(Magnitude, QuotientBehavesCorrectly) {
...
+
+    EXPECT_THAT((-two_to_63) + (-two_to_62), Eq(-(mag<3>() * two_to_62)));
+
+    EXPECT_THAT((-two_to_63) + (-two_to_63), Eq(-pow<64>(mag<2>())));
+}
+
+TEST(Magnitude, AdditionOfLargePositiveAndSmallNegative) {
+    constexpr auto two_to_63 = pow<63>(mag<2>());
+
+    EXPECT_THAT(two_to_63 + (-mag<1>()), Eq(two_to_63 - mag<1>()));
```
hoffbrinkle:
```suggestion
    EXPECT_THAT(two_to_63 + (-mag<1>()), Eq(mag<9'223'372'036'854'775'807>()));
```

au/magnitude_test.cc
```diff
@@ -116,6 +116,209 @@ TEST(Magnitude, QuotientBehavesCorrectly) {
...
+
+    EXPECT_THAT(two_to_63 + (-pow<62>(mag<2>())), Eq(pow<62>(mag<2>())));
+}
+
+TEST(Magnitude, AdditionOfLargeNegativeAndSmallPositive) {
+    constexpr auto two_to_63 = pow<63>(mag<2>());
+
+    EXPECT_THAT((-two_to_63) + mag<1>(), Eq(-(two_to_63 - mag<1>())));
```
hoffbrinkle: Maybe?

```suggestion
    EXPECT_THAT((-two_to_63) + mag<1>(), Eq(-mag<9'223'372'036'854'775'807>()));
```

au/magnitude_test.cc
```diff
@@ -116,6 +116,209 @@ TEST(Magnitude, QuotientBehavesCorrectly) {
...
+
+TEST(Magnitude, SubtractionOfLargePositives) {
+    constexpr auto two_to_63 = pow<63>(mag<2>());
+    constexpr auto two_to_62 = pow<62>(mag<2>());
+
+    EXPECT_THAT(two_to_63 - two_to_62, Eq(two_to_62));
+
+    EXPECT_THAT(two_to_62 - two_to_63, Eq(-two_to_62));
+
+    EXPECT_THAT(two_to_63 - mag<1>(), Eq(two_to_63 - mag<1>()));
```
hoffbrinkle:
```suggestion
    EXPECT_THAT(two_to_63 - mag<1>(), Eq(mag<9'223'372'036'854'775'807>()));
```

PR #648, opened 2026-01-26
Support `Constant` in all rounding functions

### LINKED ISSUE #531: Support `Constant` in ceil, floor, and round functions
This would be a nice quality of life improvement.  For example, suppose we had a crypto key with a certain number of bits.  If we want a byte array to hold it, it would be nice to be able to write `ceil_in<size_t>(bytes, MAX_KEY_LEN)` as the array size.

chiphogg:
With all of the machinery we previously built, the implementations have
become quite easy.  A rounded `Constant` is always some integer multiple
of the requested rounding units, or else `Zero`.

Includes doc support.  We also gave the rounding functions' prose a
refresh in the docs, as many small details were incomplete, outdated, or
simply needed improvement.

Fixes #531.

### DIFF
--- a/docs/reference/constant.md
+++ b/docs/reference/constant.md
@@ -375,8 +375,11 @@ For a `Constant` instance `c` and a unit `u`:
     functions.  In particular, an explicit rep is _required_ for the `_in` functions, but
     _forbidden_ for the `_as` functions!
 
-    This is a simple consequence of the fact that `Constant` _has_ no rep: it represents a single
-    abstract value.
+    This is a simple consequence of the fact that `Constant` _has_ no rep.  The type itself
+    represents only a single, specific value (that is, it is a [monovalue
+    type](./detail/monovalue_types.md)).
+
+    Here is the pattern to remember:
 
     - The `_in` functions require an explicit rep because the input `Constant` has no rep, so we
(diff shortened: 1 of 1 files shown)


docs/reference/constant.md
```diff
@@ -339,6 +339,70 @@ For instances `c1` and `c2`:
...
+- `int_floor_in<T>(u, c)` — same as `floor_in<T>(u, c)` for `Constant` inputs
+- `int_ceil_in<T>(u, c)` — same as `ceil_in<T>(u, c)` for `Constant` inputs
+
+!!! note
+    Pay attention to the peculiar pattern of presence and absence of the explicit `Rep` in these
+    functions.  In particular, an explicit rep is _required_ for the `_in` functions, but
+    _forbidden_ for the `_as` functions!
+
+    This is a simple consequence of the fact that `Constant` _has_ no rep: it represents a single
+    abstract value.
```
hoffbrinkle: > single abstract value

Is it really abstract, or is it a single concrete value?  Or a single value?  Or should we say (and link to) it's a monotype?

chiphogg: Good questions.  I tweaked the wording.

PR #649, opened 2026-01-26
Support spaceship operator for `Magnitude`/`Zero`

chiphogg:
This is pretty straightforward, and well in line with our existing `<=>`
support in `Quantity` and `QuantityPoint`: guarded behind the
appropriate feature macros.

Unfortunately, some errors on older compliers actually revealed an
Include What You Use ("IWYU") violation in those existing
implementations.  Since we return `std::strong_ordering` (although we
don't name it), we really should be including `<compare>`.  I've fixed
that here as well, for both new files and old.

Also, a bit of a tangent: every time I find myself needing to "negate"
the result of `<=>`, I always have to look up the incantation.  I think
the C++ standard should simply make `std::strong_ordering` and friends
negatable.  It would be more ergonomic, and make the language easier to
teach.

Helps #607.

### DIFF
--- a/au/math.hh
+++ b/au/math.hh
@@ -484,6 +484,11 @@ template <typename OutputRep, typename RoundingUnits, typename U, typename R>
 auto round_in(RoundingUnits rounding_units, QuantityPoint<U, R> p) {
     return static_cast<OutputRep>(round_in(rounding_units, p));
 }
+// c) Version for Constant.
+template <typename OutputRep, typename RoundingUnits, typename U>
+constexpr auto round_in(RoundingUnits rounding_units, Constant<U> c) {
+    return get_value<OutputRep>(mag_round(unit_ratio(c, rounding_units)));
+}
 
 //
 // The integral-valued Quantity or QuantityPoint, in this unit, nearest to the input.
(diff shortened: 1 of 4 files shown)


docs/reference/magnitude.md
```diff
@@ -247,11 +247,26 @@ limitations](#compile-time-arithmetic-limitations)._
...
+`ZERO OP m`, for any ordering comparison operator `OP`, where by "opposite" we mean:
+
+- `!b` for a `bool` result `b`
+- "Negation" (conceptually) for a three-way comparison (`<=>`) result^[1]
+
+[^1]: The three-way comparison operator `<=>` result is meant to be compared to `0`.  Signed
+integers are a natural model for this.  If we were using _actual_ signed integers, then we could map
+a result onto its "opposite" by simply negating it: that is, we would like to think of this as
+`-(a <=> b)`.  The correct incantation for this in C++ --- `0 <=> (a <=> b)` --- is more
+complicated, but it's still useful to think of it as essentially just negation.
```
hoffbrinkle, line 269: Is this still the correct incantation?

au/magnitude.hh
```diff
@@ -429,6 +440,17 @@ constexpr bool operator>=(Magnitude<BPs...>, Zero) {
     return IsPositive<Magnitude<BPs...>>::value;
 }
 
+#if defined(__cpp_impl_three_way_comparison) && __cpp_impl_three_way_comparison >= 201907L
+template <typename... BPs>
+constexpr auto operator<=>(Zero, Magnitude<BPs...> m) {
+    return 0 <=> get_value<int>(sign(m));
```
hoffbrinkle, line 446: Out of curiosity, how does this work with an `int` for `Rep`?  Does it limit the value of `mag` more than some of our other operations?

chiphogg: Nope, because `sign(m)` can only be +1 or -1.  (Although I admit that choosing a specific concrete rep violates my sense of aesthetics!)

docs/reference/magnitude.md
chiphogg: Yep, this is a fully general incantation for negating the result of a three way comparison.

And yep, I have to look it up, every time. 😅

au/magnitude.hh
hoffbrinkle: Ah, makes sense, thanks!

PR #652, opened 2026-01-30
Make Magnitude sum N-ary and more general

chiphogg:
When we start collecting like terms as part of adding `Constant`
instances to each other, we will naturally form N-ary sums.  To prepare
the way, we introduce a `MagSum<Ms...>` utility to perform this
operation.

Now, we _could_ just implement this in terms of the existing `operator+`
for `Magnitude`.  However, that gets a little weird: it makes the result
depend on the order.  It's possible for a subset of inputs to overflow
(thus making the binary operation return an error), when subsequent
inputs might "rescue" the result (say, by subtracting to bring it back
into range).  Thus, I redid the implementation from scratch, using
modular arithmetic for the result, and keeping track of the (signed)
number of overflows.  This gives us exact information about the true sum
(over a very very wide range of values), although we can't always _use_
this information.  Specifically, we can only use the result when the
overflow is 0 (simple), or when it is -1 _and_ the sum is not 0 (because
if sum _were_ 0 in this case it would actually represent -2^64).

This new, more symmetrical implementation immediately takes the place of
all three `Magnitude`-`Magnitude` implementations of `operator+`.

Helps #607.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -1274,7 +1274,7 @@ struct U64MagSumImpl {
         const int overflows[] = {(IsPositive<Ms>::value ? 0 : -1)...};
 
         U64SumResult result = {0u, 0};
-        for (auto i = 0u; i < sizeof...(Ms); ++i) {
+        for (std::size_t i = 0u; i < sizeof...(Ms); ++i) {
             std::uint64_t old_sum = result.sum;
             result.sum += (overflows[i] >= 0) ? abs_values[i] : -abs_values[i];
             result.overflow += overflows[i] + (result.sum < old_sum);


au/magnitude.hh
```diff
@@ -1276,4 +1252,80 @@ struct CommonMagnitudeImpl<Zero, M> : stdx::type_identity<M> {};
...
+        std::uint64_t sum = 0u;
+        int overflow = 0;
+    };
+
+    static constexpr U64SumResult compute() {
+        const std::uint64_t abs_values[] = {get_value<std::uint64_t>(Abs<Ms>{})...};
+        const int overflows[] = {(IsPositive<Ms>::value ? 0 : -1)...};
+
+        U64SumResult result = {0u, 0};
+        for (auto i = 0u; i < sizeof...(Ms); ++i) {
```
hoffbrinkle: Should this be `std::size_t` to match `sizeof`?  Granted, that be quite an input of `Ms` to process...

```suggestion
        for (std::size_t i = 0u; i < sizeof...(Ms); ++i) {
```

au/magnitude.hh
```diff
@@ -1276,4 +1252,80 @@ struct CommonMagnitudeImpl<Zero, M> : stdx::type_identity<M> {};
...
+            std::uint64_t old_sum = result.sum;
+            result.sum += (overflows[i] >= 0) ? abs_values[i] : -abs_values[i];
+            result.overflow += overflows[i] + (result.sum < old_sum);
+        }
+
+        return result;
+    }
+    static constexpr std::uint64_t sum = compute().sum;
+    static constexpr int overflow = compute().overflow;
```
hoffbrinkle, line 1286: Is the compiler smart enough not to do this twice?

au/magnitude.hh
```diff
@@ -1276,4 +1252,80 @@ struct CommonMagnitudeImpl<Zero, M> : stdx::type_identity<M> {};
...
+
+    static_assert((overflow == 0) || (overflow == -1 && sum > 0u),
+                  "Magnitude sum overflowed uint64_t");
+
+    using Sign = std::conditional_t<(overflow == -1), Magnitude<Negative>, Magnitude<>>;
+
+    using AbsMag = std::conditional_t<(overflow == 0) && (sum == 0u),
+                                      Zero,
+                                      // The surprising `sum == 0u` avoids asking for `mag<0>()`.
+                                      // It's fine, because it can never actually be used.
```
hoffbrinkle, line 1296: Is this just because the compiler can't intuit this invariant?

hoffbrinkle (approved):

au/magnitude.hh
chiphogg: I was wondering about that myself.  Ultimately I landed on "I'm curious, but it doesn't matter".  Since we store the result(s) as `constexpr` variables, and then only ever use those, the worst that could happen is a little extra compile time.  (And I do mean "little" --- I really don't see how it could be very much.)

Matter of fact... an earlier iteration just used `compute()` everywhere, so we wouldn't have to add definitions for static member variables down below!  😆

au/magnitude.hh
chiphogg: The precise problem we would have if we deleted ` + (sum == 0u)` is that whenever we take the first branch of the conditional, it amounts to `std::conditional_t<true, Zero, decltype(mag<0>())>`.  Obviously, this "should" always just be `Zero`... but the fact that we have formed `mag<0>()` at all, anywhere, is enough for the compiler to bail out.

hoffbrinkle: Fair enough.

PR #654, opened 2026-02-03
Add `UnitSumPack`, with labels

chiphogg:
The hardest part about adding and subtracting `Constant` instances is
that we need a unit type to represent this sum, and a readable,
automatically generated unit label to go along with it.  The strategy is
to break this into two parts:

1. `UnitSumPack<...>` will be the actual type to hold unit sum
   instances.  End users will _never_ form this directly.  We will not
   bother validating weird edge cases for this, because we will assume
   that every instance the _library_ forms is valid and well formed (at
   least two units, properly ordered, etc.).

2. `UnitSum<...>` is the utility that takes multiple unit instances as
   inputs, and produces the canonical `UnitSumPack<...>` output.  That
   means unpacking any inputs that are already `UnitSumPack<...>`;
   identifying unique "unscaled units"; collecting like terms;
   eliminating unscaled units that get "zeroed out"; putting what's left
   in the right order; and, handling cases where only 0 or 1 units are
   left.

This PR provides the first prong of this strategy.  We make sure that,
given well formed `UnitSumPack` instances, the result has the right
magnitude, and right unit label.

Helps #607.

### DIFF
--- a/au/math_test.cc
+++ b/au/math_test.cc
@@ -63,33 +63,31 @@ constexpr const T &std_clamp(const T &v, const T &lo, const T &hi) {
 constexpr auto SEVEN_THIRDS_METERS = make_constant(meters * mag<7>() / mag<3>());
 constexpr auto FIVE_HALVES_METERS = make_constant(meters * mag<5>() / mag<2>());
 
-// Matcher for label strings.
-MATCHER_P(LabelIs, expected, "") {
-    *result_listener << "which is \"" << arg << "\"";
-    return arg == expected;
-}
-
 // Type trait to check if a type is a Constant.
 template <typename T>
 struct IsConstantType : std::false_type {};
 template <typename U>
 struct IsConstantType<Constant<U>> : std::true_type {};
 
-MATCHER(IsConstant, "is a Constant") {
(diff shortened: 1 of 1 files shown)


au/unit_of_measure_test.cc
```diff
@@ -192,6 +192,78 @@ TEST(UnitProductPack, ExactlyCancellingInstancesYieldsNullPack) {
...
+}
+
+TEST(UnitSumPack, HasMagnitudeThatIsSumOfInputMagnitudes) {
+    StaticAssertTypeEq<MagT<UnitSumPack<Feet, Inches>>,
+                       MagProduct<MagT<Inches>, Magnitude<Prime<13>>>>();
+}
+
+TEST(UnitSumPack, IsAUnit) {
+    EXPECT_TRUE((IsUnit<UnitSumPack<Feet, Inches>>::value));
+    EXPECT_TRUE((IsUnit<UnitSumPack<Meters, Feet, Inches>>::value));
```
hoffbrinkle:
```suggestion
    EXPECT_THAT((IsUnit<UnitSumPack<Feet, Inches>>::value), IsTrue());
    EXPECT_THAT((IsUnit<UnitSumPack<Meters, Feet, Inches>>::value), IsTrue());
```

PR #659, opened 2026-02-12
Bring AU to Au

chiphogg:
It's about time we added the Astronomical Unit!

We'll use the slightly older "AU" symbol, rather than "au", to avoid
confusion with our ubiquitous namespace.

### DIFF
--- a/au/units/astronomical_units.hh
+++ b/au/units/astronomical_units.hh
@@ -1,0 +1,54 @@
+// Copyright 2026 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


au/units/astronomical_units.hh
```diff
@@ -0,0 +1,54 @@
...
+                         Prime<877>,
+                         Prime<7789>>>,
+      AstronomicalUnitsLabel<void> {
+    using AstronomicalUnitsLabel<void>::label;
+};
+constexpr auto astronomical_unit = SingularNameFor<AstronomicalUnits>{};
+constexpr auto astronomical_units = QuantityMaker<AstronomicalUnits>{};
+
+namespace symbols {
+constexpr auto AU = SymbolFor<AstronomicalUnits>{};
```
hoffbrinkle, line 52: So how _much_ confusion would be causing if this was `au` (and/or the label was `"au"`)?  Given it's in the `symbols` namespace, it _seems_ like it would be tractable?

chiphogg: Surprisingly, it... works?  <https://godbolt.org/z/x5qzvE5fK>

Feels a little like tempting fate, but I don't know... maybe we should do it?

chiphogg: Here is an example, <https://godbolt.org/z/K1j1rdsGz>, courtesy of Arthur O'Dwyer on the C++ slack, which convinces me that we should avoid reusing the namespace name `au` as an instance name.

PR #663, opened 2026-02-26
Enable modifying conversion risk policy objects

### LINKED ISSUE #662: Support combining policies (results of `ignore` and `check_for`), not just risks
Individual risks and risk sets are easy to combine via the `|` operator (and we can add others if we need them).  But we can't combine them after we turn them into actual _policies_.

What do I mean?  Suppose we had a function that checks for overflow, and then calls a conversion with `ignore(OVERFLOW_RISK)`.  This conversion might also have truncation risk, and if it does, we want to let users opt out by the usual mechanisms.  But you can't combine `ignore(OVERFLOW_RISK)` with `ignore(TRUNCATION_RISK)`.  And even if you could, the results of these are expressed as `CheckTheseRisks` types.  That means that a natural implementation of `ignore(OVERFLOW_RISK) | ignore(TRUNCATION_RISK)` would produce `check_for(OVERFLOW_RISK | TRUNCATION_RISK)`, **the exact opposite** of the expected `ignore(OVERFLOW_RISK | TRUNCATION_RISK)`!

We could fix this by giving a policy object functions like `.but_ignoring(<risk set>)` and `.also_checking_for(<risk set>)`.  There's really just no good way to do this right now.

chiphogg:
This is by far a cleaner way to do this than anything we can do in the
library today --- we would have to write custom trait structs that reach
into the `detail` namespace.  Of course, these use cases are pretty
rare, but the frequence is _not_ zero (seeing as how we were motivated
by an actual use case we couldn't satisfy).

Includes doc updates.

Fixes #662.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -1010,9 +1010,9 @@ struct GetValueResultImplForDefaultCase<T, Magnitude<BPs...>> {
         if ((widened_result.outcome != MagRepresentationOutcome::OK) ||
             !safe_to_cast_to<T>(widened_result.value)) {
             return {MagRepresentationOutcome::ERR_CANNOT_FIT};
+        } else {
+            return {MagRepresentationOutcome::OK, static_cast<T>(widened_result.value)};
         }
-
-        return {MagRepresentationOutcome::OK, static_cast<T>(widened_result.value)};
     }
 };
 


au/conversion_policy.hh
```diff
@@ -67,6 +67,20 @@ struct CheckTheseRisks<RiskSet<RiskFlags>> {
     constexpr bool should_check(ConversionRisk risk) const {
         return (RiskFlags & static_cast<uint8_t>(risk)) != 0u;
     }
+
+    // Remove risks from the checked set.
+    template <uint8_t OtherFlags>
+    constexpr CheckTheseRisks<RiskSet<RiskFlags & ~OtherFlags>> but_ignoring(
```
hoffbrinkle, line 73: Do we need the `but` as part of the API?  If we think about this as a fluent interface, it seems like the `.` serves as the `but`; the method is going to do what it says.

```c++
    check_for(ALL_RISKS).but_ignoring(OVERFLOW_RISK);
    ignore(ALL_RISKS).but_also_checking_for(TRUNCATION_RISK);
```
vs.

```c++
    check_for(ALL_RISKS).ignoring(OVERFLOW_RISK);
    ignore(ALL_RISKS).checking_for(TRUNCATION_RISK);
```

docs/reference/conversion_risk_policies.md
```diff
@@ -34,3 +34,60 @@ functions:
...
+??? example "Example: clamped conversion"
+    Suppose we want a "clamped" unit conversion: one that automatically clamps out-of-range values
+    to the nearest representable value, instead of overflowing.  We want to let the user pass their
+    own policy to control truncation checking, but whatever that policy is, we need to tweak it and
+    disable overflow checking since we're handling overflow ourselves.  It might look something like
+    this[^1]:
+
+    [^1]: For clarity, note that this function is for illustration purposes only.  It would not
+    handle cases where **all** of the following are true: (a) the initial rep is an integral type,
+    (b) the final rep is _also_ integral, _and_ &#40;c) the conversion factor is rational, but neither
```
hoffbrinkle, line 64: I trust the `&#40;` is rendering correctly in the generated docs...

chiphogg: Yep.  We were forced to do this because `(c)` renders as a copyright symbol.  :laughing: 

If you want to trust-but-verify, there's always `au-docs-serve`.

au/conversion_policy.hh
chiphogg: I went back and forth on this, and I think the `but` improves readability in the _actual use cases_, i.e., where we have a variable named like `policy` instead of a concrete policy using `check_for` or `ignore`.

```cpp
policy.checking_for(TRUNCATION_RISK)
policy.but_also_checking_for(TRUNCATION_RISK)
```

I think the latter is a lot easier to understand.  Similarly:

```cpp
policy.ignoring(OVERFLOW_RISK)
policy.but_ignoring(OVERFLOW_RISK)
```

I also thought carefully about the asymmetry of the "also" word, and I think it makes sense to have it for the positive/additive case, but omit it in the negative/subtractive case.

hoffbrinkle: My question is do we introduce any ambiguity with:

```c++
   policy.checking_for(TRUNCATION_RISK);
   policy.ignoring(OVERFLOW_RISK);
```

or maybe the API isn't quite right?  In this case, these APIs add or subtract policy from the object they are operating on.  Maybe we're trying to be too fluent?

```c++
   policy.add(TRUNCATION_RISK);
   policy.remove(OVERFLOW_RISK);
```

The PR as-is approved as I don't think I have a strong enough argument to request a change.  Feel free to land or continue the discussion.

chiphogg: Thanks!  I thought of that earlier, but decided not to go that route because "add" or "remove" might not be completely unambiguous.  If we "add" a policy, are we adding it to the set that we check, or the set that we ignore?

The lack of simple, mechanical symmetry does bug me a little, but I still think these APIs are the best ones I can come up with.

At least it's a pretty niche use case. 😅

PR #669, opened 2026-04-04
Support SFINAE in Constant representability check

chiphogg:
Right now, we have an unconditional `UnitRatio` invocation, which
produces a hard compiler error.  We thought this was fine, because
nobody should be invoking this check with different-dimension units.
However, it somehow does get invoked in Aurora-internal code.

I tried reproducing the error in the Au repo only, and couldn't do it.
But in any case, we can fix it by just avoiding the hard error.

The most straightforward way to fix it was to just make a new trait,
`IsUnitRatioRepresentableIn<T, U1, U2>`.  This doesn't seem like
something I want to expose to end users, so I kept it as a library
internal detail.


au/constant_test.cc
```diff
@@ -437,6 +437,11 @@ TEST(CanStoreValueIn, ChecksRangeOfTypeForIntegers) {
     EXPECT_THAT(decltype(c)::can_store_value_in<int16_t>(meters / second), IsFalse());
 }
 
+TEST(CanStoreValueIn, ReturnsFalseRatherThanHardErrorForDifferentDimensions) {
+    constexpr bool result = Constant<Meters>::can_store_value_in<long>(Nano<Seconds>{});
```
geoffviola, line 441: Does the size of `long` matter? It's 8 bytes for GCC and 4 bytes for MSVC: [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:7,positionColumn:2,positionLineNumber:7,selectionStartColumn:2,selectionStartLineNumber:7,startColumn:2,startLineNumber:7),source:'%23include+%3Ccstdlib%3E%0A%23include+%3Ciostream%3E%0A%0Aint+main(int,+char**)+%7B%0A++++std::cout+%3C%3C+sizeof(long)%3B%0A++++return+EXIT_SUCCESS%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'1',compiler:vcpp_v19_latest_x64,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),runtimeTools:!(),source:1,stdinPanelShown:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x64+msvc+v19.latest+(C%2B%2B,+Editor+%231)',t:'0')),header:(),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:g152,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'',source:1,stdinPanelShown:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+gcc+15.2+(C%2B%2B,+Editor+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

au/unit_of_measure.hh
```diff
@@ -482,6 +482,15 @@ constexpr auto pow(SingularNameFor<Unit>) {
 template <typename U>
 struct UnitOrderTiebreaker;
 
+// Library-internal helper to check representability of ratio between two units in a numeric type.
+//
+// This will return `false` (rather than producing a compiler error) if the units don't have the
+// same dimension.
+namespace detail {
+template <typename T, typename U1, typename U2>
+struct IsUnitRatioRepresentableIn;
```
geoffviola, line 491: Q: Does it make sense to unit test this?

au/constant_test.cc
chiphogg: In this specific case, definitely not.  The answer doesn't depend on the ratio, because the ratio is not even well defined.  It will return `false` because the dimensions are different.

au/unit_of_measure.hh
chiphogg: I hadn't been planning on it initially, but I agree it makes sense to add them.

chiphogg: Done!

geoffviola (approved):

geoffviola (approved):

PR #670, opened 2026-04-11
Placate Green Hills compiler with ternary

chiphogg:
When the Green Hills compiler encounters an if/else branch whose
predicate is known at compile time, it complains about an unreachable
statement.  Even if the predicate depends on template parameters (i.e.,
is not a strict constant _across all template instantiations_), the
policy triggers on any _individual_ instantiations that contain an
unreachable statement.

Leaving aside the wisdom of this policy as it pertains to generic
programming, the fact is that we're stuck with it and we need to keep
the compiler happy.

To fix this instance, we simply use a ternary operator.  The ternary is
just the same as an if/else branch, but it's a single statement.  In
testing on AV code, this does seem to placate the Green Hills compiler.


geoffviola (approved):
The change looks equivalent to what was there before. Hope to see some green hills results soon 😄

Issue #671, opened 2026-04-17
Support CUDA


chiphogg:
I've done some initial explorations, and it looks feasible.  We would have to get down off our high horse about macros, but I think this is acceptable: it's OK to use macros when there's literally no other feasible alternative.  We would need to put `__host__ __device__` before functions, and `__device__` before variables, in every case where either `__CUDACC__` or `__HIPCC__` is defined.

PR #672, opened 2026-04-21
Add CI coverage for CUDA builds

chiphogg:
This PR just adds a new workflow, and a vacuous unit test.  Future PRs
will expand the test, and do what is necessary to keep it passing.

Helps #671.


rootshellz (approved):
https://jira.int.aurora.tech/browse/SECSDP-444

hoffbrinkle (approved):

timhirsh (approved):
Code LGTM, but before landing I think you'll need to push an empty commit in order to trigger the new job.  I went ahead and allowlisted the action in repo settings, so it should actually trigger now instead of encounter this [startup failure](https://github.com/aurora-opensource/au/actions/runs/24799296843).

PR #673, opened 2026-05-01
Add `AU_DEVICE_FUNC` macro for CUDA/HIP support

chiphogg:
Enable Au types to be used in CUDA and HIP device code by adding
the `AU_DEVICE_FUNC` macro, which expands to `__host__ __device__`
when compiling with nvcc or hipcc, and to nothing otherwise.  We would
much rather not use macros generally, but in this case, I don't know of
any feasible alternatives.  Supporting CUDA with Au is such a huge value
add, and the macro is a small price to pay in the grand scheme of
things.

This doesn't _complete_ our support.  We will also need to add
`__device__` annotations for all of our variables (e.g., `meters`,
`kilo`, etc.).

Helps #671.

### DIFF
--- a/au/BUILD.bazel
+++ b/au/BUILD.bazel
@@ -676,6 +676,7 @@ cc_library(
         "stdx/*.hh",
         "stdx/experimental/*.hh",
     ]),
+    deps = [":config"],
 )
 
 cc_test(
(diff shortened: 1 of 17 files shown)


hoffbrinkle (approved):
Is it okay if I say "man, it is ugly..."

chiphogg:
> Is it okay if I say "man, it is ugly..."

You'd just be telling it like it is :laughing: :sob:

PR #675, opened 2026-05-08
Reflect CUDA support in docs

### LINKED ISSUE #671: Support CUDA

chiphogg:
This doc update is intentionally light.  Since CUDA should "just work"
out of the box, a full doc page wouldn't pull its weight.  We simply
advertise on the README (including a badge), mention nvcc as "best
effort" on our compilers page, and quit tooting our own horn about
macros on the alternatives page.  Fixes #671.


chiphogg:
Badge works!

image

geoffviola (approved):
Great!

Issue #484, opened 2025-07-27
Investigate proof of concept for supporting vectors and matrices


chiphogg:
The goal is to flesh out a candidate concrete plan for implementation, look for risks in the plan, and investigate them.

Right now, "support vectors and matrices" is a little nebulous.  We need to make it concrete.

Off the top of my head, this includes:
- [ ] Check details on "forward declare Eigen types" idea
- [ ] Write some tests with "shaped" units

chiphogg:
Hmm... note to self.  if we end up generalizing `RealPart` into `RealScalarPart`, and start asking users to specialize it, we'll want to move it out of `detail`, lest we end up delaying #412 by another cycle.

chiphogg:
I've completed my planned investigation into vector/matrix support for Au, including a working proof of concept in my local client, and a full test on Aurora's codebase.  Over the past several weeks, we've turned this proof of concept into a series of PRs that have now fully landed to the main branch (except for docs).   I'm ready to present my results for posterity.

# Expected lines of investigation

These are the items I had anticipated investigating when we began.

1. What traits will we need to define for our reps, and do we have a realistic plan to avoid ODR violations?
2. Eigen support: can we avoid taking a physical dependency on it?  Can we preserve Eigen's expression-template-based performance?
3. Can we support both read and write of individual elements with `[i]` and `(i, j, ...)` syntax?
4. Can we see a pathway to evolve these APIs to support heterogeneous units?  In particular, how will we handle the indices, which for heterogeneous units _cannot_ be a simple numeric type (because return values will have different _types_)?

# Executive summary

- Users will be able to create vector and matrix quantities _naturally_ (e.g., `auto q = meters(Eigen::Vector3d{1., 2., 3.})`)
- `q[i]` and `q(i, j, ...)` **read** access will automatically be available whenever the underlying rep supports it, producing a `Quantity`
- Doing **write** access without invoking UB is harder, but we can support `q.mutable_view()[i]` and `q.mutable_view()(i, j, ...)` for this.
    - **Rejected alternative:** `q[i]` could have supported read/write access by using a proxy type, but this `vector<bool>`-style approach was fraught with lifetime issues.  `mutable_view` puts the lifetime issues front and center.
    - It's critically important that `q[i]` itself not be assignable (because we'd silently discard the result!), so we ref-qualify all assignment operators to apply to lvalues only.
[...]

Issue #453, opened 2025-07-10
Add "fixed point rational" operation and use it for nontrivial N/D on integers


chiphogg:
See mpusz/mp-units#615 for ideas.

The problem is that there are certain unit conversions that are simply impossible in Au currently, safety checks or no.  Like, we can't convert between radians, and degrees or revolutions.  We have no way to form that computation in integral types.

The solution would be to "lift" the variable to be converted into a double-wide integer.  Then we could multiply it by the representation of, say, $2^N M$ for a conversion factor $M$ (assuming single-width is $N$ bits).  Finally, we discard the least-significant $N$ bits.

For a type like `int64_t` that doesn't have any twice-as-wide target, we would need to implement a manual fallback with separate `int64_t` / `uint64_t` variables for the high and low bits.

Besides unlocking computations we can't currently do, it should also be more efficient and less prone to overflow.

Once #387 is complete, it will expose this deficiency more acutely.  We can live with that for the [0.5.0](https://github.com/aurora-opensource/au/milestone/8) release, but it would be nice to get it in the next one.

chiphogg:
I started playing with this a little bit, and was embarrassed to find that I had uncritically accepted one of the claims of this method's strengths.  When I first read about this method, I read that a single (double-wide) multiplication, followed by truncation, would give the exact result that a multiplication-and-division (using integer division) would provide.  This is _often_ true, but not _always_.

As a random example, for `uint8_t`, (7/3) gets represented as (256 * 7 / 3), which is 597 (the truncated result of 597.3333...).  If we wanted to apply this factor to 3 (i.e., computing 3 * (7/3)), we find that (3 * 597 / 256) is 6.996..., which truncates to `6`, not `7`!  If we went with a policy of taking the ceiling to get the representation (using 598 for 597.3333...), then it would produce the correct result in this case, but not in others: we could cook up similarly unappealing examples.

Integer arithmetic, unlike floating point arithmetic, is exact.  I don't think we can accept any solution that fails to reproduce what users would get with raw numeric type variables.  So this solution isn't suitable for integer reps with nontrivial rational factors.

I do still think this is a great idea for _irrational_ factors on integral reps, which we cannot handle at all today.  I'll plan to pursue it for those use cases only.

For the integer-with-nontrivial-rational case, we have various options.

1. Status quo
[...]

PR #676, opened 2026-05-15
Improve overflow bounds for rational scaling factors by using modular decomposition

jmontgomery-aurora:
Units variables of integral types with large magnitudes can overflow unnecessarily when applying the scaling factor $$\frac{P \cdot X}{Q}$$.

We can mitigate this for small values of P and Q (which should be most cases) by rewriting X as a quotient and remainder relative to Q, which gives us $$P \cdot \lfloor X / Q \rfloor + \frac{P \cdot (X \pmod Q)}{Q}$$. This eliminates overflow as long as $$P \cdot (Q - 1) \le \text{TypeMax}$$. Factoring out $$GCD(P, Q)$$ gives us an even larger range of non-overflowing factors P and Q. 

The practical result is that for reasonably sized factors and most larger factors, the entire type range becomes usable without overflow. There's also some truncation risk changes to accommodate the new operation.

### DIFF
--- a/au/abstract_operations.hh
+++ b/au/abstract_operations.hh
@@ -221,8 +221,6 @@ struct ScaleByRational {
             constexpr auto g = au_gcd(p_raw, q_raw);
             constexpr auto p = p_raw / g;
             constexpr auto q = q_raw / g;
-            static_assert(q <= 1 || p <= std::numeric_limits<RealPart<T>>::max() / (q - 1),
-                          "p*(q-1) overflows T; unit ratio cannot be applied to this integral type");
             return static_cast<T>(p * (value / q) + p * (value % q) / q);
         } else {
             return static_cast<T>(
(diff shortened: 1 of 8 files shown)


chiphogg:
Thanks for this contribution!  I haven't looked into it in detail yet, but it'd be exciting to avoid overflow in more cases.

Right now, I am knee deep in implementing #484, which will provide the first support for vectors and matrices for #70.  I probably won't have any review bandwidth until that gets done.  Looks like there are some test failures anyway, now that I've unblocked the jobs.

While you're waiting for me to be able to review, what's the relationship of this change with the already-planned #453?  That change would essentially eliminate unnecessary overflow from conversions with non-trivial rational factors, at the cost of using (or, for 64 bits, synthesizing) double-wide integer types to represent the conversion as a pure multiplication, and then discarding half the bits.  I'm asking because I think there's a decent chance that work might obviate this PR.

jmontgomery-aurora:
Widening is a cleaner strategy if you have a larger native type, but I didn't want to pay the cost of the synthetic division in the internal code where I'm seeing this crop up. However, there's a better fix for that particular issue than changing au, so this is just a demo PR to see what you think of the overall approach.

Looking back through, I see some issues with the min/max template bounds that are probably causing the test failures.

PR #680, opened 2026-06-05
Update `min`/`max` to common implementation

chiphogg:
Walter Brown has given talks pointing out the design flaw in `std::min`
and `std::max` (which Stepanov has acknowledged): in cases where they
compare equal, `min` should return the first argument and `max` the
second.  We are already doing this for our hidden friend implementation
inside `Quantity`, so now we do it for the rest of them.  This also lets
us get rid of the `using std::min` and `using std::max` declarations,
and the corresponding tests.

The reason we do this now, as part of the Eigen/vector/matrix push, is
because `using_common_type` is going to become problematic, and we would
like to simply delete it.  This lets us get rid of users that exist
outside of `quantity.hh`, which would be an awkward situation if there
are no users _inside_ of `quantity.hh`.  And it's not a hugely valuable
utility anyway; it's easy to recreate the common unit/rep logic directly
(as we now do).

To help with these implementations, we add a new utility, too.
`AppropriateCommonUnit<Q, Us...>` produces `CommonUnit<Us...>` when `Q`
is `Quantity`, and `CommonPointUnit<Us...>` when `Q` is `QuantityPoint`.
Since we didn't document the existing `AppropriateAssociatedUnit`
utility, we don't document this either.  Yes, it's in `au::`, but it's a
very "in the weeds" utility that I'm ambivalent about advertising.

Helps #484.

### DIFF
--- a/au/math.hh
+++ b/au/math.hh
@@ -38,8 +38,6 @@ using std::fmod;
 using std::hypot;
 using std::isinf;
 using std::isnan;
-using std::max;
-using std::min;
 using std::remainder;
 using std::sin;
 using std::sqrt;
@@ -355,10 +353,14 @@ AU_DEVICE_FUNC constexpr auto lerp(QuantityPoint<U1, R1> p1, QuantityPoint<U2, R
 
 namespace detail {
 // We can't use lambdas in `constexpr` contexts until C++17, so we make a manual function object.
-struct StdMaxByValue {
-    template <typename T>
-    AU_DEVICE_FUNC constexpr auto operator()(T a, T b) const {
-        return std::max(a, b);
+struct MaxByValue {
+    template <template <class, class> class Q, typename U1, typename R1, typename U2, typename R2>
+    AU_DEVICE_FUNC constexpr auto operator()(Q<U1, R1> q1, Q<U2, R2> q2) const {
+        using U = AppropriateCommonUnit<Q, U1, U2>;
+        using R = std::common_type_t<R1, R2>;
+        const auto qc1 = q1.template as<R>(U{});
+        const auto qc2 = q2.template as<R>(U{});
+        return (qc2 < qc1) ? qc1 : qc2;
     }
 };
 }  // namespace detail
@@ -368,7 +370,7 @@ struct StdMaxByValue {
 // Unlike std::max, returns by value rather than by reference, because the types might differ.
 template <typename U1, typename U2, typename R1, typename R2>
 AU_DEVICE_FUNC constexpr auto max(Quantity<U1, R1> q1, Quantity<U2, R2> q2) {
-    return detail::using_common_type(q1, q2, detail::StdMaxByValue{});
+    return detail::MaxByValue{}(q1, q2);
 }
 
 // The maximum of two point values of the same dimension.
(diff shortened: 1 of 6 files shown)


au/math.hh
```diff
@@ -355,10 +353,14 @@ AU_DEVICE_FUNC constexpr auto lerp(QuantityPoint<U1, R1> p1, QuantityPoint<U2, R
...
-    AU_DEVICE_FUNC constexpr auto operator()(T a, T b) const {
-        return std::max(a, b);
+struct MaxByValue {
+    template <template <class, class> class Q, typename U1, typename R1, typename U2, typename R2>
+    AU_DEVICE_FUNC constexpr auto operator()(Q<U1, R1> q1, Q<U2, R2> q2) const {
+        using U = AppropriateCommonUnit<Q, U1, U2>;
+        using R = std::common_type_t<R1, R2>;
+        const auto qc1 = q1.template as<R>(U{});
+        const auto qc2 = q2.template as<R>(U{});
+        return (qc2 < qc1) ? qc1 : qc2;
```
hoffbrinkle, line 363: Given this stable ordering is something we feel is a correctness issue, do we have a test for it?

chiphogg: I added a bunch of tests.  It's a very subtle issue.  I decided to create a type with a label tag that doesn't participate in ordering.  I had to add a huge number of ops just to get the type to compile as a rep, and even then, things like unit conversion didn't work.  Maybe #52 will bring more clarity here, and we can simplify the test in the future (or not).

Anyway, even though I wasn't able to cover all overloads, we've got tests now, and I think the tests are good!

Interestingly, most of them already passed, I think because our hidden friend `Quantity` implementation already had this behavior.  The failing test was one of the `QuantityPoint` ones, and it passes now.

hoffbrinkle: So I checked out some of Walter's presentations, and he suggests a common implementation for making this clear:

```c++
inline bool OutOfOrderByValue(... q1, ... q2) { return q2 < q1; }

/* min */ return OutOfOrderByValue(q1, q2) ? q2 : q1;
/* max */ return OutOfORderByValue(q1, q2) ? q1 : q2;
```

Now, I don't know where I stand on the name.  It's was a bit confusing at first, but the more I think about it and watch the talk, it's not too bad.

au/math_test.cc
```diff
@@ -544,6 +581,52 @@ TEST(Min, SupportsZeroForSecondArgument) {
     EXPECT_THAT(negative_result, SameTypeAndValue(meters(-8)));
 }
 
+// Hidden friend overload (same unit, same rep).
+TEST(Max, PrefersSecondArgForSameTypeQuantity) {
+    using LN = LabeledNumber<int>;
```
hoffbrinkle: Style guide would indicate `Ln`:

```suggestion
    using Ln = LabeledNumber<int>;
```

hoffbrinkle (approved):

au/math_test.cc
chiphogg: Nice catch!  Fixed everywhere.

au/math.hh
chiphogg: I can see both sides on this one.  I think once I really internalized the idea, "return the same place argument unless they are strictly out of order", the concise implementation just made a lot of sense to me.

Ultimately, I think we're better off without the helper.  We would need to either stick it in `"quantity.hh"` (where it feels weird), or else find or invent an upstream target for it.  I think the status quo, together with the tests, will be fine here.

hoffbrinkle: I suppose the question would be "is there anywhere else we use the `less_than` operator (regardless of spelling) to return one or the other value?"

chiphogg: Good question.  Thinking it through... I'm going to guess "no", because returning one of two values depending on how they compare is pretty much the definition of "min" or "max"?  Maybe "clamp" is covered here too.  (FWIW, I checked `clamp` as part of this work, and found that it was already compliant.)

PR #682, opened 2026-06-09
Add a `View<T>` type to wrap reps

chiphogg:
This paves the way for the `q.mutable_view()[i]` implementation in our
vector/matrix support design.  Assignment writes through to the
underlying value; other operations convert easily to the underlying
type.  Importantly, it also includes `operator[]` and `operator()`, so
that we can create a quantity whose rep is a view onto a matrix type.
This will let us edit individual matrix elements in a unit-safe way,
through the view.

Helps #484.

### DIFF
--- a/au/view_test.cc
+++ b/au/view_test.cc
@@ -74,6 +74,7 @@ TEST(View, AdditionDereferencesAndReturnsValue) {
     View<double> va{a};
     View<double> vb{b};
 
+    StaticAssertTypeEq<decltype(va + vb), double>();
     EXPECT_THAT(va + vb, Eq(4.0));
     EXPECT_THAT(va + 10.0, Eq(11.5));
     EXPECT_THAT(10.0 + va, Eq(11.5));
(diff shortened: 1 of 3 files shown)


au/view_test.cc
```diff
@@ -0,0 +1,187 @@
...
+TEST(View, AdditionDereferencesAndReturnsValue) {
+    double a = 1.5;
+    double b = 2.5;
+    View<double> va{a};
+    View<double> vb{b};
+
+    EXPECT_THAT(va + vb, Eq(4.0));
+    EXPECT_THAT(va + 10.0, Eq(11.5));
+    EXPECT_THAT(10.0 + va, Eq(11.5));
+}
```
hoffbrinkle, line 80: Is this testing that it returns a value?  If it returned a view, wouldn't the view still satisfy these tests?  Should we check that these _don't_ have the `View` trait or that `a` and `b` are untouched?  Though I suppose if we were updating the through the view one of these tests would likely fail...

hoffbrinkle (approved):

au/view_test.cc
chiphogg: If it returned a view, what would it be a view _onto_?  It would have to be one of the existing backing `double` variables, and this would change the results for the following assertions.

Still, adding a `StaticAssertTypeEq` here is cheap and easy, and clarifies design intent.  I wouldn't want to force the reader to puzzle this out for themselves every time they read it.

Or, better yet, `SameTypeAndValue` makes _all_ of them clearer, so I'll go with that.

PR #685, opened 2026-06-11
Change implicit rep to natural output type

chiphogg:
From now on, whenever we do not name the rep in a conversion, we'll
obtain whatever type we would naturally get if we wrote out the
conversion by hand.  The main place this makes a difference today is
promotable integers.  In the future, this will be critical for libraries
with expression templates, particularly Eigen.

To get the mechanics right, we need to make a bunch of updates to our
`:conversion_strategy` target.  Many of these involve explicitly
templating intermediate result types, instead of assuming they're the
same as the input type.  But the main new interface property is that _an
output rep of `void` is our signal for omitting the final cast
operation_.

Then, we take advantage of this by changing our implicit-rep conversion
functions in `Quantity` to use `void` as the rep type.  (`QuantityPoint`
follows naturally because it uses `Quantity` under the hood.)

We also need to tweak our overflow risk check in `in_impl`.  We need to
make the integer promotion carve-out explicit.

One common forced downstream change is that our conversion risk checkers
need to change to their explicit-rep versions in explicit-rep contexts
that we didn't formerly recognize as such.  The main place this crops up
is in `Constant`, where we had formerly been manifesting the constant as
`T{1}`, and then just using implicit rep to convert to the target unit.

Finally, as prep for the vector/matrix changes, I realized our ordering
in the main `PermitAsCarveOutForIntegerPromotion` implementation was not
as efficient as it could be.  I decided to move the `is_integral` checks
closer to the top, as they should be a lot more common than promotion
and unsigned checks, which don't really even make sense for non-integral
types.

Helps #484.  This PR is a major unblocker for Eigen being usable and
performant.

### DIFF
--- a/au/constant_test.cc
+++ b/au/constant_test.cc
@@ -18,6 +18,8 @@
 
 #include "au/chrono_interop.hh"
 #include "au/testing.hh"
+#include "au/units/bits.hh"
+#include "au/units/bytes.hh"
 #include "au/units/degrees.hh"
 #include "au/units/feet.hh"
 #include "au/units/inches.hh"
@@ -311,6 +313,31 @@ TEST(Constant, ImplicitlyConvertsToAppropriateQuantityTypes) {
     {
         // constexpr Quantity<decltype(Meters{} / Seconds{}), int16_t> v = c;
         // (void)c;
+    }
+}
+
+TEST(Constant, OutputHasProvidedRep) {
+    constexpr auto X = make_constant(bytes * mag<10>());
+    EXPECT_THAT(X.as<uint8_t>(bits), SameTypeAndValue(bits(uint8_t{80})));
+    EXPECT_THAT(X.as<int8_t>(bits), SameTypeAndValue(bits(int8_t{80})));
+}
+
+TEST(Constant, InOutputTypeIsProvidedRep) {
+    constexpr auto X = make_constant(bytes * mag<10>());
+    EXPECT_THAT(X.in<uint8_t>(bits), SameTypeAndValue(uint8_t{80}));
+    EXPECT_THAT(X.in<int8_t>(bits), SameTypeAndValue(int8_t{80}));
+}
+
+TEST(Constant, OverflowCheckUsesExplicitRep) {
+    // 40 bytes = 320 bits: fits in int (the promoted type) but overflows uint8_t.
+    // The explicit-rep check in Constant::as correctly catches this.
+    constexpr auto X = make_constant(bytes * mag<40>());
+    EXPECT_THAT(X.can_store_value_in<uint8_t>(bits), IsFalse());
+    EXPECT_THAT(X.can_store_value_in<uint16_t>(bits), IsTrue());
+
+    // The following must not compile.  Uncomment inside the scope to check.
+    {
+        // X.as<uint8_t>(bits);
     }
 }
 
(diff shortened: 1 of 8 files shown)


au/constant_test.cc
```diff
@@ -314,6 +316,31 @@ TEST(Constant, ImplicitlyConvertsToAppropriateQuantityTypes) {
...
+    // 40 bytes = 320 bits: fits in int (the promoted type) but overflows uint8_t.
+    // The explicit-rep check in Constant::as correctly catches this.
+    constexpr auto X = make_constant(bytes * mag<40>());
+    EXPECT_THAT(X.can_store_value_in<uint8_t>(bits), IsFalse());
+    EXPECT_THAT(X.can_store_value_in<uint16_t>(bits), IsTrue());
+
+    // The following must not compile.  Uncomment inside the scope to check.
+    {
+        // X.as<uint8_t>(bits);
+    }
```
hoffbrinkle, line 341: Not for this PR:  I feel like we have a few of these sprinkled throughout our tests now.  Should we provide some sort of list or even automated patch type script to check these as part of our release process?

au/no_wconversion_test.cc
```diff
@@ -139,4 +139,13 @@ TEST(Conversions, SupportIntMHzToU32Hz) {
     EXPECT_THAT(freq, SameTypeAndValue(hertz(40'000'000u)));
 }
 
+TEST(Quantity, CanAssignFromPromotedTypeToOriginalType) {
+    // `int8_t + int8_t` promotes to `int`, just as it would without Au.
+    // Assigning back narrows, which is the same `-Wconversion` behavior as raw integers.
+    auto x = feet(int8_t{10});
+    auto promoted = x + x;
```
hoffbrinkle, line 146: :wrench: It seems like we need a test to validate that `promoted` is what we expect prior to assigning back?

hoffbrinkle (commented):
> The main places this makes a difference are
promotable integers, and expression templates such as Eigen.

Do we just want to drop the ", and expression templates such as Eigen" in the commit message?  I'm thinking we don't really support that yet (as of the current state of the repository).  When we do, that'll just be "the" behavior.

chiphogg:
> > The main places this makes a difference are
> > promotable integers, and expression templates such as Eigen.
> 
> Do we just want to drop the ", and expression templates such as Eigen" in the commit message? I'm thinking we don't really support that yet (as of the current state of the repository). When we do, that'll just be "the" behavior.

The Eigen use case is such a critical motivator for this feature that I still wanted to mention it.  I edited it, and I hope the new wording improves on the old wording.

au/constant_test.cc
chiphogg: At this point, I'm inclined to have Claude Code just uncomment them all one at a time, build, and produce a report on what kind of error message we saw for each.

au/no_wconversion_test.cc
chiphogg: Done.

au/constant_test.cc
hoffbrinkle: LLMs aren't super reliable for things we want to repeat reliably.  What an LLM would be good at is writing the tooling to support said strategy reliably.  For instance, we could create a series of patches, with a script that applies the patch, verifies the test failure, un-applies the patch, rinse and repeat.

hoffbrinkle (approved):
> The Eigen use case is such a critical motivator for this feature that I still wanted to mention it. I edited it, and I hope the new wording improves on the old wording.

Works better for me!

au/constant_test.cc
chiphogg: Yeah, I agree.  I suppose the sweet spot would be to have Claude whip up an actual death test mechanism, that can test compile time failure, and match on expected error text.

PR #686, opened 2026-06-27
Fix compile-time factorization hitting constexpr step limit (#328)

### LINKED ISSUE #328: Improve performance for modular arithmetic
Right now, there are numbers we know we can't factor at compile time: see comments in #217.  We're using too many steps, and clang bails.  However, those comments also suggest that it should be possible to get away with far far fewer steps.  It turns out that we are spending almost all of our steps on our custom implementation of `mul_mod`.  If we can find a much more efficient implementation, we should be able to factor a much wider variety of numbers.  Leading candidate options include:

1. Implement Montgomery/REDC form
2. Do manual "double-wide" multiplication and mod

Whichever route we go, getting a more efficient `mul_mod` should also make it worthwhile to switch to a "batched" implementation of Pollard's rho, which replaces many GCD calls (slow no matter what) with modular multiplication calls (which can be made fast).  The batched implementation is a pretty marginal win if we implement it now (and is in some cases a little bit worse), but is a much bigger win if multiplication is cheap.

StefanoD:
## Summary

`mag<N>()` factors `N` at compile time, but some values could not be factored
because the work exceeded the compiler's constexpr step limit. For example (the
minimal repro from #328):

```cpp
using namespace au;
mag<9'000'000'000'000'000'500LL>();  // error: constexpr step limit exceeded
```

This PR implements the two improvements suggested in the issue — a "double-wide"
`mul_mod` and a batched Pollard's rho — which together bring the cost back well
under the default limit.

Fixes #328.

## Root cause

The example factors as `2² · 5³ · 89278723 · 201615787`. Factoring the composite
`18000000000000001` blew the step limit in two successive places:

1. **`mul_mod`** reduces products that overflow `uint64_t` by recursion. The
   recursion is bounded (`O(log b)`), but `mul_mod` is called so heavily during
   factorization that this reduction dominated the step budget.
2. Once that was addressed, **Pollard's rho** became the bottleneck: it took a
   `gcd` on every single step.

## Changes

- **`au/utility/mod.hh`** — `mul_mod` now forms the full-width product and
  reduces it in a single step using `__uint128_t` when the compiler provides it
  (`__SIZEOF_INT128__`). Compilers without a 128-bit integer type (e.g. MSVC)
  fall back to `mul_mod_via_chunking` — the original reduction, factored out
  under its own name and otherwise unchanged.
- **`au/utility/factoring.hh`** — Pollard's rho (`pollard_rho_attempt`) is
  reimplemented using the batched form of Brent's algorithm: it accumulates the
  product of position differences modulo `n` and takes a single `gcd` per
  128-step batch (with one-step backtracking when a batch overshoots). This
  trades many expensive `gcd` calls for cheap `mul_mod` calls. Its locals are
  named after their roles (`anchor`, `cursor`, `segment_length`, …).

## Performance (constexpr step cost)

As @chiphogg measured in the review — and I reproduced locally by binary-searching
clang's `-fconstexpr-steps` for the smallest budget each factorization compiles
under — the design choices bear out:

- The `__int128` path (which most users get) is roughly an order of magnitude
  cheaper than the pre-PR code.
- The portable chunking fallback plus the batched rho comes in under the pre-PR
  code on every input tried.
- An earlier revision of this PR used a repeated-*doubling* fallback instead of
  chunking. It measured several times more expensive (~3–5×) on the path where
  the overflow branch is actually taken — enough to push numbers the library
  factors today (e.g. the existing `55'245'642'489'451` test) past clang's
  default budget. The portable fallback therefore keeps the original chunking
  reduction.

## Notes on portability

On compilers with `__int128` (GCC, Clang) the example now compiles using a small
fraction of the previous step budget. On MSVC (no `__int128`) the fallback is the
original chunking reduction, so those targets are no worse than before; but
factoring numbers this large at compile time inherently needs more than MSVC's
default `/constexpr:steps` (100,000), so such cases still require raising that
limit. This matches the situation of the existing large-number factoring tests,
which are exercised on the GCC/Clang/CMake jobs (the MSVC CI builds only the
single-file package).

## Test plan

Added unit tests (fail before, pass after):

- `mod_test.cc` — `mul_mod` with two large operands (overflowing path), plus a
  direct cross-check of the portable `mul_mod_via_chunking` against `mul_mod`.
- `factoring_test.cc` — `constexpr` factorization of `18'000'000'000'000'001`
  (compile-time regression guard).
- `magnitude_test.cc` — end-to-end `mag<9'000'000'000'000'000'500>()` against its
  known factorization.

Additional local verification:

- `mul_mod` (both paths) cross-checked against an independent `__uint128_t`
  reference over millions of random inputs; `find_prime_factor` outputs verified
  to be prime and to divide the input across many random and hard cases.
- Builds clean under `-std=c++14/17/20` with `-Werror -Wall -Wextra
  -Wconversion -Wsign-conversion -Wpedantic`, and is `clang-format`-clean.

Note: This PR has been done with Claude Opus. Not sure if this work is compatible with your "Aurora Individual Contributor License Agreement". Please let me know.

### DIFF
--- a/au/utility/factoring.hh
+++ b/au/utility/factoring.hh
@@ -41,6 +41,63 @@ constexpr std::uintmax_t absolute_diff(std::uintmax_t a, std::uintmax_t b) {
     return a > b ? a - b : b - a;
 }
 
+// A single attempt at Pollard's rho, using Brent's cycle detection method, with the polynomial
+// `x^2 + t`.  Returns a nontrivial factor of `n` on success, or `n` itself if this particular
+// parameterization fails to find one (in which case the caller should retry with a different `t`).
+//
+// To keep the number of (relatively expensive) `gcd` calls small, we accumulate the product of the
+// position differences modulo `n` across a batch of steps, and take only a single `gcd` per batch.
+// This is the standard batched form of Brent's algorithm; it trades many `gcd` calls for many
+// (much cheaper) `mul_mod` calls, which is what makes it tractable at compile time.  See
+// <https://github.com/aurora-opensource/au/issues/328>.
+//
+// Precondition: `n` is known to be composite.
+constexpr std::uintmax_t pollard_rho_attempt(std::uintmax_t n, std::uintmax_t t) {
+    constexpr std::uintmax_t batch_size = 128u;
+
+    std::uintmax_t anchor = 2u;
+    std::uintmax_t cursor = 2u;
+    std::uintmax_t batch_start = 2u;
+    std::uintmax_t diff_product = 1u;
+    std::uintmax_t factor = 1u;
+    std::uintmax_t segment_length = 1u;
+
+    do {
+        anchor = cursor;
+        for (std::uintmax_t i = 0u; i < segment_length; ++i) {
+            cursor = x_squared_plus_t_mod_n(cursor, t, n);
+        }
+
+        std::uintmax_t offset_in_segment = 0u;
+        while (offset_in_segment < segment_length && factor == 1u) {
+            batch_start = cursor;
+            const std::uintmax_t remaining = segment_length - offset_in_segment;
+            const std::uintmax_t steps = (batch_size < remaining) ? batch_size : remaining;
+            for (std::uintmax_t i = 0u; i < steps; ++i) {
+                cursor = x_squared_plus_t_mod_n(cursor, t, n);
+                diff_product = mul_mod(diff_product, absolute_diff(anchor, cursor), n);
+            }
+            factor = gcd(diff_product, n);
+            offset_in_segment += batch_size;
+        }
+
+        segment_length *= 2u;
+    } while (factor == 1u);
+
+    // If the batched product happened to accumulate _all_ of `n`'s factors at once (`factor == n`),
+    // the batch hid the real factor.  Recover it by walking the same steps one at a time, taking a
+    // `gcd` after each, until we isolate a single nontrivial factor.
+    if (factor == n) {
+        factor = 1u;
+        do {
+            batch_start = x_squared_plus_t_mod_n(batch_start, t, n);
+            factor = gcd(absolute_diff(anchor, batch_start), n);
+        } while (factor == 1u);
+    }
+
+    return factor;
+}
+
 // Pollard's rho algorithm, using Brent's cycle detection method.
 //
 // Precondition: `n` is known to be composite.
@@ -50,23 +107,8 @@ constexpr std::uintmax_t find_pollard_rho_factor(std::uintmax_t n) {
     // will succeed on the first iteration, and we don't expect that any will _ever_ come anywhere
     // _near_ to hitting this limit.
     for (std::uintmax_t t = 1u; t < n / 2u; ++t) {
-        std::size_t max_cycle_length = 1u;
-        std::size_t cycle_length = 1u;
-        std::uintmax_t tortoise = 2u;
-        std::uintmax_t hare = x_squared_plus_t_mod_n(tortoise, t, n);
-
-        std::uintmax_t factor = gcd(n, absolute_diff(tortoise, hare));
-        while (factor == 1u) {
-            if (max_cycle_length == cycle_length) {
-                tortoise = hare;
-                max_cycle_length *= 2u;
-                cycle_length = 0u;
-            }
-            hare = x_squared_plus_t_mod_n(hare, t, n);
-            ++cycle_length;
-            factor = gcd(n, absolute_diff(tortoise, hare));
-        }
-        if (factor < n) {
+        const std::uintmax_t factor = pollard_rho_attempt(n, t);
+        if (factor != n) {
             return factor;
         }
     }
--- a/au/utility/mod.hh
+++ b/au/utility/mod.hh
@@ -81,11 +81,6 @@ constexpr uint64_t mul_mod_via_chunking(uint64_t a, uint64_t b, uint64_t n) {
 // Precondition: (a < n).
 // Precondition: (b < n).
 constexpr uint64_t mul_mod(uint64_t a, uint64_t b, uint64_t n) {
-    // Start by trying the simplest case, where the product "fits" in a `uint64_t`.
-    if (b == 0u || a < std::numeric_limits<uint64_t>::max() / b) {
-        return (a * b) % n;
-    }
-
 #if defined(__SIZEOF_INT128__)
     // If the compiler provides a 128-bit integer type, we can form the full-width product and
     // reduce it in a single step.  This is dramatically cheaper at compile time than the portable
--- a/au/utility/mod.hh
+++ b/au/utility/mod.hh
@@ -44,11 +44,18 @@ constexpr uint64_t sub_mod(uint64_t a, uint64_t b, uint64_t n) {
     }
 }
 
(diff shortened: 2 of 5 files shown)


chiphogg:
Using Claude Opus doesn't preclude you from signing the Individual CLA.  In general, when you prepare a PR, you use whatever tools you use; that's your business.  From our end, we just want you to personally understand and stand behind every line of code in the change before you send it to another human.  🙂

Additionally, we would be a little worried about reviewing gigantic AI-generated PRs.  But I glanced at the changes briefly, and this PR doesn't look at _all_ like that.  Instead, it looks very well scoped and nicely factored.

Thanks for making this contribution!  I hope I'll get a chance to review it soon.

au/utility/mod.hh
```diff
@@ -44,29 +44,47 @@ constexpr uint64_t sub_mod(uint64_t a, uint64_t b, uint64_t n) {
...
+#if defined(__SIZEOF_INT128__)
+    // If the compiler provides a 128-bit integer type, we can form the full-width product and
+    // reduce it in a single step.  This is dramatically cheaper at compile time than the portable
+    // fallback, which matters because `mul_mod` dominates the cost of compile-time prime
+    // factorization (see https://github.com/aurora-opensource/au/issues/328).
+    return static_cast<uint64_t>((static_cast<__uint128_t>(a) * static_cast<__uint128_t>(b)) % n);
+#else
+    // No wider integer type is available (e.g. MSVC), so fall back to the portable algorithm.
+    return mul_mod_via_doubling(a, b, n);
+#endif
```
chiphogg, line 93: Could we restore the previous recursive implementation here instead?  I had Claude dig in and measure the doubling implementation, and it turned out to be a regression whenever this branch actually gets taken.  The cost (in `constexpr` instructions) seemed to be about 3 to 5 times higher.

I think the cleanest approach would be to rename the old function to something like `mul_mod_via_chunking`: just one line changed (well, 2, for the recursion).  And then the new `mul_mod` can go just below.  That will be a diff that's easy to review.

I saw your Claude's concern about possibly unbounded recursion, but it's not accurate.  It's bounded to O(log b).

- `b` gets divided by `chunk_size` each step
- `chunk_size` starts out at least 1, and can't be only 1 for more than 1 step
- therefore, `b` gets halved each step after the first, _at least_ (in reality, it should usually be a lot faster)

au/utility/test/mod_test.cc
```diff
@@ -59,6 +60,40 @@ TEST(MulMod, HandlesHugeNumbers) {
...
+    EXPECT_THAT(mul_mod(MAX - 1u, MAX - 2u, MAX), Eq(2u));
+    EXPECT_THAT(mul_mod(MAX - 3u, MAX - 5u, MAX), Eq(15u));
+}
+
+TEST(MulModViaDoubling, HandlesSimpleCases) {
+    EXPECT_THAT(mul_mod_via_doubling(6u, 7u, 10u), Eq(2u));
+    EXPECT_THAT(mul_mod_via_doubling(13u, 11u, 50u), Eq(43u));
+}
+
+TEST(MulModViaDoubling, MatchesMulModIncludingOverflowingProducts) {
```
chiphogg: Let's definitely keep these two tests, but apply them to `MulModViaChunking` instead, once we make that the fallback.  And thanks for adding this important coverage!

au/utility/factoring.hh
```diff
@@ -41,6 +41,62 @@ constexpr std::uintmax_t absolute_diff(std::uintmax_t a, std::uintmax_t b) {
...
+    std::uintmax_t g = 1u;
+    std::uintmax_t r = 1u;
+
+    do {
+        x = y;
+        for (std::uintmax_t i = 0u; i < r; ++i) {
+            y = x_squared_plus_t_mod_n(y, t, n);
+        }
+
+        std::uintmax_t k = 0u;
```
chiphogg: I find this new version harder to understand, because the variable names are too short.  They feel cryptic.  Can we make the names similar to the old ones?  Here's a suggestion for each:

- `x` -> `anchor`
- `y` -> `cursor`

The old code used "tortoise" and "hare", but that mental model doesn't cleanly apply here, because only one of the "pointers" is actually moving.  So, "anchor" and "cursor" seem better.

- `ys` -> `batch_start`
- `q` -> `diff_product`
- `g` -> `factor`
- `r` -> `segment_length`
- `k` -> `offset_in_segment`

chiphogg (changes requested):
Thanks for your patience --- and for this contribution!  I appreciate how cleanly factored it is.  Should be good to go after some tweaks I'll outline below.

The main delay for the review was that I wanted to get a handle on the performance implications.  Looks like the `__int128` path is a huge performance win, so no surprise there.  The results for the "doubling" fallback were interesting and surprising, though:

| Input | Old code | This PR (`__int128`) | This PR (fallback) | Batched rho + old `mul_mod` |
|---|---|---|---|---|
| 55,245,642,489,451 (existing test) | 394k | **40k** | 1.52M | 268k |
| 18,000,000,000,000,001 (this PR's regression test) | 6.18M | **412k** | 20.1M | 3.8M |

(For reference: clang's default budget is 1,048,576 steps; GCC's `-fconstexpr-ops-limit` default is 33.5M.)

So, the "doubling" fallback approach was actually 3 to 5 times _worse_ than the original implementation!  In the current state, we'd actually break some numbers that the library can currently handle (at least, for users stuck on the fallback path).

Fortunately, I also found that swapping out the doubling for the original implementation --- call it "chunking" --- recovers all that speed and then some.  With your batched rho, it becomes a clean win everywhere I checked.

Finally, I filed #690 as the placeholder for the remaining work.  This PR will fix #328, delivering solid improvements across the board, and _very_ significant improvements for people who can use `__int128_t` (which I believe is most users).  However, I believe we can get even more gains with Montgomery REDC form, which I'm excited to learn about.  I have some very specific API ideas for how we could do the bookkeeping cleanly, and I'm looking forward to writing them.  But it's definitely getting into the realm of diminishing returns, and this present PR should deliver much more bang for the buck, and should be good enough for us for a while.

Thanks again!  Once you address these comments, we can get this landed.  I've moved #328 into our current release, thanks to your contribution!

StefanoD:
Tried to address your issues. You can also take over this PR when you think it makes sense. :-)

au/utility/mod.hh
```diff
@@ -69,6 +76,28 @@ constexpr uint64_t mul_mod(uint64_t a, uint64_t b, uint64_t n) {
...
+// (a * b) % n
+//
+// Precondition: (a < n).
+// Precondition: (b < n).
+constexpr uint64_t mul_mod(uint64_t a, uint64_t b, uint64_t n) {
+    // Start by trying the simplest case, where the product "fits" in a `uint64_t`.
+    if (b == 0u || a < std::numeric_limits<uint64_t>::max() / b) {
+        return (a * b) % n;
+    }
+
```
chiphogg:
```suggestion
```

I think this is totally extraneous.  The 128-bit case wouldn't need it, and the fallback case already gets it from `mul_mod_via_chunking`.

chiphogg (approved):
Changes look great!  I think there is just one duplicate chunk of code to get rid of.

chiphogg:
I've applied the tweak and merged in the latest main, so we'll see how the remote checks go.

PR #687, opened 2026-06-28
Enable using ref for unscaled quantity inputs

chiphogg:
This might be the trickiest and most subtle part of the whole
matrix/vector/Eigen support chain.  I went back and forth on the names
and shapes of the APIs many times, even though I got all of the _tests_
to pass a long time ago.

Consider adding two Eigen vectors with different units, where one of the
inputs is already equivalent to their common unit.  That input won't be
scaled.  But the current mixed-unit addition implementation (built on
`detail::using_common_type`) _unconditionally copies_ its inputs.  That
(internal!) copy is a _temporary_, and the expression template we return
holds a _pointer_ to that temporary.  Bottom line: a library-internal
lifetime bug that wasn't present in raw Eigen, and that end users can't
do anything to fix.  Ouch!

So, we need to perform the arithmetic _directly_.  And we need a _new_
helper, which passes along the reference for unscaled inputs, but does
the usual scaling (which produces expression templates) for inputs that
need scaling.  The helper follows these rules:

1. If we're not scaling (target unit is quantity-equivalent to input),
   just return a reference unconditionally (using `.data_in`).

2. If we _are_ scaling, we need to choose between implicit and explicit
   rep.

Branch (2) is where we hit the tricky and subtle parts.  How do we know
when implicit rep is OK?  And if it's not, than _which_ rep do we use
for "explicit rep"?

We can't really hope to get a fully generic answer right now, so we're
starting with a reasonable first-draft policy, and focusing on making
sure it has the right _API shape_.

- On the shape: the inputs to this sub-utility are our own rep, the
  scale factor, and the rep of the other operand.  (This is _asymmetric_
  in the two inputs, in general.)

- On the policy: we use the common rep when both inputs are arithmetic
  types, and implicit rep otherwise.

Overall, this seems pretty safe.  We're confident in reasoning about the
built-in arithmetic types, and this covers edge cases like integral to
floating point, and integer promotion.  And "implicit rep" for third
party types is a minimal and direct translation of "do what the raw code
would do".

We also fix one implementation that was failing under `-Wconversion`.
We checked all similar functions, but found that this was the only one
that was affected.

Helps #484: this prevents what would have been a terrible bug in our
(still in-progress) Eigen support.

### DIFF
--- a/au/quantity_test.cc
+++ b/au/quantity_test.cc
@@ -64,12 +64,70 @@ struct Vec {
     }
 };
 
+// Minimal expression template mock.  Arithmetic on HeapDouble returns a lightweight HeapExpr that
+// stores a pointer to the source data on the heap, deferring evaluation.  Because HeapDouble
+// heap-allocates, ASAN reliably detects use-after-free if a conversion path lets the source die
+// before the expression is evaluated --- exactly the bug that RefOrScaledCopyIn prevents.
+struct HeapExpr {
+    const double *source;
+    double scale;
+    double eval() const { return *source * scale; }
+
+    friend double operator+(HeapExpr a, HeapExpr b) { return a.eval() + b.eval(); }
+    friend double operator-(HeapExpr a, HeapExpr b) { return a.eval() - b.eval(); }
+};
+
+struct HeapDouble {
+    double *ptr;
+
+    HeapDouble() : ptr(new double(0.0)) {}
+    explicit HeapDouble(double v) : ptr(new double(v)) {}
+    HeapDouble(const HeapDouble &other) : ptr(new double(*other.ptr)) {}
+    HeapDouble(HeapDouble &&other) noexcept : ptr(other.ptr) { other.ptr = nullptr; }
+    HeapDouble &operator=(const HeapDouble &other) {
+        *ptr = *other.ptr;
+        return *this;
+    }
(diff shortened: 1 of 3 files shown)


chiphogg:
| PR | Branch | Title
| --- | --- | --- |
:arrow_right: | **`chiphogg/ref-or-scaled-copy#484`** | **Enable using ref for unscaled quantity inputs** |

au/quantity_test.cc
```diff
@@ -653,6 +719,10 @@ TEST(Quantity, IgnoringOverflowRiskCanProduceOverflow) {
                 SameTypeAndValue(milli(seconds)(uint8_t{1'000 % 256})));
 }
 
+TEST(Quantity, MixedRepMixedUnitSubtractionWorks) {
+    EXPECT_THAT(radians(1.0) - degrees(57), IsNear(degrees(0), degrees(1)));
```
hoffbrinkle, line 723: :pick: I find myself reaching for `DistanceFrom` even when `IsNear` is applicable.  Mostly because I can never remember if `IsNear` is `Lt` or `Le`. 😁

hoffbrinkle (approved):

au/quantity_test.cc
chiphogg: I somewhat optimistically tried it out just now... but I guess we don't have it yet.  Something to look forward to!

PR #689, opened 2026-07-06
Update install doc to for bzlmod

hoffbrinkle:
Now that we have Au available in the Bazel Central Registry, it's
significantly easier for users to use Au in their Bazel project.


chiphogg (approved):
I checked whether `#####` would actually render usefully.  Somewhat to my surprise, it did!

PR #691, opened 2026-07-07
Mitigate unnecessary copies in function chains

chiphogg:
Preliminary benchmarking revealed that sometimes `eval(...)` was causing
more copies than necessary.  We traced this to both the `eval`
implementation itself, and to various signature choices in our core
library machinery (similar to the recent changes to take inputs by
`const&` in many places).

To fix our machinery, we switch to using _forwarding references_ in the
`StaticCast`, `ImplicitConversion`, and `OpSequence` operations, as well
as the identity specialization for `MultiplyTypeBy`.  Then, in
`quantity.hh` and `quantity_point.hh`, we change the `make_quantity`-ish
functions to use forwarding references too, while switching several
constructors to use `const&`.  Finally, we take a previously missed
opportunity to `move` inside of the value constructor.

(We subsequently found that many of these "library entry" forwarding
references will break for bitfield inputs.  The principled fix was to
split based on value category: we use forwarding references for _rvalue
inputs only_, and take lvaues by `const&`.)

And now we are ready to fix `eval(...)`.  The subtlety here is that
Eigen returns a `const` object from `.eval()`.  The workaround is to
store that object in a non-`const` local variable, and then move the
variable into the `make_quantity` call.

Overall, these changes restored Au to performance parity with raw Eigen,
at least on the examples we have tested so far.

### DIFF
--- a/au/compatibility/materialization_copy_count_test.cc
+++ b/au/compatibility/materialization_copy_count_test.cc
@@ -1,0 +1,236 @@
+// Copyright 2026 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Deterministic copy-count gate for the materialization APIs.
+//
+// Heap-backed reps (the motivating example is `Eigen::VectorXd`) must not be silently deep-copied
+// as they flow through `make_quantity`, the `QuantityMaker`, the `Quantity` constructors, and the
+// `.as()` / `.in()` conversion chain.  Timing benchmarks can hide a stray copy behind
+// memory-bandwidth noise; exact copy counts cannot.  So here we use an *instrumented* rep that
+// counts every copy, move, and "evaluation" (materialization of a lazy expression rep), and assert
+// the theoretical minimum for each path.  Each `Tracked` copy is O(n) (it duplicates a
+// `std::vector<double>`); each move is O(1); so "copies" is the number that actually matters.
+
+#include <utility>
+#include <vector>
+
+#include "au/au.hh"
+#include "au/compatibility/eigen.hh"  // For the unit-aware `eval()` free function.
+#include "au/testing.hh"
+#include "gtest/gtest.h"
+
+namespace au {
+namespace {
+
+using ::testing::Eq;
+
+struct Meters : UnitImpl<Length> {};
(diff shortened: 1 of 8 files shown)


au/compatibility/materialization_copy_count_test.cc
```diff
@@ -0,0 +1,236 @@
...
+
+    // Element-wise add/subtract.  These exist because instantiating `Quantity<Meters, Tracked>`
+    // requires `Tracked + Tracked` and `Tracked - Tracked` to form valid types (the same
+    // requirement Au's arithmetic operators place on any rep, met by Eigen out of the box).  Each
+    // builds a fresh `Tracked`, so neither is a copy of an existing one.
+    Tracked operator+(const Tracked &other) const {
+        std::vector<double> out = data_;
+        for (std::size_t i = 0; i < out.size(); ++i) {
+            out[i] += other.data_[i];
+        }
```
hoffbrinkle: Would you want to use something like `std::transform` for this?

```suggestion
        std::vector<double> out(data_.size());
        std::transform(data_.begin(),
                       data_.end(),
                       other.data_.begin(),
                       out.begin(),
                       std::plus<double>{});
```

au/compatibility/materialization_copy_count_test.cc
```diff
@@ -0,0 +1,236 @@
...
+Tracked make_tracked(std::size_t n) {
+    std::vector<double> data(n, 1.0);
+    return Tracked{std::move(data)};
+}
+
+// (1a) `meters(v)` from an lvalue must copy exactly once (the input is not ours to steal).
+TEST(MaterializationCopyCount, MakerFromLvalueCopiesExactlyOnce) {
+    Tracked v = make_tracked(8);
+
+    reset_counts();
```
hoffbrinkle: It seems like we would want to use a test fixture for this so the tests themselves don't have to remember to do a `reset_counts`?

hoffbrinkle (approved):

au/compatibility/materialization_copy_count_test.cc
chiphogg: Great call.  Done.

au/compatibility/materialization_copy_count_test.cc
```diff
@@ -0,0 +1,237 @@
...
+Tracked make_tracked(std::size_t n) {
+    std::vector<double> data(n, 1.0);
+    return Tracked{std::move(data)};
+}
+
+// Zeroes the copy/move/eval tallies before each test, so no test has to remember to call
+// `reset_counts()` itself.
+class MaterializationCopyCount : public ::testing::Test {
+ protected:
+    void SetUp() override { reset_counts(); }
```
hoffbrinkle, line 157: Prefer to [use the constructor](https://google.github.io/googletest/faq.html#CtorVsSetUp) for things that can be done there.  It allows for inherited fixtures to be re-used without having to remember to call all the SetUp methods.

```suggestion
    MaterializationCopyCount() { reset_counts(); }
```

chiphogg: Oof, that's embarrassing!  Fixed in #693.

PR #694, opened 2026-07-08
Support decimals, scientific notation in mag UDLs

chiphogg:
Having seen more `Constant` use cases in the wild, I find they can be
pretty awkward.  We end up having to write `3_mag / 2_mag`, say, when
what we'd really like to write is `1.5_mag`.

This PR expands `_mag` UDLs to support a decimal point, as well as
optional scientific notation, using `e` or `E`.  Every such literal
always represents a _fully exact rational_ magnitude.  We also update
the docs.

### DIFF
--- a/au/magnitude.hh
+++ b/au/magnitude.hh
@@ -744,7 +744,7 @@ constexpr std::uintmax_t parse_magnitude_integer() {
 }
 
 // Count the number of mantissa digits after the decimal point in a `_mag` literal.  For example,
-// `12.34_mag` has two decimal places, while `1234_mag` and `1.234e3_mag`... the latter has three.
+// `12.34_mag` has two decimal places, while `1234_mag` has 0, and `1.234e3_mag` has three.
 template <char... Cs>
 constexpr int count_decimal_places() {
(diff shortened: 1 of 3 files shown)


au/magnitude.hh
```diff
@@ -699,36 +699,101 @@ AU_DEVICE_FUNC constexpr auto mag() {
...
+        }
         if (digits[i] >= '0' && digits[i] <= '9') {
             result = result * 10u + static_cast<std::uintmax_t>(digits[i] - '0');
         }
     }
     return result;
 }
+
+// Count the number of mantissa digits after the decimal point in a `_mag` literal.  For example,
+// `12.34_mag` has two decimal places, while `1234_mag` and `1.234e3_mag`... the latter has three.
```
hoffbrinkle: This comment reads funny to me?  Are we missing something about `1234_mag` having 0?

hoffbrinkle (approved):

au/magnitude.hh
chiphogg: Yeah, that's just weird.  Not sure what happened.  Fixed.

PR #695, opened 2026-07-09
SFINAE out impossible hidden friend conversions

chiphogg:
Surprisingly, `(v * t) + (0.5 * a * t * t)` _doesn't compile_ when `v`
and `a` are Eigen vector-backed quantities!  The problem is that both of
these are _different_ expression templates.  Right now, we directly use
`std::common_type` for these conversions, and `std::common_type` does
not have a specialization for combinations of different Eigen expression
templates.

The solution is to make the implicit constructors that these hidden
friend functions rely on SFINAE-friendly, via the conversion risk
policy. This means neither direction of hidden friend conversion will
work (which also precludes us from running into ambiguous overload
issues). Therefore, we fall back to the non-member `operator+`, which is
able to handle this kind of heterogeneous operation.

Helps #484: we will now be able to support this very basic operation.

### DIFF
--- a/au/conversion_policy.hh
+++ b/au/conversion_policy.hh
@@ -198,6 +198,7 @@ struct PassesConversionRiskCheck
           PermitAsCarveOutForIntegerPromotion<Rep, ScaleFactor, SourceRep>,
           ConversionRiskAcceptablyLow<
               ConversionForRepsAndFactor<CastStrategy, SourceRep, Rep, ScaleFactor>>> {};
+
 // If the reps have no common type, the conversion arithmetic can't even be formed, so the
 // conversion is simply not permitted.  (This keeps the check SFINAE-friendly for reps like distinct
 // Eigen expression templates, rather than hard-erroring deep inside `std::common_type`.)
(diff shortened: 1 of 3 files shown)


au/conversion_policy.hh
```diff
@@ -188,12 +188,22 @@ struct PermitAsCarveOutForIntegerPromotion
...
+          typename Rep,
+          typename ScaleFactor,
+          typename SourceRep,
+          bool = HasConversionRep<SourceRep, Rep>::value>
 struct PassesConversionRiskCheck
     : stdx::disjunction<
           PermitAsCarveOutForIntegerPromotion<Rep, ScaleFactor, SourceRep>,
           ConversionRiskAcceptablyLow<
               ConversionForRepsAndFactor<CastStrategy, SourceRep, Rep, ScaleFactor>>> {};
+// If the reps have no common type, the conversion arithmetic can't even be formed, so the
```
hoffbrinkle, line 202: Would we normally put vertical whitespace here?

```suggestion
              ConversionForRepsAndFactor<CastStrategy, SourceRep, Rep, ScaleFactor>>> {};

// If the reps have no common type, the conversion arithmetic can't even be formed, so the
```

hoffbrinkle (approved):

au/conversion_policy.hh
chiphogg: It varies.  We often "smoosh" different specializations of the same template together when they're small.  That said, in this case, I don't think it makes sense, because I didn't add an "overall" comment for `PassesConversionRiskCheck` (which I basically always do when we do this).  So I think unsmooshing is best here.  Thanks!

PR #696, opened 2026-07-09
Tweak comparison structure to accommodate Eigen

chiphogg:
Here, we really pare back the reliance on `std::common_unit` in
comparisons, reserving it for arithmetic types.  This prevents Eigen
from tripping over the `std::common_type` machinery in cases where it
doesn't exist.  We include regression test cases.  Helps #484.

### DIFF
--- a/au/compatibility/eigen_test.cc
+++ b/au/compatibility/eigen_test.cc
@@ -180,7 +180,7 @@ TEST(EigenCompatibility, EqualityOfDistinctExpressionRepsWorks) {
     // templates.  Like heterogeneous addition, this used to fail to compile inside
     // `std::common_type` (undefined for two distinct Eigen expressions).  Unlike order comparisons,
     // equality is well-defined for vectors --- Eigen's `operator==` compares all coefficients ---
-    // so it's something we should expect it to work.  This regression test makes sure that it does.
+    // so it's something we should expect to work.  This regression test makes sure that it does.
     const Eigen::Vector3d values{4.0, 5.0, 6.0};
     const auto base = meters(values);
 
--- a/au/compatibility/eigen_test.cc
+++ b/au/compatibility/eigen_test.cc
@@ -180,7 +180,7 @@ TEST(EigenCompatibility, EqualityOfDistinctExpressionRepsWorks) {
     // templates.  Like heterogeneous addition, this used to fail to compile inside
     // `std::common_type` (undefined for two distinct Eigen expressions).  Unlike order comparisons,
     // equality is well-defined for vectors --- Eigen's `operator==` compares all coefficients ---
-    // so it should work.
+    // so it's something we should expect it to work.  This regression test makes sure that it does.
     const Eigen::Vector3d values{4.0, 5.0, 6.0};
     const auto base = meters(values);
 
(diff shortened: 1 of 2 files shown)


au/compatibility/eigen_test.cc
```diff
@@ -175,6 +175,45 @@ TEST(EigenCompatibility, DifferentUnitEqualityWorks) {
     EXPECT_THAT(q_m == q_cm, IsTrue());
 }
 
+TEST(EigenCompatibility, EqualityOfDistinctExpressionRepsWorks) {
+    // `==` / `!=` between two same-unit quantities whose reps are *different* lazy Eigen expression
+    // templates.  Like heterogeneous addition, this used to fail to compile inside
+    // `std::common_type` (undefined for two distinct Eigen expressions).  Unlike order comparisons,
+    // equality is well-defined for vectors --- Eigen's `operator==` compares all coefficients ---
+    // so it should work.
```
hoffbrinkle: I think I know what you mean by "should work", but the comment may lead someone to interpret it as "we're pretty sure it should work, let's see what happens!"  Maybe we just drop the "so it should work" or change it to "so we expect that it works" (which we neatly `EXPECT` below...)?

hoffbrinkle (approved):

au/compatibility/eigen_test.cc
chiphogg: Thanks!  I didn't think about the ambiguity in the wording.  I hope the new version is clearer.

hoffbrinkle: Yep.  There's not too many places in Au where we just say YOLO!

au/compatibility/eigen_test.cc
```diff
@@ -175,6 +175,44 @@ TEST(EigenCompatibility, DifferentUnitEqualityWorks) {
     EXPECT_THAT(q_m == q_cm, IsTrue());
 }
 
+TEST(EigenCompatibility, EqualityOfDistinctExpressionRepsWorks) {
+    // `==` / `!=` between two same-unit quantities whose reps are *different* lazy Eigen expression
+    // templates.  Like heterogeneous addition, this used to fail to compile inside
+    // `std::common_type` (undefined for two distinct Eigen expressions).  Unlike order comparisons,
+    // equality is well-defined for vectors --- Eigen's `operator==` compares all coefficients ---
+    // so it's something we should expect it to work.  This regression test makes sure that it does.
```
hoffbrinkle:
```suggestion
    // so it's something we should expect to work.  This regression test makes sure that it does.
```

au/compatibility/eigen_test.cc
```diff
@@ -175,6 +175,44 @@ TEST(EigenCompatibility, DifferentUnitEqualityWorks) {
...
+    const Eigen::Vector3d values{4.0, 5.0, 6.0};
+    const auto base = meters(values);
+
+    auto twice = base * 2.0;        // (8, 10, 12) m; rep: one scalar-product node
+    auto also_twice = base + base;  // (8, 10, 12) m; rep: one sum node (different type)
+    auto thrice = base + twice;     // (12, 15, 18) m; different value *and* rep type
+    static_assert(!std::is_same<decltype(twice), decltype(also_twice)>::value, "");
```
hoffbrinkle, line 190: Ah, C++14 needs the `""`.

hoffbrinkle (approved):

PR #698, opened 2026-07-10
Add UDLs for `Constant` of each unit

chiphogg:
This piggybacks on the new `_mag` literals, and should be extremely
flexible.

There are, of course, concerns about compile time impact.  But we can
mitigate those by putting each in its own _separate_ header, which users
include only when they actually use it.

Includes doc updates.

### DIFF
--- a/au/units/literals/farads.hh
+++ b/au/units/literals/farads.hh
@@ -19,13 +19,13 @@
 #include "au/units/farads.hh"
 
 namespace au {
-namespace literals {
+namespace au_literals {
 
 // `1.28e-4_F` is a `Constant` equivalent to `make_constant(1.28e-4_mag * farads)`.
 template <char... Cs>
 constexpr auto operator""_F() {
-    return make_constant(farads * ::au::au_literals::operator""_mag < Cs... > ());
+    return make_constant(farads * operator""_mag<Cs...>());
 }
 
-}  // namespace literals
+}  // namespace au_literals
 }  // namespace au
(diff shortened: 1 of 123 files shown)


chiphogg:
| PR | Branch | Title
| --- | --- | --- |
:arrow_right: | **`chiphogg/unit_udls`** | **Add UDLs for `Constant` of each unit** |

docs/howto/new-units.md
```diff
@@ -40,6 +40,13 @@ a complete sample definition of a new Unit, with these features annotated and ex
     constexpr auto ftm = au::SymbolFor<Fathoms>{};                  //  [6]
     }
 
+    namespace literals {                                            //  [7]
```
hoffbrinkle, line 43: `literals` or `au_literals`?

au/units/literals/farads.hh
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include "au/constant.hh"
+#include "au/magnitude.hh"
+#include "au/units/farads.hh"
+
+namespace au {
+namespace au_literals {
+
+// `1.28e-4_F` is a `Constant` equivalent to `make_constant(1.28e-4_mag * farads)`.
```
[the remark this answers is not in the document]
chiphogg: The plan of record is to lean on #699 to make this true.  I will resolve all of these for that reason.

docs/howto/new-units.md
chiphogg: `literals`.  Yes, I understand the double take.  :slightly_smiling_face:  But `au_literals` is an Au library namespace, and if a user is making their own, they should follow their own conventions, which would be unlikely to produce `au_literals`.

hoffbrinkle (approved):

PR #699, opened 2026-07-10
Expand Magnitude arithmetic operations

chiphogg:
The basic idea is that everywhere we can multiply or divide by `Constant`, we should be able to multiply or divide by a `Magnitude`, and it should do the same thing as the `Constant` for the dimensionless unit with that magnitude.

We also enable Magnitude to work in unit slots.

This table summarizes the big picture of what is changing.

* = new on this branch, Y = pre-existing.

| Partner            | M*x | x*M |  M/x  | x/M |
|--------------------|-----|-----|-------|-----|
| Magnitude          |  Y  |  Y  |   Y   |  Y  |
| Zero               |  Y  |  Y  |  (1)  |  Y  |
| Constant           |  Y  |  Y  |   Y   |  Y  |
| SymbolFor          |  Y  |  Y  |   Y   |  Y  |
| Unit instance      |  *  |  Y  |   *   |  Y  |
| QuantityMaker      |  *  |  Y  |   *   |  Y  |
| QuantityPointMaker |  *  |  Y  |  (2)  |  Y  |
| SingularNameFor    |  *  |  *  |   *   |  *  |
| Quantity           |  *  |  *  | * (3) |  *  |
| Raw numeric        |  *  |  *  | * (3) |  *  |

(1) Absent by design: this would be division by zero.
(2) Deliberately omitted: the inverse of a point unit is meaningless.
(3) static_asserts when the divisor is integral (integer rep or
    integer number), matching existing `Constant` behavior.

### DIFF
--- a/au/constant_test.cc
+++ b/au/constant_test.cc
@@ -118,6 +118,16 @@ TEST(MakeConstant, MakesConstantFromSymbol) {
     EXPECT_THAT(123 * ad_hoc_c, QuantityEquivalent(123 * c));
 }
 
+TEST(MakeConstant, MakesDimensionlessConstantFromMagnitude) {
+    constexpr auto three = make_constant(mag<3>());
+    StaticAssertTypeEq<decltype(make_constant(mag<3>())),
+                       Constant<decltype(UnitProduct<>{} * mag<3>())>>();
+
+    // A bare magnitude in numeric arithmetic acts like its corresponding constant.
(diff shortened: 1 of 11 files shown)


hoffbrinkle (approved):
Funny, I actually tried to use `/ mag<4>()` a few days ago without knowing about this, so it would seem this is a natural extension.

PR #702, opened 2026-07-16
Support version definition macros

### LINKED ISSUE #667: Add version macro defines
The library uses no macros at all, which is great! The _complete_ lack of them does present an issue, though; if you have users that might want to detect if the library having been included, and issue a warning or `#error` if it has not been, this isn't currently possible. (This isn't hypothetical; I have a real use-case for this).

There's also no way to programmatically detect the library version; very helpful for migration scenarios.

A great solution to both problems is to have a version macro somewhere, which is a popular/standard practice across many open-source projects:

- fmt: [`FMT_VERSION`](https://github.com/fmtlib/fmt/blob/4968433a6abecbbf7e72e97188926586a7fd0662/include/fmt/base.h#L23-L24)
- nlohmann::json: [`NLOHMANN_JSON_VERSION_*`](https://github.com/nlohmann/json/blob/eba0a92bfb1abf242f30e894b96ba11c9f4048c7/include/nlohmann/detail/abi_macros.hpp#L21-L23)
- toml++: [`TOML_LIB_*`](https://github.com/marzer/tomlplusplus/blob/6626578220391cb1da6f79f9a4f12066c79f2d26/include/toml%2B%2B/impl/version.hpp#L7-L9)

chiphogg:
Add version-detection macros for Au in a new `version.hh` header, included
everywhere via `config.hh`:

- `AU_VERSION_MAJOR` / `AU_VERSION_MINOR` / `AU_VERSION_PATCH` — the components.
- `AU_VERSION` — a single monotonically increasing integer for ordinary `<`/`>=`
comparisons.
- `AU_VERSION_NUMBER(major, minor, patch)` — build a comparable value for an
arbitrary version (each component < 1000).

`version.hh` is the single source of truth for the library version:
`CMakeLists.txt` parses the component macros to populate `project(... VERSION
...)`, so the two build systems can't drift.  `RELEASE.md` is updated to cover
bumping them.

These macros are a contract for _tagged releases_ only.  On a release,
`AU_VERSION` names exactly that release's feature set, so version comparisons
are sound release to release.  On `main` they name the _most recent release_, so
**behavior should not be version-gated against a `main` checkout**: the number
lags the changes actually landed there, and the two directions fail differently
(additive `>=` under-reports safely; breaking `<` can silently pick old behavior
on a commit that already has the new one).  For robust detection of a _specific_
change, we would have to introduce a dedicated per-feature macro introduced in
the same commit as the change, not `AU_VERSION`, and I don't want to do that
without a concrte use case.

`main` reports the last release and is bumped only when the next release is cut
(including patches, which `main` already contains since they're cherry-picks
from it).  We deliberately do *not* follow the common "bump to the next dev
version right after a release" convention, for two reasons: (1) it forces
guessing the next version number before it's known, and (2) other projects make
it safe with a pre-release marker (`0.6.0-dev`) that our integer-only macros
can't express: without it, bumping ahead would make additive checks fail
_unsafe_.  Bump-right-before avoids both these problems.

We considered adding some kind of other macro to indicate whether we're on a
"real" release vs. `main` --- say, `AU_VERSION_IS_RELEASE` --- but we rejected
it because that is not how people commonly use release version macros, and the
added complexity raises corner cases and makes it easier to use incorrectly.

Fixes #667.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,8 +38,9 @@ set(_au_version_patch "${CMAKE_MATCH_1}")
 
 # Fail loudly if any component failed to parse, rather than silently building a malformed version
 # such as "0.5." (which would otherwise flow into `project()`, the docs URL, and installed configs).
-foreach(_part major minor patch)
-   if(NOT _au_version_${_part} MATCHES "^[0-9]+$")
+foreach(_part MAJOR MINOR PATCH)
+   string(TOLOWER "${_part}" _part_lower)
+   if(NOT _au_version_${_part_lower} MATCHES "^[0-9]+$")
       message(FATAL_ERROR
          "Could not parse AU_VERSION_${_part} from au/version.hh. "
          "Expected a line like '#define AU_VERSION_${_part} <number>'.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,11 +24,35 @@
 # [1]: https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html
 cmake_minimum_required(VERSION 3.27...3.29)
 
+# Derive the project version from `au/version.hh`, which is the single source of truth for the
+# library version (it also defines the `AU_VERSION*` macros for C++ code).  This keeps the CMake
+# version and the C++ macros from ever drifting apart, and means a release only needs to bump the
+# version in one place.
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/au/version.hh" _au_version_hh)
+string(REGEX MATCH "#define AU_VERSION_MAJOR ([0-9]+)" _ "${_au_version_hh}")
+set(_au_version_major "${CMAKE_MATCH_1}")
+string(REGEX MATCH "#define AU_VERSION_MINOR ([0-9]+)" _ "${_au_version_hh}")
+set(_au_version_minor "${CMAKE_MATCH_1}")
+string(REGEX MATCH "#define AU_VERSION_PATCH ([0-9]+)" _ "${_au_version_hh}")
+set(_au_version_patch "${CMAKE_MATCH_1}")
+
+# Fail loudly if any component failed to parse, rather than silently building a malformed version
+# such as "0.5." (which would otherwise flow into `project()`, the docs URL, and installed configs).
+foreach(_part major minor patch)
+   if(NOT _au_version_${_part} MATCHES "^[0-9]+$")
+      message(FATAL_ERROR
+         "Could not parse AU_VERSION_${_part} from au/version.hh. "
+         "Expected a line like '#define AU_VERSION_${_part} <number>'.")
+   endif()
+endforeach()
+
+set(AU_VERSION "${_au_version_major}.${_au_version_minor}.${_au_version_patch}")
+
 project(
    Au
-   VERSION 0.5.0
+   VERSION ${AU_VERSION}
    DESCRIPTION "A C++ quantities and units library, by Aurora Innovation"
-   HOMEPAGE_URL "https://aurora-opensource.github.io/au/0.5.0/"
+   HOMEPAGE_URL "https://aurora-opensource.github.io/au/${AU_VERSION}/"
    LANGUAGES CXX
 )
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -71,16 +71,29 @@ Any empty section can be omitted.
 We try to follow [semantic versioning](https://semver.org/).  Since we are currently in major
 version zero (0.y.z), incompatible changes don't force a major version upgrade.
 
-### Update the CMake version number
+### Update the version number
 
-Edit the `CMakeLists.txt` file in the root folder, updating the version number in the `project`
-command to the number chosen above.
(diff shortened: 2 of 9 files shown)


RELEASE.md
```diff
@@ -172,7 +185,19 @@ https://aurora-opensource.github.io/au/0.3.1/...
                                        ^^^^^
 ```
 
-We do this in a PR on the release branch in order to avoid churn commits on the main branch.
+In this same PR, flip the release flag in `au/version.hh`:
+
+```
+Find this:      #define AU_VERSION_IS_RELEASE 0
```
marzer: For my purposes at least this part would be overkill, since I'm using the tags/hashes of my dependencies' releases anyways, but I suppose others might find it helpful.

marzer (approved):
Thanks for this! 🙏

RELEASE.md
chiphogg: Thanks!  I'm pretty new to this use case; maybe it shows.  😅  I added this because it felt weird to have the version macros say that we're some specific version, when we're not --- and when we could easily have breaking changes relative to that version.  But maybe I'm overthinking it.

In any case, it seems harmless enough (just a little extra work for us at release time), so I'll probably leave it in.

docs/reference/version.md
```diff
@@ -0,0 +1,63 @@
+# Version macros
+
+Au provides preprocessor macros that let downstream code detect that the library is present, and
+which version of it is present.  They live in [`"au/version.hh"`](../../au/version.hh), which every
```
geoffviola: This doesn't seem to work. I think you need a full path.
```
WARNING -  Doc file 'reference/version.md' contains a link '../../au/version.hh', but the target '../au/version.hh' is not found among documentation files.
```

CMakeLists.txt
```diff
@@ -24,11 +24,35 @@
 # [1]: https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html
 cmake_minimum_required(VERSION 3.27...3.29)
 
+# Derive the project version from `au/version.hh`, which is the single source of truth for the
```
geoffviola, line 27: I usually see folks do this the other way where the version is in the CMake file and the header gets [generated from a template](https://cmake.org/cmake/help/v3.18/guide/tutorial/#adding-a-version-number-and-configured-header-file). With some more work, you can run git to determine `AU_VERSION_IS_RELEASE`.

au/version_test.cc
```diff
@@ -0,0 +1,72 @@
...
+#ifndef AU_VERSION_MAJOR
+    FAIL() << "AU_VERSION_MAJOR is not defined";
+#endif
+#ifndef AU_VERSION_MINOR
+    FAIL() << "AU_VERSION_MINOR is not defined";
+#endif
+#ifndef AU_VERSION_PATCH
+    FAIL() << "AU_VERSION_PATCH is not defined";
+#endif
+    // Each component must fit in the three decimal digits that `AU_VERSION_NUMBER` allots it.
```
geoffviola, line 37: Should it include MAJOR?

CMakeLists.txt
```diff
@@ -24,11 +24,35 @@
...
+set(_au_version_major "${CMAKE_MATCH_1}")
+string(REGEX MATCH "#define AU_VERSION_MINOR ([0-9]+)" _ "${_au_version_hh}")
+set(_au_version_minor "${CMAKE_MATCH_1}")
+string(REGEX MATCH "#define AU_VERSION_PATCH ([0-9]+)" _ "${_au_version_hh}")
+set(_au_version_patch "${CMAKE_MATCH_1}")
+
+# Fail loudly if any component failed to parse, rather than silently building a malformed version
+# such as "0.5." (which would otherwise flow into `project()`, the docs URL, and installed configs).
+foreach(_part major minor patch)
+   if(NOT _au_version_${_part} MATCHES "^[0-9]+$")
```
geoffviola: This should make the error message nicer.
```suggestion
foreach(_part MAJOR MINOR PATCH)
   string(TOLOWER "${_part}" _part_lower)
   if(NOT _au_version_${_part_lower} MATCHES "^[0-9]+$")
```

geoffviola (approved):
This should work. We may want to revisit the process to make it more automated.

docs/reference/version.md
chiphogg: Fixed.

CMakeLists.txt
chiphogg: Interesting idea.  We could try doing this later, although I'm hesitant to avoid make CMake a dependency for the main library, since we're bazel-first.

au/version_test.cc
chiphogg: Yep, fixed!

CMakeLists.txt
chiphogg: Done.  Thanks!

chiphogg:
I'm going to land this now, but I wanted to note @geoffviola that I did make some substantive changes since your review.  If you're fine with them, you don't need to do anything.  If there are any changes you think we should make, write back here, and then we can make a follow-on PR.

PR #703, opened 2026-07-17
Use the same clang-format in CI and locally

chiphogg:
When we upgraded to clang-format 17, we forgot about the GitHub check,
and left it at clang-format 14.  We _could_ simply bump the version
number, but that means the same problem will just recur the next time we
upgrade our toolchain.  So instead, we drop the pre-created action, and
replace it with a direct command that uses `tools/bin/clang-format`.

Unfortunately, clang-format has its issues.  There is a bug that was
introduced between versions 14 and 17, which formats variadic UDLs badly
(see llvm/llvm-project#210135).  Therefore, we retain the old, good,
formatting for all of these, leaving an indicator as to which bug must
be fixed before we can migrate.

This unblocks #702.  Once this lands, we can merge in the latest main,
and take it out of draft.

### DIFF
--- a/.github/workflows/clang-format-lint.yml
+++ b/.github/workflows/clang-format-lint.yml
@@ -26,9 +26,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b #v4.1.5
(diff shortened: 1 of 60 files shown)


geoffviola (approved):
Seems like a reasonable workaround. It's a lot of comments to remove a space, but the source does look slightly better.

timhirsh (approved):

PR #708, opened 2026-08-06
Check risks by default in explicit-rep overloads

### LINKED ISSUE #122: Consider separating "forcing" from explicit-Rep

chiphogg:
Instead of `ignore(ALL_RISKS)`, these functions now get
`check_for(ALL_RISKS)`.  Some other implementations that didn't have a
policy at all now get an explicit `ignore(ALL_RISKS)`: particularly, the
`coerce` functions that we are about to deprecate.

The constructor change isn't strictly necessary, but it's a slight
improvement.  The `EnableIfImplicitOkIs` template parameter really is
the source of truth for whether this constructor should be allowed.
Using `ignore(ALL_RISKS)` instead of `check_for(ALL_RISKS)` removes the
possibility for the gates to disagree if we change it in the future, and
also avoids re-doing redundant compile time checks that we already
performed.

And, of course, we update various unit tests that are affected by this
interface change, and also update the docs.  Fixes #122.

### DIFF
--- a/au/quantity.hh
+++ b/au/quantity.hh
@@ -167,8 +167,9 @@ class Quantity {
               typename Enable = EnableIfImplicitOkIs<true, OtherUnit, OtherRep>>
     AU_DEVICE_FUNC constexpr Quantity(
         const Quantity<OtherUnit, OtherRep> &other)  // NOLINT(runtime/explicit)
-        : value_{other.template in_impl<detail::UseImplicitConversion, Rep>(
-              UnitT{}, check_for(ALL_RISKS))} {}
+        // `ignore(ALL_RISKS)` because we already determined that this implicit conversion is OK.
+        : value_{other.template in_impl<detail::UseImplicitConversion, Rep>(UnitT{},
+                                                                            ignore(ALL_RISKS))} {}
 
     // EXPLICIT constructor for another Quantity of the same Dimension.
     template <typename OtherUnit,
@@ -212,7 +213,7 @@ class Quantity {
     // `q.as<Rep>(new_unit)`, or `q.as<Rep>(new_unit, risk_policy)`
     template <typename NewRep,
               typename NewUnitSlot,
-              typename RiskPolicyT = decltype(ignore(ALL_RISKS)),
+              typename RiskPolicyT = decltype(check_for(ALL_RISKS)),
               std::enable_if_t<!IsConversionRiskPolicy<NewUnitSlot>::value, int> = 0>
     AU_DEVICE_FUNC constexpr auto as(NewUnitSlot u, RiskPolicyT policy = RiskPolicyT{}) const {
         return make_quantity<AssociatedUnit<NewUnitSlot>>(
@@ -229,7 +230,7 @@ class Quantity {
     // `q.in<Rep>(new_unit)`, or `q.in<Rep>(new_unit, risk_policy)`
     template <typename NewRep,
               typename NewUnitSlot,
-              typename RiskPolicyT = decltype(ignore(ALL_RISKS))>
+              typename RiskPolicyT = decltype(check_for(ALL_RISKS))>
     AU_DEVICE_FUNC constexpr auto in(NewUnitSlot u, RiskPolicyT policy = RiskPolicyT{}) const {
         return in_impl<detail::UseStaticCast, NewRep>(u, policy);
     }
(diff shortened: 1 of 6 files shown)


au/quantity.hh
```diff
@@ -755,7 +756,7 @@ struct AreQuantityTypesEquivalent<Quantity<U1, R1>, Quantity<U2, R2>>
 // Cast Quantity to a different underlying type.
 template <typename NewRep, typename Unit, typename Rep>
 AU_DEVICE_FUNC constexpr auto rep_cast(Quantity<Unit, Rep> q) {
-    return q.template as<NewRep>(Unit{});
+    return q.template as<NewRep>(Unit{}, ignore(ALL_RISKS));
```
hoffbrinkle, line 759: So this would be the argument to prefer `.as<Rep>()` vs `rep_cast<Rep>()`.  Do we need to bring more attention to this in the docs with a note or warning designation or something?

au/quantity_test.cc
```diff
@@ -737,7 +739,7 @@ TEST(Quantity, AsCanExplicitlyOptOutOfOverflowRiskCheck) {
 }
 
 TEST(Quantity, AsCanExplicitlyOptOutOfOverflowRiskCheckForExplicitRep) {
-    // In future versions (see #122), this would fail to compile without `ignore(OVERFLOW_RISK)`.
+    // This line would fail to compile without the `ignore(OVERFLOW_RISK)` argument.
```
hoffbrinkle: Do we need these comments at all now?  I think the test case (with `Explicitly`) covers us?

docs/reference/quantity.md
```diff
@@ -814,7 +780,7 @@ the denominator.
 The `divide_using_common_unit()` utility takes two `Quantity` inputs, `a`, and `b`, whose dimension is
 the same.  It first converts each input to their [common
 unit](../discussion/concepts/common_unit.md), and then performs regular division.  The result
-will be a dimensionless and unitless number (or, after [#185], a `Quantity`).
+will be a dimensionless and unitless `Quantity`.
```
hoffbrinkle, line 783: Did we just miss this with #185?

au/quantity.hh
chiphogg: IMO, it's already pretty well reflected in the docs ([example](https://aurora-opensource.github.io/au/main/reference/quantity/#rep_cast)), but you can let me know if you think it needs more emphasis.

docs/reference/quantity.md
chiphogg: Yep. 😅  Figured I'd fix my oversight here.

au/quantity_test.cc
chiphogg: Good catch.  Relying on the test case name is better.  I also updated the cases that _didn't_ have explicit rep, just because the test case name argument applies there too, and it's better to be consistent.

au/quantity.hh
hoffbrinkle: Sorry, what I was wondering about was `!!! note` or `!!! warning` on that last line.

We now have significantly fewer APIs that do things without any safety checks.  Is this the only one?  I'm thinking for the ones that we have left, we may want to call out a bit more?

chiphogg: Done.  Here's a before:

image

...and after:

image

Off the top of my head, I couldn't think of any other APIs that force through risky conversions without a risk policy parameter.

hoffbrinkle: Yeah, I couldn't think of any either, which definitely makes the `rep_cast` API an outlier...  Which makes me wonder how long we should keep it around...

hoffbrinkle (approved):

au/quantity.hh
chiphogg: I get the sentiment, but I feel it maps very cleanly onto `static_cast`.  I feel like it _actually is_ the kind of API that I merely _mistook_ our explicit-rep overloads for.  When I reach for it, it feels very natural.

hoffbrinkle: For sure, but even `static_cast` is something I try to avoid as much as possible.  I can see it still being used, and I've had a use case for it recently.  I can't remember what the use case was just now.  It might be the case that `.as<Rep>()` would've worked in that situation.  I'll keep an eye out for it in case it's something I haven't landed yet.

Issue #481, opened 2025-07-23
Deprecate all "coerce" APIs


chiphogg:
In 0.6.0, we should deprecate all "coerce" APIs, assuming our experience with the new #387 APIs is positive.  We'll plan to remove them in 0.7.0.

PR #710, opened 2026-08-07
Remove all library-internal uses of coerce APIs

chiphogg:
This will pave the way for #481.

### DIFF
--- a/fuzz/quantity_runtime_conversion_checkers_test.cc
+++ b/fuzz/quantity_runtime_conversion_checkers_test.cc
@@ -101,7 +101,7 @@ TYPED_TEST_P(QuantityRuntimeConversionChecker, RoundTripIsIdentityIffConversionN
     for (auto _ = 1'000'000u; --_;) {
         const auto value = meters(generator.next_value());
 
-        const bool expect_loss = is_conversion_lossy<Rep>(value, destination_unit);
+        const bool expect_loss = is_conversion_lossy(value, destination_unit);
 
         const auto round_trip =
             value.as(destination_unit, ignore(ALL_RISKS)).as(meters, ignore(ALL_RISKS));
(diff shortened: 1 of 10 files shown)


fuzz/quantity_runtime_conversion_checkers_test.cc
```diff
@@ -101,9 +101,10 @@ TYPED_TEST_P(QuantityRuntimeConversionChecker, RoundTripIsIdentityIffConversionN
     for (auto _ = 1'000'000u; --_;) {
         const auto value = meters(generator.next_value());
 
-        const bool expect_loss = is_conversion_lossy<Rep>(value, destination_unit);
+        const bool expect_loss = is_conversion_lossy(value, destination_unit);
```
hoffbrinkle: Just curious about this change?

chiphogg: I thought that was weird too, but I went with it because it was clearly more correct, since it fixed an implicit/explicit rep mismatch in the APIs we were using.  But it still bugged me subconsciously, so I'm glad you raised it.  I dug in and got to the bottom of it.

It turns out that we never updated the `coerce` functions to stop using explicit rep.  This would have been a merely aesthetic problem, except for our other recent change where "implicit rep" no longer _always_ means "same rep".  So in the cases where implicit rep changes the rep, `coerce_foo(x)` would not be exactly equivalent to `foo(x, ignore(ALL_RISKS))`.

I ended up deciding to revert the various commits in this PR that were related to this issue, and putting that fix in its own new PR, #711.  The reverts are local for now, but I'll merge in the new main once #711 lands and then push them up.  It should make this PR more single-purpose and easier to understand.

(Also, just for reference, in #711, I fixed the mismatch in this _particular_ line in the _other_ direction: I kept the explicit rep, but made the other conversions explicit rep too.  That is more consistent with the goals of the fuzz helpers.)

hoffbrinkle (approved):

Issue #712, opened 2026-08-10
Provide mechanism to request "same rep" for unit conversions


chiphogg:
Implicit rep used to mean "same rep", but doesn't anymore.  We should provide a mechanism for users who explicitly want to use the same rep at every step.  This will preserve the value of our overflow safety surface for promotable integer types.  It will also give a convenient option for Eigen users to force materializing unit conversion results.

PR #713, opened 2026-08-10
Update most docs in light of Eigen/vec/mat changes

chiphogg:
This includes several brand new files to cover Eigen (a how-to guide, a
reference page, and a safety discussion page).  We also update all the
remaining doc pages to refer to the new vector and matrix capabilities
where applicable, as well as other changes (implicit rep no longer same
rep, min/max consistently use corrected semantics everywhere) that we
hadn't yet updated.

The main things we're leaving for future PRs are:

- Overflow and truncation risk updates (the former is blocked on the new
  issue #712)
- An optional "library design discussion page" for Eigen, which explains
  the core differences between Au and other libraries

Helps #70.


docs/reference/eigen.md
```diff
@@ -0,0 +1,466 @@
...
+
+template <typename V, typename U, typename R>
+auto cwiseProduct(const V &a, const Quantity<U, R> &b);
+```
+
+!!! warning "Lifetime risk"
+    This operation is **lazy**: its result is an expression template, which refers to its operands
+    instead of owning its own copy of the data.  The result is valid only as long as its operands
+    are alive.  If it needs to outlive them, materialize it with [`eval()`](#eval).  See our
+    [Eigen safety guide](../discussion/concepts/eigen_safety.md) to learn more.
```
hoffbrinkle: Is there any way to insert a boilerplate text snippet like this with our doc tool?

chiphogg: This turned out to be not too much effort, so I factored these out.  We could consider using this same technique for other pre-existing repetitive doc sessions later on.

doc_snippets/eigen-lifetime-risk-lazy.md
hoffbrinkle, line 1: Should this be `eigen-lifetime-risk-lazy.md`?  Not sure if expect another library to have lifetime risks, but the snipped does reference Eigen docs.

hoffbrinkle (approved):

doc_snippets/eigen-lifetime-risk-lazy.md
chiphogg: Makes sense.  Done.

hoffbrinkle (approved):

PR #716, opened 2026-08-10
Formally deprecate coerce functions

### LINKED ISSUE #481: Deprecate all "coerce" APIs

chiphogg:
We provide a deprecation message that links to the migration guide in
the 0.5.1 docs.  (We will update this to the 0.6.0 version, which will
include this PR's tweaks, as part of the release process.)

Then we update all doc references to the coerce functions.

Fixes #481.


docs/reference/quantity.md
```diff
@@ -364,16 +364,11 @@ To be concrete, here are the signatures of the functions that support the policy
...
-    In the meantime, here is how you convert.
+    These functions are deprecated.  Here is how you convert them.
 
     First, figure out which conversion risks you are trying to override: **overflow**, or
-    **truncation**, or **both**.  (If you don't have an explicit `<Rep>` argument, you can simply
-    delete the `"coerce_"` word and compile, and the error message will tell you which one is
-    relevant.  Otherwise, you will need to use your knowledge of the types and units involved to
-    figure this out.)
+    **truncation**, or **both**.  (You can simply delete the `"coerce_"` word and compile, and the
+    error message will tell you which one is relevant.)
```
hoffbrinkle: 💬 

So someone doesn't clean up a `coerce` that wasn't needed in the first place and isn't surprised when there are no compiler errors...

```suggestion
    error message will tell you which one, if any, is relevant.)
```

We could also expound on the "if any" if we think that's needed?

chiphogg: I think "if any" suffices.  Good catch!

docs/reference/quantity_point.md
```diff
@@ -287,16 +287,11 @@ To be concrete, here are the signatures of the functions that support the policy
...
-    In the meantime, here is how you convert.
+    These functions are deprecated.  Here is how you convert them.
 
     First, figure out which conversion risks you are trying to override: **overflow**, or
-    **truncation**, or **both**.  (If you don't have an explicit `<Rep>` argument, you can simply
-    delete the `"coerce_"` word and compile, and the error message will tell you which one is
-    relevant.  Otherwise, you will need to use your knowledge of the types and units involved to
-    figure this out.)
+    **truncation**, or **both**.  (You can simply delete the `"coerce_"` word and compile, and the
+    error message will tell you which one is relevant.)
```
hoffbrinkle:
```suggestion
    error message will tell you which one, if any, is relevant.)
```

hoffbrinkle (approved):

PR #721, opened 2026-08-12
Update overflow and trunctation docs post-Eigen

### LINKED ISSUE #70: Support matrices, vectors, and linear algebra

chiphogg:
These needed to wait a bit more than the other docs, because the process
of writing them uncovered the need for `SameRep` and `cast<T>`.

Fixes #70.

### DIFF
--- a/docs/discussion/concepts/overflow.md
+++ b/docs/discussion/concepts/overflow.md
@@ -130,6 +130,59 @@ ingredient that lets Au users use a wide variety of integral types with confiden
 
 ![The overflow safety surface](../../assets/overflow-safety-surface.png)
 
+#### Integer promotion pitfalls {#integer-promotion}
+
+There's a key subtlety to appreciate here: the safety surface applies to the _destination_ rep.  For
+"explicit rep" callsites (e.g., `q.as<T>(unit)`), the destination rep is plain to see (it's `T`).
+But "implicit rep" callsites (e.g., `q.as(unit)`) require more care.
+
+Through Au 0.5.0, the destination rep for an implicit-rep callsite was just the same as the _input_
+rep: if `q` has type `Quantity<U, R>`, it would be `R`.  But we realized this choice was suboptimal
+overall: instead, Au should do as C++ does, and return the natural type.  [0.6.0] will be the first
+release that does so.  While it's a better state for the library overall, it does have some
+surprising implications.
+
+As a concrete example, consider `feet(int8_t{20}).as(inches)`.  This conversion multiplies the
+underlying `int8_t` value by 12.  In C++, any arithmetic on an `int8_t` first promotes it to `int`
+(this is [integer promotion]), so the result is **an `int`**, not an `int8_t`.  Therefore, we get
+the wider overflow safety surface for `int`, not the tiny one for `int8_t`, and the conversion is
+permitted.
+
+The conversion itself is perfectly safe, but the real danger comes if you assign back to `int8_t`:
+
+```cpp
+Quantity<Feet, int8_t> length_ft = feet(20);
+
+// Risky conversion: not caught by overflow safety surface!
+Quantity<Inches, int8_t> length_in = length_ft.as(inches);
+```
+
+In the second statement, `length_ft.as(inches)` produces a `Quantity<Inches, int>` --- note: `int`,
+not `int8_t`.  This can _implicitly_ convert to `Quantity<Inches, int8_t>`, but only because `int`
+is the _specific_ type that `int8_t` promotes to.  (If we disallowed this implicit conversion, then
+integer promotion would make small integer types completely unusable in practice.)  And so the
+overflow lands silently: 20 feet is 240 inches, which doesn't fit in an `int8_t`, and `length_in`
+ends up holding **-16 inches**.
+
+All these decisions --- returning the promoted type, permitting implicit conversion to the original
+type --- are reasonable individually, but taken together, they effectively bypass Au's overflow
+protection for small integer types.  Fortunately, there's a solution: _name the desired rep when you
+do the conversion_.  For small integer types, this should almost always be the same rep, so we
+provide a convenient syntax ([SameRep]) to ask for this.  Continuing with our earlier example:
+
+```cpp
+// Risky conversion caught when output rep specified:
+Quantity<Inches, int8_t> length_in = length_ft.as<SameRep>(inches);
+//      Names the rep at the point of conversion ^^^^^^^^^
+```
+
+The `<SameRep>` tag lets the conversion machinery know how much room we actually have to work with,
+and empowers Au to provide the full safety check.
+
+Users working with small integer types must already wrestle with their quirky properties in C++.
+We recommend these users make this technique a habit whenever they want to stay within the type in
+Au operations.
+
 ### Check every conversion at runtime {#check-at-runtime}
 
 While the overflow safety surface is a leap forward in safety and flexibility, it's still only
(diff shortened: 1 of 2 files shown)


docs/discussion/concepts/overflow.md
```diff
@@ -130,6 +130,59 @@ ingredient that lets Au users use a wide variety of integral types with confiden
...
+type --- are reasonable individually, but taken together, they effectively bypass Au's overflow
+protection for small integer types.  Fortunately, there's a solution: _name the desired rep when you
+do the conversion_.  For small integer types, this should almost always be the same rep, so we
+provide a convenient syntax ([SameRep]) to ask for this.  Continuing with our earlier example:
+
+```cpp
+// Risky conversion caught when output rep specified:
+Quantity<Inches, int8_t> length_in = length_ft.as<SameRep>(inches);
+//      Names the rep at the point of conversion ^^^^^^^^^
+```
```
hoffbrinkle, line 177: So this is an interesting example.  If I coding this, I'd naturally go for:

```c++
Quantity<Inches, int8_t> length_in = length_ft.as<int8_t>(inches);
```

What I'm really concerned about is that my result fits in the destination type.

Actually, I'd likely use `auto` here, as I've already specified what I want on the right hand side, so what I'm getting on the left (in the `auto`) is pretty straight-forward to deduce:

```c++
auto length_in = length_ft.as<int8_t>(inches);
```

All that to say, I'm not sure this is a great example of when `SameRep` would be used...  I could maybe see it with:

```c++
auto length_in length_ft.as<SameRep>(inches);
```

docs/discussion/concepts/truncation.md
```diff
@@ -73,11 +73,23 @@ point representation is a statement, by the user, that _exact values do not matt
 application.  It would be inappropriate for us to raise warnings about perfectly routine properties
 of the user's chosen type.  Hence: _floating point never truncates_.
 
+### Compound types
+
+Not every rep is a plain arithmetic type.  Complex numbers, and Eigen vectors and matrices, are
```
hoffbrinkle:
```suggestion
Not every rep is a plain arithmetic type.  Complex numbers and Eigen vectors and matrices, are
```

I think the `Eigen` types are a single item in this list?  With a third item, it would be:

```
Complex numbers, Eigen vectors and matrices, and Some Other Fun Library Type 
are common examples ...
```

docs/discussion/concepts/truncation.md
```diff
@@ -73,11 +73,23 @@ point representation is a statement, by the user, that _exact values do not matt
...
+common examples of _compound_ reps: types built up out of some underlying _scalar_.  Au
+characterizes each such rep by that scalar --- its
```
hoffbrinkle: Oh, an em-dash!

docs/discussion/concepts/truncation.md
```diff
@@ -107,10 +119,23 @@ point as a **non-truncating** operation.
...
 
 Here, too, our support for non-arithmetic rep types is limited (see [#52]), and we take
-a conservative approach.  Any cast involving a non-arithmetic type, either as source or destination,
-is considered to have truncation risk, and will not be allowed by default: users must pass
-`ignore(TRUNCATION_RISK)` as a second argument to override this.  We hope to have better default
-behavior once we support non-arithmetic types more fully.
+a conservative approach.  Any cast involving a rep with non-arithmetic scalar type, either as source
+or destination, is considered to have truncation risk, and will not be allowed by default: users
+must pass `ignore(TRUNCATION_RISK)` as a second argument to override this.  We hope to have better
+default behavior once we support non-arithmetic types more fully.
```
hoffbrinkle: Do we want to reference #52 here as well?

```suggestion
default behavior once we support non-arithmetic types more fully ([#52]).

File: docs/discussion/concepts/overflow.md (hunk 1 above)
Comment (chiphogg) [reply docs/discussion/concepts/overflow.md:177]: Great callout.  I really wrestled with what I was trying to say here.

I even realized there are actually three categories.

1. Explicit rep
2. `SameRep` (as source)
3. Same rep _as target_

After reading your example, I had started working on an example where we assign to a field in a struct: it seemed better motivated and easier to understand than manually naming a type like `Quantity<Inches, int8_t>`.  That was when I realized that neither explicit rep nor `SameRep` gives us what we want.  The most ergonomic solution I could think of was something like:

```cpp
convert_into(make_in_out(data.field), q, target_units);
```

Obviously, `make_in_out` isn't something we could ship in Au; that uses Aurora-internal utilities.  And I wouldn't want to just take a bare reference, because the callsite readability is poor!  So _actually adding_ a feature like this is not a clear slam dunk win.  Moreover, it's very easy to write this utility in a client codebase without changing Au at all.

So Au natively supports the first two categories, and it's up to end users to support the third --- definitely today, possibly forever.

And I didn't want to muddle the docs with _even more_ nuance.

So where does that leave us?

I ended up liking the `auto` in your final example the best.  The `auto` actually pairs really nicely with the `SameRep`: it's a clear expression of author intent, "stay within the numeric type".  I added a brief sentence after just saying that we dropped the explicit typename because it's now clearly determined by the RHS.  As the above novel ^ indicates, yes, there is still a lot of subtlety lurking around these issues, but I'm hard pressed to improve this example any further at the moment.

docs/discussion/concepts/truncation.md
chiphogg: I ended up doing "vectors/matrices", which avoids the confusion you pointed out here.

docs/discussion/concepts/truncation.md
chiphogg: Yep --- I explicitly configured the site to replace `---` with actual em-dashes in the rendered output.  We were liberally using em-dashes long before the LLM era! 🤓

docs/discussion/concepts/truncation.md
chiphogg: Done.  In fact, I think it's better to reference it here only.  The other one felt a little redundant, and the link makes more sense here.

hoffbrinkle (approved):

PR #722, opened 2026-08-13
Add rep-named aliases for 8 and 16 bit integers

chiphogg:
Pet peeve: it always feels like these should exist, but they don't.
This fills that gap.


hoffbrinkle (approved):
> Pet peeve: it always feels like these should exist, but they don't.

I, too, have long wondered about this.

Issue #529, opened 2025-08-28
Add code examples to README and doc main page


chiphogg:
For each of these, it'd be good to have one concise example, and link to a bunch more.

Honestly, we should have an "examples" section in our doc website, so we can go into more depth about each of them.

Many thanks to reddit users Arghnews and serendib for [making this suggestion](https://www.reddit.com/r/cpp/comments/1n2elqv/comment/nb5st8l/)!

PR #726, opened 2026-08-18
Add basic machinery for code examples

chiphogg:
We add an `examples/` folder, where we can store small, complete,
runnable programs that show how migrating to Au will change code.  We
also have CI jobs that verify that both the raw and Au versions produce
identical output.

Each example gets _directly inlined_ via snippets into its corresponding
doc page, which goes in the new "examples" section in our docs.  This
has a nice, "blinkable" A/B comparison, and an extended discussion
section, with highly relevant links at the bottom to learn more.  The
A/B tabs reuse Au's canonical colorblind-friendly scheme, found most
prominently on our alternatives page.

Stylistic presentation was a critical decision point.  I decided to
write up a new discussion doc with a set of recommended style points, so
that we could both apply them to these examples, and recommend them for
real projects to follow.  This unlocks a collapsible "front matter"
section in the A/B comparison, which gets includes and `using`
declarations out of the way of the core code, while still making them
easily accessible so that users can get a runnable example.

For now, we just add a single example, which is enough to get the
machinery up and running.  I chose the angular/linear velocity example,
as a good well rounded example.  We also add it in the two places users
would most expect to see an example: our main `README.md`, and our doc
website's main page.  Both of them link to the full version of _this_
example, and to the examples section generally.  And they both get CI
coverage to prevent them from drifting away from the example targets,
which we check on every commit.

Helps #529.  We'll add more examples before the release, but at least
this gives us a place to put them!


docs/index.md
```diff
@@ -7,6 +7,41 @@ developer experience.
...
+    ```
+
+The `2π` and the `60` vanish: in their place, Au _automatically_ generates the [_single_
+number](./discussion/implementation/applying_magnitudes/#summary-and-conclusion) that represents the
+overall conversion --- _at compile time_, so there's no runtime penalty.  And the code now directly
+states the physical truth: converting between linear and angular speed is just applying
+a length/angle ratio, and `rad / r` --- "one radian per radius" --- is nothing more than the
+_definition_ of a radian, applied as a formula.
+<!-- The last sentence above, in particular, sounds very "Claude-y" to me.  But I actually wrote
+this prose myself, from scratch!  I guess I have a robotic personality.  Beep boop. -->
```
chiphogg, line 40: I was leaning towards keeping this amusing comment in as an Easter Egg, but I could easily be persuaded to just take it out. 😅

README.md
```diff
@@ -44,6 +44,54 @@ website](https://aurora-opensource.github.io/au).
...
+// The types state the units.  Nothing to remember; nothing to convert.
+QuantityF<Rpm> wheel_rpm(QuantityF<MetersPerSecond> v, QuantityF<Meters> r) {
+    return v * rad / r;
+}
+```
+<!-- END EXAMPLE -->
+
+So: what changed?
+
+- 🛡️ Both the inputs and outputs are **robustly safe**: callers can only pass a speed for the first
```
hoffbrinkle: Do we use icons like this elsewhere?  I personally have an aversion to them as an indication of "likely to be LLM written, I probably don't want to read it."  (Right up there with "why this matters...")

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+|--------|----------|
+| `"au/io.hh"` | `operator<<`, for printing quantities to a stream |
+| `"au/std_format.hh"` | `std::format` support |
+| `"au/testing.hh"` | Utilities for writing googletest tests |
+
+This is the same principle as above, applied to _features_ rather than units.  `"au/io.hh"` is the
+one you'll meet first, and it makes the point well: it pulls in the C++ streams machinery, which is
+famously expensive, and which many embedded projects avoid entirely.  Charging that cost to every
+Au user ---
+including those who never print a quantity --- would be a poor trade.  So we don't: you include it
```
hoffbrinkle: Odd line break?  Wouldn't be in the generated docs, but odd in the unformatted docs.

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+1. The prefix may be _small_, but Au is designed for _composability_, so it can be repeated many
+   times.  Compare `speed.in(au::kilo(au::meters) / au::hour)` to `speed.in(kilo(meters) / hour)`:
+   the latter has notably better readability and flow.
+
+2. The cost is small: it only affects one place in the file, and the location is predictable.
+
+3. The more times you use a name, the more the `using` declaration pays off.
+
+What you will _not_ find in our examples is `using namespace au;`.  We don't recommend it, so we
+don't model it:[^1] `au::` is deliberately two characters, and a directive that drags in every unit,
```
hoffbrinkle, line 80: Do we want to make an analogy (either here or in the footnote) to `using namespace std;`?

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+
+[^1]: We _do_ use `using namespace au;` in the canonical "quickstart" [godbolt link] from our
+README, because the goal of that page is to make it as frictionless as possible to write Au code.
+It's not intended to showcase _landable_ code; it's intended to make it as easy as possible to see
+how the library behaves in some situation, or to use it to get some result.  To further illustrate
+this point, note that our godbolt link also includes every unit, constant, and literal in the whole
+library, even though we would _never_ recommend this in production code, because the compile time
+cost doesn't scale.
+
+[godbolt link]: https://godbolt.org/z/o91nfh5Wc
```
hoffbrinkle, line 93: We already have a godbolt reference in README.md.  Does it make sense to share that reference or just point to the main readme here?

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+The core reason is that `#include` is simple textual inclusion in C++, and it works transitively.
+This means any names we expose in a header file will _silently leak_ into every translation unit
+that includes it: an unbounded set that grows over time.  This problem is known as "namespace
+pollution", and experience has shown that it causes a litany of problems.
+
+- Dependencies become entangled and hard to extract.
+
+- It becomes hard to understand where names come from.
+
+- The build can even break due to name collisions.
```
hoffbrinkle: The "even" reads funny to me?

```suggestion
- The build can break due to name collisions.
```

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+```cpp
+QuantityF<decltype(Kilo<Meters>{} / Hours{})> speed;
+
+QuantityF<UnitQuotient<Kilo<Meters>, Hours>> speed;
+```
+
+We can make this easier to read by just defining a _type alias_ for the unit:
+
+```cpp
+using KiloMetersPerHour = UnitQuotient<Kilo<Meters>, Hours>;
```
hoffbrinkle, line 149: So I tend to go with the `decltype`

```c++
using KiloMetersPerHour = decltype(Kilo<Meters>{} / Hours{});
```

The operator being easier than using the named implementation and the extra `<>`.

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+A type alias is a different thing from a `using` _declaration_, and the rules are correspondingly
+looser: it introduces one name you chose, rather than importing a set of names you did not.  That
+makes it fine at namespace scope even in a header.  That said, we recommend the same instincts as
+above:
+
+- If you're in an implementation file (`.cc`, `.cpp`, etc.), use them without worrying about it.
+
+- If you're in a header file (`.hh`, `.hpp`, etc.), a unit alias at namespace scope is safe, but
+  keep in mind that it _is_ still a name your consumers will see.  Give it a descriptive name, or
+  hide it inside a function or class.
```
hoffbrinkle: The Google Styleguide words this pretty well:

> Don't put an alias in your public API just to save typing in the implementation; do so only if you intend it to be used by your clients.

I'm thinking part of our wording should note that `using` in the header becomes part of your public API?

docs/discussion/idioms/namespaces-and-includes.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Summary
+
+Every project has its own style, conventions, and rules.  We've provided the conventions here as
+a good _default_ style for using Au, and explained the reasons that we recommend them.  These aren't
+hard and fast rules that we intend to force on all projects; they simply convey our vision for how
+to use the library ergonomically.  We hope users will find them useful.
```
hoffbrinkle, line 184: Should we make some mention that these guidelines are based on experience with large C++ projects, but are also useful for small projects, which often have a habit of become larger projects than originally envisioned?

docs/index.md
hoffbrinkle: I don't think Claude could've written anything this elegant!  Let's keep the comment...

examples/BUILD.bazel
```diff
@@ -0,0 +1,55 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load(":defs.bzl", "ab_example")
+
+package(default_visibility = ["//visibility:public"])
```
hoffbrinkle: What other build packages use these targets?  I'm assuming, for example, our source code packages don't consume these?

examples/BUILD.bazel
```diff
@@ -0,0 +1,55 @@
...
+    "check_ab_example.sh",
+    "check_readme_snippets.sh",
+])
+
+# The doc website inlines these sources directly (via `pymdownx.snippets`), so that every code
+# example on the site is real source that CI compiles and runs.
+filegroup(
+    name = "doc_sources",
+    srcs = glob([
+        "**/*.cc",
```
hoffbrinkle: Recursive glob is strongly discouraged.  They break bazel's incrementality engine and result in longer builds.

examples/BUILD.bazel
```diff
@@ -0,0 +1,55 @@
...
+    expected_output = "409.256 rev / min\n",
+)
+
+# The doc website inlines example source directly, so it cannot go stale.  The README has to keep a
+# copy instead, because GitHub Markdown has no include mechanism -- so check the copy against the
+# source it came from.
+sh_test(
+    name = "readme_snippets_test",
+    srcs = ["check_readme_snippets.sh"],
+    args = ["$(location //:README.md)"],
```
hoffbrinkle: I think this should be `rootpath`...

```suggestion
    args = ["$(rootpath //:README.md)"],
```

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+Note what invariant 2 is *not*: it is not "pad both files to equal length."  Align at structural
+landmarks (includes, signature, the key computation, the closing brace) and let genuine differences
+show.  Hiding a real difference under padding would defeat the example.
+
+**Padding at the very top or bottom of a region does nothing.**  The renderer drops leading and
+trailing blank lines from a code block, so blanks there buy no height on the page.  Put padding
+between two non-blank lines, where it survives --- or restructure so it isn't needed, which is why
+`angular_velocity` keeps its unit aliases in `frontmatter` rather than at the head of the region.
+`check_ab_example.sh` counts the way the renderer does, so it will catch this; the earlier version
+counted raw source lines and called a visibly broken page aligned.
```
hoffbrinkle: The earlier version?  I'm assuming this was earlier in development, so probably irrelevant here.

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Namespaces, and the `frontmatter` region
+
+Every file here carries a `frontmatter` region --- includes, plus the `using` declarations ---
+marked the same way as `example`, and every doc page shows it in a collapsed "Includes and usings"
+block inside each tab.  Add one to any new example.  Collapsing it is what makes it safe to put in
+both tabs: collapsed blocks are the same height in each, so the alignment of the code below
+survives.
+
+The rule the frontmatter exists to make visible:
```
hoffbrinkle: This sentence reads funny, not sure what exactly it's supposed to be.  What rule are we talking about?

Oh, I think I see it.  Is it something like "the frontmatter exists such that we can follow our style guidance in these examples without distracting from the example itself?"

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+  rather than quietly making both sides print something neutral.
+
+- **Don't reach for `setprecision`.**  The test only requires that the two sides agree with each
+  other and with `expected_output`.  Default stream precision is fine; if the printed digits
+  change, update `expected_output` rather than contorting the example to preserve an old string.
+  Do check that the two sides still agree bit-for-bit, since Au composes conversion factors
+  differently from hand-written arithmetic and the last digit can move.
+
+Printing pulls in `<iostream>`, so the Au target needs `//au:io`.  Many embedded projects avoid
+`<iostream>` for code size; nothing in Au depends on it, and `adc_millivolts` --- the embedded
```
hoffbrinkle, line 109: I assume the embedded example is coming?

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+moves when you flip.
+
+**Line numbers off.**  A 1-digit versus 2-digit gutter would shift the columns horizontally.
+
+## Verifying
+
+The tests cover output and alignment:
+
+```sh
+bazel test //examples/...:all
```
hoffbrinkle: FYI, these are equivalent:

```suggestion
bazel test //examples/...
```

I usually save `:all` for the case where you want all targets in a package without recursion: `//examples:all`.

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+CI builds `//...:all` with `-Werror`, so these examples are compiled by the clang11/14/17,
+gcc12/15, C++20, asan, and ubsan jobs.  All were verified clean.  The MSVC and CUDA jobs do not
```
hoffbrinkle: Were or are?

```suggestion
gcc12/15, C++20, asan, and ubsan jobs.  All are verified clean.  The MSVC and CUDA jobs do not
```

examples/angular_velocity/au.cc
```diff
@@ -0,0 +1,72 @@
...
+// limitations under the License.
+
+// NOTE TO EDITORS: this file is line-aligned with `raw.cc`.  See the note in that file.
+//
+// The unparenthesized `v * rad / r` below is deliberate.  Writing `v * (rad / r)` asks for the
+// ratio as a value, and since `rad` carries no number, that means materializing the reciprocal
+// `1 / r`: a division *and* a multiplication, where `v * rad / r` costs only the division.  The doc
+// page claims two floating point instructions against the raw version's three.  Parenthesizing
+// makes that claim false while leaving the output, the line count, and every test unchanged --- so
+// nothing but this note will catch it.
```
hoffbrinkle, line 22: Do we have any designs for additionally showing the generated object output side by side?  It might be useful in a few examples (I've got at least one in mind) to highligth the ability of the compiler to optimize when using Au.

examples/angular_velocity/au.cc
```diff
@@ -0,0 +1,72 @@
...
+// The types state the units.  Nothing to remember; nothing to convert.
+QuantityF<Rpm> wheel_rpm(QuantityF<MetersPerSecond> v, QuantityF<Meters> r) {
+    return v * rad / r;
+}
+// --8<-- [end:headline]
+
+int main() {
+
+    const auto omega = wheel_rpm((meters / second)(15.0f), milli(meters)(350.0f));
+    std::cout << omega.as(revolutions / minute) << std::endl;
```
hoffbrinkle: No performance gain here, but a [stylistic suggestion](https://clang.llvm.org/extra/clang-tidy/checks/performance/avoid-endl.html):

```suggestion
    std::cout << omega.as(revolutions / minute) << '\n';
```

examples/check_ab_example.sh
```diff
@@ -0,0 +1,81 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Checks that an A/B example pair stays trustworthy.  See `ab_example.bzl` for the rationale.
+#
+# Usage: check_ab_example.sh RAW_BIN AU_BIN EXPECTED_TXT RAW_SRC AU_SRC
+
+set -eu
```
hoffbrinkle: Do you want `pipefail`?

```suggestion
set -euo pipefail
```

Might have to use `/bin/bash` to get that support...

examples/check_readme_snippets.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+#     <!-- BEGIN EXAMPLE: examples/angular_velocity/au.cc:headline -->
+#     ```cpp
+#     ...copied code...
+#     ```
+#     <!-- END EXAMPLE -->
+#
+# Usage: check_readme_snippets.sh README.md
+
+set -eu
```
hoffbrinkle:
```suggestion
set -euo pipefail
```

examples/defs.bzl
```diff
@@ -0,0 +1,86 @@
...
+
+    sh_test(
+        name = "{}_test".format(name),
+        srcs = ["check_ab_example.sh"],
+        args = [
+            "$(location :{}_raw)".format(name),
+            "$(location :{}_au)".format(name),
+            "$(location :{}_expected.txt)".format(name),
+            "$(location {})".format(raw_srcs[0]),
+            "$(location {})".format(au_srcs[0]),
```
hoffbrinkle: I think these are `rootpath`.  Though I see there is some conflict in the Bazel documentation.  I'll see if I can't run down the conflict...

```suggestion
            "$(rootpath :{}_raw)".format(name),
            "$(rootpath :{}_au)".format(name),
            "$(rootpath :{}_expected.txt)".format(name),
            "$(rootpath {})".format(raw_srcs[0]),
            "$(rootpath {})".format(au_srcs[0]),
```

docs/discussion/idioms/namespaces-and-includes.md
chiphogg: I'm of two minds.  I'm obviously worried about the link going stale, but I also didn't want to force people to jump through an extra hoop just to follow the link.  Ultimately, I figured I'd be likely to grep the codebase for `godbolt` when updating it, so I went with the direct approach.

docs/discussion/idioms/namespaces-and-includes.md
chiphogg: Fixed.

docs/discussion/idioms/namespaces-and-includes.md
chiphogg: Good call.  Done.

docs/discussion/idioms/namespaces-and-includes.md
chiphogg: Added this as an option.  It gives a nice symmetry to the above section.

docs/discussion/idioms/namespaces-and-includes.md
chiphogg: This is much clearer.  Thanks!  Done.

docs/discussion/idioms/namespaces-and-includes.md
chiphogg: Added a paragraph at the end.

examples/angular_velocity/au.cc
chiphogg: I would love to see this as future work.  There probably would be a way, although it's definitely going to have to wait until after 0.6.0.

By the way: I have a (currently private) GitHub repo, "The C++ Units Library Ecosystem".  It's going to be a website someday, where units library authors can add implementations for a variety of examples using their units library, idiomatically.  I'm even planning to support different releases of the same library so we can see evolution over time.  And it automatically measures and compares compile time.

Anyway, one feature it has is that it compares the object code for the raw implementation and each units library's implementation.  If they're identical, we show a checkmark.  I hadn't thought of ways to show where we're _better_ than raw, but that would be a great expansion.

examples/BUILD.bazel
chiphogg: Yep, definitely not.  I restricted the visibility to the one package that needs this, and I put it on the one target here that is needed.

examples/BUILD.bazel
chiphogg: Claude was a little dubious on this reasoning ("glob results are tracked"), but thought a better reason was to avoid silently pulling in any stray `.cc` files that aren't examples.  I myself don't understand bazel's handling of globs at all, so I can't comment on the first reason, but the second makes sense to me.  Anyway, the glob is gone!

examples/README.md
chiphogg: Yep, that's right.  "an earlier dev version" should preserve the meaning without misleading.

examples/README.md
chiphogg: Yeah, that reads better.

examples/README.md
chiphogg: Yep!  😅  I left the reference in because I didn't want to have to remember to update this.  (This file is quasi-private anyway.)

README.md
chiphogg: The funny thing is: I typed these out by hand, to make it stand out more!  Learning the Ubuntu emoji keyboard has been life changing.  But... you're right.  This does make it look more LLM-y.  I'm fine with deleting them to give off better anti-slop vibes.

examples/angular_velocity/au.cc
hoffbrinkle: I'm guessing in most cases, one could write it raw to match, but would be unlikely to do so since it would be "clever" code.

examples/BUILD.bazel
hoffbrinkle: > I myself don't understand bazel's handling of globs at all

If you want a less than 10 minute crash course from 2 years ago: https://bazelcon2024.sched.com/event/1h6SI/how-bazel-handles-globs-nathan-harmata-google

examples/angular_velocity/raw.cc
```diff
@@ -0,0 +1,45 @@
...
+// Speed must be m/s.  Radius must be meters.  Returns RPM.
+float wheel_rpm(float v_mps, float r_m) {
+    return v_mps / (2.0f * static_cast<float>(M_PI) * r_m) * 60.0f;
+}
+// --8<-- [end:headline]
+
+int main() {
+    // The wheel radius is 350 mm, so convert it to meters by hand first.
+    const float omega_rpm = wheel_rpm(15.0f, 350.0f / 1000.0f);
+    std::cout << omega_rpm << " rev / min" << std::endl;  // Unit label typed by hand.
```
hoffbrinkle:
```suggestion
    std::cout << omega_rpm << " rev / min" << '\n';  // Unit label typed by hand.
```

hoffbrinkle (approved):

PR #729, opened 2026-08-20
Respect guarded includes in single-file script

chiphogg:
Consider something like this, from `au/operators.hh`:

```cpp
#if defined(__cpp_impl_three_way_comparison) && __cpp_impl_three_way_comparison >= 201907L
#include <compare>
#endif
```

Right now, the single-file script pulls out the include line
unconditionally, which causes a build error on compilers that don't know
about that file.

This change stops us from pulling out those includes.  They will
naturally occur in their current home (i.e., at the top of a file's
section, somewhere in the middle of the single-file include).  This
unblocks us from checking GCC 5.3 for the Au 0.6.0 release.

### DIFF
--- a/tools/bin/make-single-file
+++ b/tools/bin/make-single-file
@@ -210,6 +210,7 @@ class SourceFile:
         self.lines = []
 
         in_copyright_header = False
+        preproc_depth = 0
         with open(filename) as f:
             was_last_line_blank = False
             for line in f:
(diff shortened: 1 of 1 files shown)


tools/bin/make-single-file
```diff
@@ -232,16 +233,25 @@ class SourceFile:
...
-                    if target.startswith("au"):
-                        self.graph_includes.append(target)
-                    else:
-                        self.global_includes.append(line.rstrip())
+                if target and target.startswith("au"):
+                    self.graph_includes.append(target)
+                elif target and preproc_depth == 0:
+                    self.global_includes.append(line.rstrip())
                 else:
                     self.lines.append(line.rstrip())
```
hoffbrinkle, line 256: I don't necessarily recommend this, but in general I have an aversion to else (for me, it adds complexity, but that doesn't seem to be a universal sentiment).

```suggestion
                if target:
                    if target.startswith("au"):
                        self.graph_includes.append(target)
                        continue
                    
                    if preproc_depth == 0:
                        self.global_includes.append(line.rstrip())
                        continue
                    
                self.lines.append(line.rstrip())
```

tools/bin/make-single-file
```diff
@@ -232,16 +233,25 @@ class SourceFile:
                         continue
                     was_last_line_blank = True
 
+                # Track `#if`/`#endif` nesting depth, so that we can tell
+                # whether an `#include` is guarded.
+                if line.startswith("#if"):
+                    preproc_depth += 1
+                elif line.startswith("#endif"):
```
hoffbrinkle: I guess similar, the `else` isn't technically required (I tend to let optimizers handle things like efficiency).

```suggestion
                if line.startswith("#endif"):
```

hoffbrinkle (approved):

tools/bin/make-single-file
chiphogg: I didn't like the `[el]if target and` repetition here, but I decided to live with it on the first posting, as it seemed like a wash to me overall.

As for this specific snippet, I think leaning on `continue` is a slight step back in understandability, so I'll pass on this for now.  This also reduces the delta between what we land and what was already tested.  I do agree this is ripe for some later polishing/improvement though.

hoffbrinkle: Agreed that `continue` can be a discussion in terms of complexity/readability/maintainability.  I've long used preferred myself, so I'm comfortable with the idiom (way back in C with single return functions, goto error handling, etc.).

For me going to C++ with the ability easily support multiple return style programming has led me to view a `continue` more as a "`return` from this loop iteration" if that makes any sense.  So in that context, it makes it even easier for me to use now.  That being said, as soon as I see a continue, I can be done with that context and can read on to see what else is being done/skipped.

Issue #391, opened 2025-04-22
Improve error messages when a constructed Quantity truncates


hoffbrinkle:
Given some units defined as such[^1]:

```c++
struct PerMille : decltype(Percent{} / mag<10>());
struct PerMyriad : decltype(PerMille{} / mag<10>());
struct PerCentMille : decltype(PerMyriad{} / mag<10>());
struct PartsPerMillion : decltype(PerCentMille{} / mag<10>());
struct PartsPerBillion : decltype(PartsPerMillion{} / mag<1'000>());
struct PartsPerTrillion : decltype(PartsPerBillion{} / mag<1'000>());
struct PartsPerQuadrillion : decltype(PartsPerTrillion{} / mag<1'000>());
struct PartsPerQuintillion : decltype(PartsPerQuadrillion{} / mag<1'000>());
```

[^1]: https://godbolt.org/z/GjG38c7Yc

Trying to assign from a `QuantityI64<PartsPerQuintillion>` to a `QuantityI64<PartsPerBillion>` results in an error message that says there isn't a constructor available:

```console
note: candidate function (the implicit move assignment operator) not viable: no known conversion from 'Quantity<av::embedded::firmware::protocol::ptp::PartsPerQuintillion, [...]>' to 'Quantity<av::embedded::firmware::protocol::ptp::PartsPerBillion, [...]>' for 1st argument
```

Manually requesting the conversion using `.as(PartsPerBillion{})` yields a more useful error:

```console
error: static assertion failed due to requirement 'IMPLICIT_OK': Dangerous conversion for integer Rep!  See: https://aurora-opensource.github.io/au/main/troubleshooting/#dangerous-conversion
```

Which can then be resolved with a `.coerce_as`.  It would be useful to get the conversion error in the first instance.

chiphogg:
One thing I was thinking about here: if we simplify this and just delegate to `.as`, then it will affect the results of `std::is_constructible` for various `Quantity` types.  The better error messages are nice, but I'm not sure we want to give up the ability to query these traits.  Something to think about.

chiphogg:
I tried simplifying the constructors and running our tests, and uncovered some concrete use cases that depend on the status quo.

- Basically all of `chrono_policy_validation_test.cc` uses this to test that we support/omit the same operations as the chrono library.
[...]

PR #732, opened 2026-08-20
Update troubleshooting guide for Au 0.6.0

chiphogg:
This step is extremely manual and tedious, so I just had Claude do it,
and spot checked the results to make sure they look reasonable.

One of our errors changed enough to justify a new section title, so we
made the change, and retained the old error message in a bullet point.

Claude was able to substantially improve the ergonomics of the MSVC
updates, too.  Formerly, I had to push a separate commit for every
individual block.  Claude wrote a script (`split-error-examples`,
included here) that enabled it to get the results in a single push.
We'll use this approach from now on.

Unblocks the 0.6.0 release.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -82,50 +82,47 @@ It's the "classic" error the units library aims to prevent.
     au/error_examples.cc:34:17: error: calling a private constructor of class 'au::Quantity<au::Seconds, double>'
         set_timeout(0.5);
                     ^
-    ./au/quantity.hh:483:15: note: declared private here
-        constexpr Quantity(Rep value) : value_{value} {}
-                  ^
+    ./au/quantity.hh:678:30: note: declared private here
+        AU_DEVICE_FUNC constexpr Quantity(Rep value) : value_{std::move(value)} {}
+                                 ^
     au/error_examples.cc:37:33: error: calling a private constructor of class 'au::Quantity<au::Meters, double>'
         constexpr QuantityD<Meters> length{5.5};
                                     ^
-    ./au/quantity.hh:483:15: note: declared private here
-        constexpr Quantity(Rep value) : value_{value} {}
-                  ^
+    ./au/quantity.hh:678:30: note: declared private here
+        AU_DEVICE_FUNC constexpr Quantity(Rep value) : value_{std::move(value)} {}
+                                 ^
     ```
 
-    **Compiler error (gcc 10)**
+    **Compiler error (gcc 12)**
     ```
     au/error_examples.cc: In function 'void au::example_private_constructor()':
-    au/error_examples.cc:34:20: error: 'constexpr au::Quantity<UnitT, RepT>::Quantity(au::Quantity<UnitT, RepT>::Rep) [with UnitT = au::Seconds; RepT = double; au::Quantity<UnitT, RepT>::Rep = double]' is private within this context
+    au/error_examples.cc:34:16: error: 'constexpr au::Quantity<UnitT, RepT>::Quantity(Rep) [with UnitT = au::Seconds; RepT = double; Rep = double]' is private within this context
        34 |     set_timeout(0.5);
-          |                    ^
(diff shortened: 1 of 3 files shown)


docs/troubleshooting.md
```diff
@@ -301,91 +303,103 @@ you the risk set for your conversion, so you don't have to guess.
...
             ]
     ```
 
-## No type named 'type' in 'std::common_type'
+## Can only compute ratio of same-dimension units
+
+**Other variants:**
+
+- "No type named 'type' in 'std::common_type'" (this was the main form of this error in Au 0.5.0 and
+  earlier, and it can still show up as a follow-on error)
```
hoffbrinkle, line 402: FYI, I trip over this one a bunch when I use `.as()` with a unit that doesn't work.

Somewhat related, there a bunch of situations where the library won't do a conversion automatically.  I _think_ this happens when I've got some Quantity that would trip a risk check if it were converted, so the `.as()` has to be done manually.  I'll keep an eye out for this one, or maybe try to reproduce on Godbolt.  But it comes up often for me.

tools/bin/split-error-examples
```diff
@@ -0,0 +1,64 @@
...
+import os
+import re
+
+BAR = "/" * 100
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--input", default=os.path.join("au", "error_examples.cc"))
+    parser.add_argument("--outdir", default="sections")
```
hoffbrinkle: Suggest using `import pathlib`, much nicer ("pythonic") file experience...

```suggestion
    parser.add_argument("--input", type=pathlib.Path, default=pathlib.Path("au") / "error_examples.cc")
    parser.add_argument("--outdir", type=pathlib.Path, default=pathlib.Path("sections"))
```

tools/bin/split-error-examples
```diff
@@ -0,0 +1,64 @@
...
+    parser.add_argument("--single-file", action="store_true")
+    args = parser.parse_args()
+
+    lines = open(args.input).read().split("\n")
```
hoffbrinkle:
```suggestion
    lines = args.input.read_text().split("\n")
```

tools/bin/split-error-examples
```diff
@@ -0,0 +1,64 @@
...
+        for i in includes:
+            lines[i] = ""
+        lines[includes[0]] = '#include "au.hh"'
+
+    starts = [i for i, l in enumerate(lines) if l.startswith(BAR)]
+    names = [lines[i + 1].replace("// SECTION: ", "") for i in starts]
+    end = next(i for i, l in enumerate(lines) if l.startswith("}  // namespace au"))
+    bounds = [(starts[k], starts[k + 1] if k + 1 < len(starts) else end) for k in range(len(starts))]
+
+    os.makedirs(args.outdir, exist_ok=True)
```
hoffbrinkle:
```suggestion
    args.outdir.mkdir(exist_ok=True)
```

tools/bin/split-error-examples
```diff
@@ -0,0 +1,64 @@
...
+    for k, (name, _) in enumerate(zip(names, bounds)):
+        out = list(lines)
+        for j, (c, d) in enumerate(bounds):
+            if j != k:
+                for i in range(c, d):
+                    out[i] = ""
+        slug = re.sub(r"[^a-z0-9]+", "_", name.lower()).strip("_")
+        path = os.path.join(args.outdir, f"{k:02d}_{slug}.cc")
+        open(path, "w").write("\n".join(out))
```
hoffbrinkle:
```suggestion
        outfile = args.outdir / f"{k:02d}_{slug}.cc"
        outfile.write_text("\n".join(out))
```

hoffbrinkle (approved):

docs/troubleshooting.md
chiphogg: Thanks!  Do keep an eye out.  BTW, is this the same thing as #391 (which you filed a while back), or something else?

PR #735, opened 2026-08-20
Add ADC code example

chiphogg:
This gives us our first explicitly embedded use case, and shows off our
integer arithmetic handling.

Along the way, I alphabetized the links in the example index page.  The
angular velocity page had a "Linear velocity..." title, so they looked
off in the sidebar, so I changed its title to be consistent.  I also
fixed a broken link that the doc build identified, in the main docs
page.

### DIFF
--- a/examples/adc_millivolts/au.cc
+++ b/examples/adc_millivolts/au.cc
@@ -41,13 +41,13 @@ using au::volts;
 // millivolt (mV) reference.  One count is one LSB: 3300 mV / 2^12.
 //
 // With Au, we can capture this as a custom unit.  Au even generates a readable label!
-using AdcCounts = decltype(Milli<Volts>{} * mag<3300>() / pow<12>(mag<2>()));
-constexpr auto adc_counts = QuantityMaker<AdcCounts>{};
+using AdcVolts = decltype(Milli<Volts>{} * mag<3300>() / pow<12>(mag<2>()));
+constexpr auto adc_volts = QuantityMaker<AdcVolts>{};
 
 
 
 int main() {
-    const auto v = adc_counts(2000);
+    const auto v = adc_volts(2000);
     // Exact rational conversion, applied once.  Without `ignore(...)`, this would not compile.
     std::cout << v.as(milli(volts), ignore(TRUNCATION_RISK)) << '\n';
 }
--- a/examples/adc_millivolts/au.cc
+++ b/examples/adc_millivolts/au.cc
@@ -1,0 +1,53 @@
+// Copyright 2026 Aurora Operations, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NOTE TO EDITORS: this file is line-aligned with `raw.cc`.  See the note in that file.
+
+// --8<-- [start:frontmatter]
+#include "au/au.hh"
+
+#include <iostream>
+
+#include "au/io.hh"
+#include "au/prefix.hh"
+#include "au/units/volts.hh"
+
+// This is a `.cc` file, so we import the names we use, one at a time.  See the "Namespaces and
+// includes" discussion page for why we do this rather than `using namespace au;`.
+using au::mag;
+using au::Milli;
+using au::milli;
+using au::QuantityMaker;
+using au::TRUNCATION_RISK;
+using au::Volts;
+using au::volts;
+// --8<-- [end:frontmatter]
+
+// clang-format off
+// --8<-- [start:example]
+// A 12-bit analog-to-digital converter (ADC) reports `counts` out of 4095,
+// spanning a 3300 millivolt (mV) reference.  That ratio is exact, so name it a unit.
+struct AdcCounts : decltype(Milli<Volts>{} * mag<3300>() / mag<4095>()) {
+    static constexpr const char label[] = "counts";
+};
+constexpr const char AdcCounts::label[];  // C++14 needs this; C++17 does not.
+constexpr auto adc_counts = QuantityMaker<AdcCounts>{};
+
+int main() {
+    const auto v = adc_counts(2048);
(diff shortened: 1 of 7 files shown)


examples/adc_millivolts/au.cc
```diff
@@ -0,0 +1,53 @@
...
+using au::TRUNCATION_RISK;
+using au::Volts;
+using au::volts;
+// --8<-- [end:frontmatter]
+
+// clang-format off
+// --8<-- [start:example]
+// A 12-bit analog-to-digital converter (ADC) reports `counts` out of 4095,
+// spanning a 3300 millivolt (mV) reference.  That ratio is exact, so name it a unit.
+struct AdcCounts : decltype(Milli<Volts>{} * mag<3300>() / mag<4095>()) {
```
hoffbrinkle: I think I misspoke when you asked me about this directly.

Is there a "cleaner" way to do this?  The ADC provides 12 bits of resolution.  `2^12` is `4096`.  Which means we can represent the values of `0` to `4095`.  The resolution of the unit is going to be `1 / 2^12` (`3.3 V / 2^12`).

```suggestion
// A 12-bit analog-to-digital converter (ADC) reports `counts` out of 4095,
// spanning a 3300 millivolt (mV) reference.  That ratio is exact, so name it a unit.
struct AdcCounts : decltype(Milli<Volts>{} * mag<3300>() / pow<12>(mag<2>())) {
```

I was conflating that with the interesting side effect of not actually being able to read the value of `3.3 V` from the device, as the saturated value will be `3.3 V - (3.3 V / 2^12)`.  I think that would be modeled by this really being a `uint12_t` that saturates rather than rolls over, which is likely beyond the scope of what Au should handle directly.

examples/adc_millivolts/au.cc
```diff
@@ -0,0 +1,53 @@
...
+    static constexpr const char label[] = "counts";
+};
+constexpr const char AdcCounts::label[];  // C++14 needs this; C++17 does not.
+constexpr auto adc_counts = QuantityMaker<AdcCounts>{};
+
+int main() {
+    const auto v = adc_counts(2048);
+    // Exact rational conversion, applied once.  Without `ignore(...)`, this would not compile.
+    std::cout << v.as(milli(volts), ignore(TRUNCATION_RISK)) << std::endl;
```
hoffbrinkle:
```suggestion
    std::cout << v.as(milli(volts), ignore(TRUNCATION_RISK)) << '\n';
```

examples/adc_millivolts/raw.cc
```diff
@@ -0,0 +1,39 @@
...
+int adc_to_millivolts(int counts) {
+    return counts * 3300 / 4095;  // Multiply first: 3300 / 4095 would truncate to 0.
+}
+
+
+
+int main() {
+    const int mv = adc_to_millivolts(2048);
+    // Careful: `mv / 1000` would quietly report 1650 mV as "1 V".
+    std::cout << mv << " mV" << std::endl;
```
hoffbrinkle:
```suggestion
    std::cout << mv << " mV" << '\n';
```

examples/adc_millivolts/au.cc
```diff
@@ -0,0 +1,53 @@
...
+using au::Volts;
+using au::volts;
+// --8<-- [end:frontmatter]
+
+// clang-format off
+// --8<-- [start:example]
+// A 12-bit analog-to-digital converter (ADC) reports `counts` out of 4095,
+// spanning a 3300 millivolt (mV) reference.  That ratio is exact, so name it a unit.
+struct AdcCounts : decltype(Milli<Volts>{} * mag<3300>() / mag<4095>()) {
+    static constexpr const char label[] = "counts";
```
hoffbrinkle: The `counts` part of this unit:  Historically I've tended to call this type of unit `AdcVolts`, treating it as scaling factor on the underlying base unit of `Volts`.  And then I would have a label of `adc volts`.  (I would substitute the specific part number for the `Adc`/`adc` portion.)

Internally, I've been using `count` when the `Uno` is the underlying base unit.

hoffbrinkle: On the label side of the discussion:  I've actually found myself going more toward the

```c++
using AdcCounts = decltype(Milli<Volts>{} * mag<3300>() / pow<12>(mag<2>()));
```

syntax and not providing a custom label at all.  If I see the following in some output:

```console
1 count
1 adc volt
```

I need to go either the source code (in the case of the former) or at least the corresponding data sheet (in the case of the latter) to know what that "means."

If we use the `using` syntax, I get:

```console
1 [(825 / 1024) mV]
```

Which, though maybe not as slick, allows me to quickly understand what that value is.  I suppose the context of what one is using the unit for in the output is probably a driving factor here.

If it's some debug output, the latter is usually more helpful.  If it's some sort of curated UI/display, the former is likely more applicable.

examples/adc_millivolts/au.cc
chiphogg: Thanks for explaining why 4096 is the way to go.  A really nice side effect is that the integer division becomes just a bit shift on the Au side!

examples/adc_millivolts/au.cc
chiphogg: Thanks for this.  I would have gone with "adc volts", except I liked your comment below even better, and decided to go with the anonymous scaled unit.

chiphogg: Nice!  I went with this approach and updated the text to match.

Zooming out and thinking more broadly about this, I guess we should probably prefer anonymous scaled units all the time, _unless_ the named unit has some "significance" to end users (i.e., they won't feel motivated to look at a spec sheet to see its definition).  That's different from the early days of the library, where named units were more important.

examples/adc_millivolts/au.cc
```diff
@@ -0,0 +1,55 @@
...
+// --8<-- [end:frontmatter]
+
+// clang-format off
+// --8<-- [start:example]
+// A 12-bit analog-to-digital converter (ADC) reports `counts` from 0 to 4095, spanning a 3300
+// millivolt (mV) reference.  One count is one LSB: 3300 mV / 2^12.
+//
+// With Au, we can capture this as a custom unit.  Au even generates a readable label!
+using AdcCounts = decltype(Milli<Volts>{} * mag<3300>() / pow<12>(mag<2>()));
+constexpr auto adc_counts = QuantityMaker<AdcCounts>{};
```
hoffbrinkle: FYI, my comment about count/volts also applies here.  I'd call this `AdcVolts` and `adc_volts`.  The reading is just a scaled voltage.  If one LSB was 1/1000th of a volt, we'd call it `millivolts`. 😁

hoffbrinkle (approved):

examples/adc_millivolts/au.cc
chiphogg: Ah, good point!  Done.

PR #736, opened 2026-08-22
Support streaming and unit symbols for Eigen

chiphogg:
I was working on an Eigen example, and found that some basic stuff
people would expect wasn't working yet.  So I added tests and fixed it.

On streaming: we've been using the "unary plus" for a long time in order
to promote small integers that alias to `char`.  The problem is that not
every rep necessarily supports unary plus.  Therefore, we guard it
behind a utility called as `promote_for_streaming(x, 0)`, and SFINAE out
the unary plus when it doesn't exist.

On unit symbols: apparently, we had been constraining them to arithmetic
types only.  Well, we have a canonical trait for what's a valid rep:
`IsValidRep`.  Even if it's not fully fleshed out yet, it's the
authoritative answer.  So now we just use that consistently.

### DIFF
--- a/au/io.hh
+++ b/au/io.hh
@@ -25,15 +25,23 @@
 
 namespace au {
 
+namespace detail {
+// Unary `+` promotes a char-like rep (e.g. `int8_t`), so that `<<` prints a number rather than a
+// character.  Not every rep has one --- an Eigen vector does not --- so promote only where we can.
+template <typename T>
+constexpr auto promote_for_streaming(const T &x, int) -> decltype(+x) {
+    return +x;
+}
+template <typename T>
+constexpr const T &promote_for_streaming(const T &x, ...) {
+    return x;
+}
+}  // namespace detail
+
 // Streaming output support for Quantity types.
 template <typename U, typename R>
 std::ostream &operator<<(std::ostream &out, const Quantity<U, R> &q) {
-    // In the case that the Rep is a type that resolves to 'char' (e.g. int8_t),
-    // the << operator will match the implementation that takes a character
-    // literal.  Using the unary + operator will trigger an integer promotion on
-    // the operand, which will then match an appropriate << operator that will
-    // output the integer representation.
-    out << +q.in(U{}) << " " << unit_label(U{});
+    out << detail::promote_for_streaming(q.in(U{}), 0) << " " << unit_label(U{});
     return out;
 }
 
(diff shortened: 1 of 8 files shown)


au/io.hh
```diff
@@ -25,15 +25,23 @@
...
 // Streaming output support for Quantity types.
 template <typename U, typename R>
 std::ostream &operator<<(std::ostream &out, const Quantity<U, R> &q) {
-    // In the case that the Rep is a type that resolves to 'char' (e.g. int8_t),
-    // the << operator will match the implementation that takes a character
-    // literal.  Using the unary + operator will trigger an integer promotion on
-    // the operand, which will then match an appropriate << operator that will
-    // output the integer representation.
-    out << +q.in(U{}) << " " << unit_label(U{});
+    out << detail::promote_for_streaming(q.in(U{}), 0) << " " << unit_label(U{});
```
hoffbrinkle, line 44: It's part of the `detail` API, so not critical, but the `0` for the second parameter is not-intuitive.

And really, the only type we're really worried about is the single byte integers which happen to be a `char` underneath, and `char` has this unfortunate ambiguity.  I suppose there's no good way to select on that?

au/quantity.hh
```diff
@@ -792,19 +792,11 @@ AU_DEVICE_FUNC constexpr auto rep_cast(Zero z) {
...
-// For now, we are restricting this to arithmetic types.  This doesn't mean they're the only reps we
-// support; it just means they're the only reps we can _construct via this method_.  Later on, we
-// would like to have a well-defined concept that defines what is and is not an acceptable rep for
-// our `Quantity`.  Once we have that, we can simply constrain on that concept.  For more on this
-// idea, see: https://github.com/aurora-opensource/au/issues/52
-struct NoTypeMember {};
-template <typename T>
-struct TypeIdentityIfLooksLikeValidRepImpl
-    : std::conditional_t<std::is_arithmetic<T>::value, stdx::type_identity<T>, NoTypeMember> {};
+// The identity on `T`, but only for a `T` we will accept as a rep, in `3.5 * m` and friends.
```
```suggestion
// The identity on `T`, but only for a `T` we will accept as a `Rep`.
```

I think we've got `Rep` documented well enough (externally as well) that we can just reference it directly.

hoffbrinkle (approved):

au/io.hh
chiphogg: Agreed on all points.  If you happen to find a better way to do this, I'm all ears!  😅  (It's an open invitation, so keep it in mind.)

PR #737, opened 2026-08-24
Add Eigen kinematics example

chiphogg:
This shows off a nice simplification of some Eigen code.  We get good
mileage out of automatic unit conversions, and even the gravity
constant.

Along the way, we tweak the naming and ordering in the index file, to be
more consistent with the titles that show up in the sidebar.

### DIFF
--- a/docs/examples/eigen-kinematics.md
+++ b/docs/examples/eigen-kinematics.md
@@ -1,0 +1,203 @@
+# Eigen: 3D vector kinematics
+
+A body has a position, a velocity, and an acceleration, each a 3D vector.  Where is it one timestep
+later, and how far from the origin is it?
+
+<!--
+  AUTHORING NOTE.  The two tabs below are a blink comparison: readers flip between them and compare
+  corresponding lines in place.  Keep each banner to one short line, or the code will start at
+  a different height in each tab and the comparison stops working.  The real explanation belongs
+  under "What's happening", not in the banner.
+-->
+
+=== "⚠️ Before: raw C++"
+
+    ⚠️ **Before** --- every vector is a bare `Vector3d`, and the units live only in the names.
+    { .ab-banner .ab-before }
+
+    ??? note "Includes and usings"
+
+        ```cpp
+        --8<-- "examples/eigen_kinematics/raw.cc:frontmatter"
+        ```
+
+    ```cpp
+    --8<-- "examples/eigen_kinematics/raw.cc:example"
+    ```
+
+=== "✅ After: with Au"
+
+    ✅ **After** --- each vector is a `Quantity` whose rep is `Vector3d`, so the units are checked.
+    { .ab-banner .ab-after }
+
+    ??? note "Includes and usings"
+
+        ```cpp
+        --8<-- "examples/eigen_kinematics/au.cc:frontmatter"
+        ```
+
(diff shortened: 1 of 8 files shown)


docs/examples/eigen-kinematics.md
```diff
@@ -0,0 +1,203 @@
...
+
+So, using concrete types for the input parameters and the return value automatically guarantees
+lifetime safety at those boundaries --- and an expression you compute and consume on the spot is
+safe, because nothing gets deferred past its inputs.
+
+## Summary
+
+Au's Eigen support makes it easier to get your units right _robustly_, and often makes your code
+easier to read.  Lifetime safety is the one thing it leaves exactly as it found it, for better and
+for worse, so make sure you're familiar with the Eigen [safety guide] before you start using Au with
```
hoffbrinkle, line 188: Just out of curiosity, I was curious what kind of documentation Eigen had for all this.  Mostly because I was thinking this could be simplified to "before you start using Eigen" (because Au doesn't add anything).

Having checked it out, we should leave it as you wrote it.  The [Eigen stuff](https://libeigen.gitlab.io/eigen/docs-nightly/TopicPitfalls.html) is a little more scary ("don't use `auto` unless you are 100% sure)...

examples/eigen_kinematics/au.cc
```diff
@@ -0,0 +1,91 @@
...
+// --8<-- [end:frontmatter]
+
+// clang-format off
+// --8<-- [start:example]
+// The types state the units.  Nothing to remember; nothing to convert.
+Position advanced_position(const Position &x,
+                           const Velocity &v,
+                           const Acceleration &a,
+                           QuantityD<Seconds> dt) {
+    return x + v * dt + 0.5 * a * dt * dt;
```
hoffbrinkle, line 73: Do you want to use `int_pow` here?

```suggestion
    return x + v * dt + 0.5 * a * int_pow<2>(dt);
```

hoffbrinkle (approved):

docs/examples/eigen-kinematics.md
chiphogg: Good find.  In some ways, I think our docs are better, because they're more clear about the _specific ingredients_ for a problem.  Not just "avoid `auto`" (which was my mindset when I set out to write the article), but "here are the two ingredients for a lifetime bug, and you need both".

In some ways, I think we were more incentivized than Eigen to find and articulate this clarity in our docs, because `auto` is such a core part of using Au.

examples/eigen_kinematics/au.cc
chiphogg: We could, but it seems like kind of a wash at best.

If we had more terms, we could build up the individual powers one at a time and store them in variables (`dt2`, `dt3`, etc.) to minimize recomputation.

docs/examples/eigen-kinematics.md
hoffbrinkle: Yeah, I don't know what the state of `auto` is in the broader industry at large.  If I found "our" (good job!) docs on the safety pretty clearly written.  It also helps understand template expressions pretty well.  Almost seems like Eigen itself could stand to update at least two different parts of their docs into a similar type of doc.

PR #738, opened 2026-08-25
Add doc for abbreviated quantity construction

chiphogg:
Our new unit literals are one of the most exciting features of Au 0.6.0,
but most users aren't well equipped to appreciate it, or use it
effectively.  I felt we needed a discussion doc to tackle this head-on.
Hence, the new "abbreviated quantity construction" doc.  It explains the
evolution of these features and the pros/cons of the two approaches we
ended up with.

Along the way, I updated related docs to make sure they reflect our best
practices guidance and the current state of the code.

I expect this to be the final "normal" commit for 0.6.0; after this, we
should just be doing mechanical commits from the release process.

### DIFF
--- a/docs/discussion/idioms/abbreviated-quantity-construction.md
+++ b/docs/discussion/idioms/abbreviated-quantity-construction.md
@@ -1,0 +1,414 @@
+# Abbreviated Quantity Construction
+
+Ever since C++11 brought [user-defined literals] (UDLs) to the language, most C++ units libraries
+have provided them out of the box for a wide variety of units.  It's a very popular feature ---
+after all, who _wouldn't_ rather write `65_mph` than `miles_per_hour(65)`?  Checking off the
+"user-defined literals" box makes a library seem more complete, more appealing.
+
+We'd like to reframe that box in terms of the need it's actually meeting.  Instead of "user-defined
+literals", think of it as "abbreviated quantity construction".  Users want concise, readable ways to
+make quantity objects.  UDLs are one way to do that, but not the only way.  In fact, for quantities
+in particular, UDLs have serious flaws that competing approaches don't suffer from.
+
+This article will explain how UDLs generally work in a quantity context, and the downsides that
+turned Au's _reluctance_ into outright _rejection_.  We'll explain an alternative approach, _unit
+symbols_, that solves virtually all of UDLs' problems.  Finally, we'll share the surprising
+discovery that put UDLs back on the table: by giving them _non-quantity_ return types, they turned
+out to handle the _quantity_ use cases better than any other library on the market!  We'll close
+with guidance on which tool to reach for, when.
+
+## UDLs for quantities
+
+The basic definition of a UDL is simple.  Although the _spelling_ of the function name is somewhat
+arcane (for example, the `_m` literal is spelled `operator""_m`), it is, ultimately, just a
+function.  As the [docs][user-defined literals] explain, the set of permissible parameter types is
+heavily restricted. For quantity literals, the most relevant options are `long double` for floating
+point literals, and `unsigned long long` for integer literals.
+
+Let's define one now.  We'll write what we'd get if Au were to follow the common style in other
+libraries, using `meters` as an example:
+
+```cpp
+constexpr auto operator""_m(long double v) {
+    return meters(static_cast<double>(v));
+}
+```
+
+With this, we can write `3.5_m`, and we'll get a `QuantityD<Meters>` object with a value of `3.5`.
+This is a huge benefit for migrating callsites, especially when functions have many parameters.
+Having many literals in a callsite is common in unit test files.  You can migrate those tests by
+just adding a little unit suffix on each number.  This kind of change will be a lot easier to review
+at a glance, and much less likely to overflow the line width than spelling out all the unit names in
+full.
+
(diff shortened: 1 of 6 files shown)


docs/discussion/idioms/abbreviated-quantity-construction.md
```diff
@@ -0,0 +1,414 @@
+# Abbreviated Quantity Construction
+
+Ever since C++11 brought [user-defined literals] (UDLs) to the language, most C++ units libraries
+have provided them out of the box for a wide variety of units.  It's a very popular feature ---
+after all, who _wouldn't_ rather write `65_mph` than `miles_per_hour(65)`?  Checking off the
+"user-defined literals" box makes a library seem more complete, more appealing.
+
+We'd like to reframe that box in terms of the need it's actually meeting.  Instead of "user-defined
+literals", think of it as "abbreviated quantity construction".  Users want concise, readable ways to
```
hoffbrinkle: :pick:

```suggestion
literals", think of it as "abbreviated quantity construction for literals".  Users want concise, readable ways to
```

❓

docs/discussion/idioms/abbreviated-quantity-construction.md
```diff
@@ -0,0 +1,414 @@
...
+
+These oddities turn out to be just the tip of the iceberg.  Mateusz Pusz (lead author of the
+[mp-units] library) was the first person we know of to make a _systematic case_ against using UDLs
+for quantities.  Some of the key [problems] he identifies include:
+
+- **UDLs compose poorly.**  If you have `_m` and `_s`, you can't make one for `_mps` (meters per
+  second) without writing an entirely new overload.  Actually, make that _two_ new overloads,
+  because...
+
+- **UDLs are expensive.**  We need two copies for every unit, so the _maintenance_ cost is high.
```
hoffbrinkle: I think the `because...` was meant to be used when these were in a different order (i.e. what is the third item in the list used to be the second)?

docs/discussion/idioms/abbreviated-quantity-construction.md
```diff
@@ -0,0 +1,414 @@
...
+
+Unit symbols are [_monovalue types_][monovalue types].  They're empty, so the "multiplication" or "division" _never
+has any runtime cost_.  Instead, it just tells the compiler how to change the units of some _other_
+variable that _does_ have a runtime value.
+
+Let's see how all of our UDL problems melt away with unit symbols:
+
+- **Unit symbols compose naturally.**  We've already seen how `m / s` is a symbol for meters per
+  second, composed on the fly.  We can also apply prefixes inline: `kilo(m)` does just what it looks
+  like, and we can assign it to a local variable called `km` for even better readability.
```
hoffbrinkle: > and we can assign it to a local variable called `km` for even better readability.

But please don't... 😉  (I.e. encoding the unit into the variable name.)

docs/discussion/idioms/abbreviated-quantity-construction.md
```diff
@@ -0,0 +1,414 @@
...
+
+- **Unit symbols work with any rep.**  It's perfectly natural.  `3.5 * m` gives us a rep of
+  `double`, and `3.5f * m` gives us a rep of `float`.  In fact, `Eigen::Vector3d{1.0, 2.0, 3.0} * m`
+  even gives us a rep of `Eigen::Vector3d`!
+
+- **Unit symbols support variables.**  You can write `my_legacy_raw_numeric_variable_m * m`, and see
+  at a glance (`..._m * m`) that we've gotten the units right.
+
+It's not as if UDLs have _no_ advantages.  `3.5_mps` has a more concise, unified appearance than
+`3.5 * m / s`, or even than the closer analogue `3.5 * mps` (if we wanted to define an ad hoc `mps`
```
hoffbrinkle: :pick: is the "even than" a bit of a mouthful here?

```suggestion
`3.5 * m / s`, or the closer analogue `3.5 * mps` (if we wanted to define an ad hoc `mps`
```

hoffbrinkle (approved):

docs/discussion/idioms/abbreviated-quantity-construction.md
chiphogg: No, the order was correct.  I just wanted `because...` to hit "We need two copies...".  But it just... _doesn't_, so thanks for pointing that out.

I reworded this and I'm a lot happier now.

docs/discussion/idioms/abbreviated-quantity-construction.md
chiphogg: Sorry, I think you misunderstood me.  A local variable called `km` was meant to be _literally the whole name_: a `constexpr auto km = kilo(m);` type deal.

But this means this prose was hard to understand, so I massaged it to be clearer, and added a footnote to expand even further.  The footnote links to our Eigen example where we do exactly this.

docs/discussion/idioms/abbreviated-quantity-construction.md
chiphogg: I took the text out, and read it, and found myself wanting to put it back in to say what I wanted to say.  But you're right that it's awkward!  So I reworded for clarity and I'm happier with it now.

Issue #715, opened 2026-08-10
Make ScalarOfTrait required


chiphogg:
This needs to be done after 0.6.0, for the 0.7.0 release.

PR #740, opened 2026-08-26
DO NOT LAND: Require `ScalarOf<T>` for any rep `T`

chiphogg:
Helps #715.  This commit is the "preview" version for the future-proof release `0.6.0-future-715`.

This PR is for running all CI, and for Au team members to do a post hoc review when they're back from vacation.

### DIFF
--- a/au/rep.hh
+++ b/au/rep.hh
@@ -246,6 +246,12 @@ using ScalarOfOrVoid = stdx::experimental::detected_or_t<void, ::au::ScalarOf, T
 template <typename T>
 struct HasQuantityLikeScalar : LooksLikeAuOrOtherQuantity<ScalarOfOrVoid<T>> {};
 
+// Every rep must have a `ScalarOf`: it's what the overflow and truncation risk checks are built on.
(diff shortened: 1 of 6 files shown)


au/rep.hh
```diff
@@ -246,13 +246,20 @@ using ScalarOfOrVoid = stdx::experimental::detected_or_t<void, ::au::ScalarOf, T
 template <typename T>
 struct HasQuantityLikeScalar : LooksLikeAuOrOtherQuantity<ScalarOfOrVoid<T>> {};
 
+// Every rep must have a `ScalarOf`: it's what the overflow and truncation risk checks are built on.
+// Most types are covered by the automatic detection, but a type that matches no probe needs a
```
hoffbrinkle, line 250: "matches no" or "does not match any"?

hoffbrinkle (approved):

au/rep.hh
chiphogg: Yep, the latter is a less ambiguous way of saying what I was trying to say.

PR #743, opened 2026-08-28
Add support for C++20 modules

mikomikotaishi:
This PR adds support for C++20 modules, enabled with `AU_BUILD_MODULE` on CMake (requires a build system like Ninja, which has module support) on at least CMake 3.28. Creates a module `au`.

### DIFF
--- a/.github/license-check/license-config.json
+++ b/.github/license-check/license-config.json
@@ -10,6 +10,7 @@
       "tools/bin/au-docs-serve",
       "tools/bin/buildifier",
       "tools/bin/check-cmake-headers",
+      "tools/bin/check-module-exports",
       "tools/bin/clang-format",
       "tools/bin/make-single-file",
       "tools/bin/mkdocs",
@@ -39,6 +40,7 @@
   {
     "include": [
       "**/*.cc",
+      "**/*.cppm",
       "**/*.cu",
       "**/*.hh",
       "docs/assets/hrh4.js"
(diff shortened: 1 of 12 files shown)


chiphogg:
Thanks for this contribution!  I haven't reviewed the code yet, but strategically, here's what we'd need in order to get this reviewed and landed.

1. We need some way to test this in CI.  None of the maintainers actually use modules, so we don't know how to do that.
2. We need some way to test that this doesn't go stale.  Right now, we're reproducing a lot of identifiers in the `.cppm` file.  The CMake/bazel correspondence has some CI tests to make sure they don't accidentally drift.  Can you think of something like that for the modules/include correspondence?

I'd imagine we need to see (1) actually implemented in this PR.  For (2), I guess that could be a follow-on, but we'd need to at least sketch out a plausible design so we could be confident we'd avoid being in a situation where we land this, but it drifts.

Thanks!

chiphogg (commented):
OK, this looks ready for me to do a first pass review.

Full disclosure --- I'm pretty swamped right now, trying to catch up on some things that got away from me.  So this probably won't be a _super_ quick review, but it is on my radar, and the ball's in my court.  Just wanted to set expectations up front. 🙂

mikomikotaishi:
That's fine; in the meantime I'll look at what else I could do to improve this or ensure it's maintainable for future maintainers.

chiphogg:
For transparency's sake: I haven't had time to work on this review yet.  I had a family emergency over the Labor Day weekend, and now that I'm back at work, we're in a crunch time.  That said --- rest assured that this is very much on my radar, and I'll get to it as soon as I can!  I'm hoping/expecting I can finish the review within two weeks, so feel free to check in if Sep 25 comes and goes without an update.

Issue #748, opened 2026-09-01
Move mag literal call outside of `make_constant` in literal implementations


chiphogg:
This is a pure flexibility upgrade.  If the magnitude literal happens to produce `Zero` (#747), then this will let our unit literal return `Zero`, instead of producing a compiler error.  A `Constant` can be multiplied by `Zero`, but the quantity maker inside of the `make_constant` call cannot.  We could add it, but multiplying a `QuantityMaker` by `Zero` would need to produce `Zero` conceptually, which would _fundamentally change the interface_ (quantity makers are callable, but `Zero` is not and should remain uncallable).

chiphogg:
I guess merging #751 didn't close this because it didn't merge to the `main` branch.

PR #750, opened 2026-09-04
Pull mag literal outside in unit literals

chiphogg:
This makes the implementations more flexible.  It means they will still
work if `operator""_mag` returns `Zero`, which it soon will be able to.

Helps #748.  (We still need to cherry pick to the release branch.)

### DIFF
--- a/au/units/literals/amperes.hh
+++ b/au/units/literals/amperes.hh
@@ -21,7 +21,7 @@
 namespace au {
 namespace au_literals {
 
-// `1.28e-4_A` is a `Constant` equivalent to `make_constant(1.28e-4_mag * amperes)`.
+// `1.28e-4_A` is a `Constant` equivalent to `make_constant(amperes) * 1.28e-4_mag`.
 template <char... Cs>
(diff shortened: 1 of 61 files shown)


au/units/literals/amperes.hh
geoffviola: Does this comment need to be updated?

geoffviola (approved):
Do we need to update the [new-units](https://github.com/aurora-opensource/au/blob/f389e44f743d12b92ca505e42395ba1a454b44f2/docs/howto/new-units.md?plain=1#L46) doc?

au/units/literals/amperes.hh
chiphogg: Great catch!

chiphogg:
> Do we need to update the [new-units](https://github.com/aurora-opensource/au/blob/f389e44f743d12b92ca505e42395ba1a454b44f2/docs/howto/new-units.md?plain=1#L46) doc?

Yes, and the Constant doc too.  Great catch(es)!

Issue #752, opened 2026-09-04
Fix branching instructions for future proof releases and patch releases


chiphogg:
Right now, future proof releases are PRs against the release branch.  This means that when we cherry pick onto the release branch for a patch release, it includes the future proof branches.  We need to update our instructions so this isn't so goofy.  Meanwhile, I am going to put up a reverting PR against the release branch for the future proofing releases.

chiphogg:
Actually, this was just me being dumb: I forgot to create an 0.6.1 release branch.

PR #753, opened 2026-09-04
Revert future proof commits from patch release

chiphogg:
Our release instructions have bad interactions between future proof
releases, and patch releases.  The future proof commits are on the
release branch, so right now, they will get included in the patch
release (undesirable!) unless we do something.

I filed #752 to fix this.  For now, I'm going to just revert the 3
future proof commits on the release branch so that its contents
correspond to the patch release.


hoffbrinkle (approved):

chiphogg:
This was me being dumb, actually.  What we were supposed to do was create a `release-0.6.1` branch.  I've done that now, and I cherry picked the PR for #748 before I pushed, so it's in the right state.

What this really shows is a gap in our release instructions around how we handle patch branches.  I will put up a PR.

PR #758, opened 2026-09-12
Update python requirements

chiphogg:
This time around, I leaned on Claude for reading the release notes and
checking for changes.  It found that all differences were due to three
sources: the `generator` meta tag, content-hashed CSS and JS filenames,
and a change in pygments to output literal quote characters (both `'`
and `"`) instead of HTML-encoded ones.  After normalizing for these, it
found that 68 of 68 pages were byte identical.

There was also one CSS diff, which changed `white-space:nowrap` to
`white-space:pre` in the search bar.  I tried a query in both and it
looked the same.  I also looked at the atomic units example to do
another check on the pygments change in a code-heavy file, and it seemed
fine.

Finally, the `mike` changes simplify our monkey patch.  We need to get
rid of one of the two monkey patches, because it no longer applies.
After simplifying, `bazel run //:mike -- list` now works again.

Assuming all of our dependabot alerts can be simply fixed by updating to
the latest version, this will fix dependabot alerts 26, 27, 28, 29, 30,
31, 32, 33, and 34.


requirements_lock.txt
```diff
@@ -273,9 +323,9 @@ mergedeep==1.3.4 \
     # via
     #   mkdocs
     #   mkdocs-get-deps
-mike==2.1.3 \
-    --hash=sha256:abd79b8ea483fb0275b7972825d3082e5ae67a41820f8d8a0dc7a3f49944e810 \
-    --hash=sha256:d90c64077e84f06272437b464735130d380703a76a5738b152932884c60c062a
+mike==2.2.0 \
```
geoffviola, line 326: Claude outlined that mike might have issues. For example, the following command will fail
```shell
bazel run //:mike -- deploy --push main
```

You might want to consider deleting the override. Here's a Claude suggested diff
```diff
---
diff --git a/mike_bin.py b/mike_bin.py
index d8d71aa9..11774061 100644
--- a/mike_bin.py
+++ b/mike_bin.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import os
-import re
 import subprocess
 import sys
 
@@ -24,28 +23,9 @@ from mike.mkdocs_utils import docs_version_var
 
 
 # We need to monkey patch any function in mike which calls `mkdocs` via its
-# command line interface.  Fortunately, there are only two: `version` and
-# `build`.
-
-
-def override_version():
-    """Override mike's native logic for retrieving the version."""
-    # Original code:
-    #    output = subprocess.run(
-    #        ['mkdocs', '--version'],
-    #        check=True, stdout=subprocess.PIPE, universal_newlines=True
-    #    ).stdout.rstrip()
-    #    m = re.search('^mkdocs, version (\\S*)', output)
-    #    return m.group(1)
-    # Changed version (reformatted with Black):
-    output = subprocess.run(
-        ["update_docs", "--version"],
-        check=True,
-        stdout=subprocess.PIPE,
-        universal_newlines=True,
-    ).stdout.rstrip()
-    m = re.search("^\\S+, version (\\S*)", output)
-    return m.group(1)
+# command line interface.  As of mike 2.2.0, there is only one: `build`.  (Mike
+# used to shell out to `mkdocs --version` as well, but it now reads the version
+# from `importlib.metadata` at import time, so there is nothing to patch.)
 
 
 def override_build(config_file, version, *, quiet=False, output=None):
@@ -74,6 +54,5 @@ if __name__ == "__main__":
     # We need to add this folder to the PATH.
     os.environ["PATH"] = f"{os.path.join(os.getcwd())}:{os.environ['PATH']}"
 
-    with mock.patch("mike.mkdocs_utils.version", override_version):
-        with mock.patch("mike.mkdocs_utils.build", override_build):
-            sys.exit(mike.driver.main())
+    with mock.patch("mike.mkdocs_utils.build", override_build):
+        sys.exit(mike.driver.main())

```

geoffviola (approved):
LGTM. Just one potential improvement.

requirements_lock.txt
chiphogg: Great catch!  We would have had to discover that the next time we ran `mike` otherwise.  I've included the patch simplification in this PR now.

geoffviola (approved):

Issue #744, opened 2026-08-28
Support `Constant` in `fmod`, `remainder`, `copysign`


chiphogg:
The most likely pathway to this is to make them hidden friends of `Quantity`.

chiphogg:
This lets us do something like the following:

```cpp
constexpr auto FULL_TURN = make_constant(revolution);
const auto wrapped = fmod(angle, FULL_TURN);
```

Then `FULL_TURN` automatically adapts to the units and rep of `angle`, which is awesome!

chiphogg:
Cherry-pick to 0.6.1 branch complete.

PR #767, opened 2026-09-23
Support shapeshifter types in more math functions

chiphogg:
The motivating use case was something like:

```cpp
constexpr auto FULL_TURN = make_constant(revolution);
const auto wrapped = fmod(angle, FULL_TURN);
```

This has great ergonomics: it automatically adapts to `angle`'s type!
But we needed to add shapeshifter support.

The best way to do this was with hidden friends.  There's precedent, as
we do this with `min` and `max`.  This does feel like a little bit of an
expansion, but I think these functions are OK here.  They're still just
`<cmath>` functions, and they're pretty generically useful for
`Quantity`.

Also, note that none of these make sense for `QuantityPoint`, so no need
to add them there.

PR #768, opened 2026-09-23
Support shapeshifter types in more math functions (#767)

chiphogg:
This is the cherry pick of #767 to the 0.6.1 release branch.


chiphogg:
This will fix #744.

docs/reference/math.md
```diff
@@ -52,20 +52,29 @@ Adapts [`std::copysign`](https://en.cppreference.com/w/cpp/numeric/math/copysign
 
 ```cpp
 // 1: First argument Quantity, second argument raw numeric
+// (`T` must be a valid rep)
```
geoffviola, line 55: Is this implicit to having a compiled program? I think the paragraph below explains things well. I know there is a specific `IsValidRep` definition, but colloquially it feels redundant to saying that it compiles.

docs/reference/math.md
```diff
@@ -52,20 +52,29 @@ Adapts [`std::copysign`](https://en.cppreference.com/w/cpp/numeric/math/copysign
...
 // 2: First argument raw numeric, second argument Quantity
+// (`T` must be a valid rep)
 template <typename T, typename U, typename R>
 constexpr auto copysign(T mag, Quantity<U, R> sgn);
 
 // 3: Both arguments Quantity
 template <typename U1, typename R1, typename U2, typename R2>
 constexpr auto copysign(Quantity<U1, R1> mag, Quantity<U2, R2> sgn);
+
+// 4: Both arguments the same Quantity type (hidden friend)
```
geoffviola, line 68: Nit: The hidden friend detail seems overly verbose

au/quantity.hh
```diff
@@ -602,6 +603,41 @@ class Quantity {
...
+    // are handled by the function templates in "au/math.hh".
+    //
+    // Each is a template on a defaulted parameter `T`, which is always `Rep`.  This keeps the
+    // return type out of the class instantiation, so that reps which have no `std::fmod` (say,
+    // `std::complex<double>`) simply don't get these overloads, instead of failing to compile.
+    // The _parameters_ stay concrete, which is the whole point.  Note that the return types
+    // promote the rep exactly as the underlying `std` functions do, which keeps these consistent
+    // with their "au/math.hh" counterparts.
+    template <typename T = Rep>
+    friend AU_DEVICE_FUNC auto fmod(Quantity a, Quantity b)
```
geoffviola, line 620: Did we want to make `fmod` and `remainder` `constexpr` if `std::fmod` and `std::remainder` are? That might help C++23 users where  `std::fmod` and `std::remainder` are `constexpr`.

geoffviola (approved):
Changes, tests, and documentation look good. I left some nitpicks and questions.

au/quantity.hh
chiphogg: This seems desirable, but not yet.  I think we'd want to wait until we get a C++23 test, and I don't think any of our current toolchains support C++23.

docs/reference/math.md
chiphogg: I think this is more about documenting to the user what kinds of things they should _try_ to pass.  In other words, this is the _contract_, and whether or not it compiles is down to the _implementation_.

That said, I'm making a follow-on PR to update some comments anyway.  I could easily be off base on this, so please feel free to bring it up in that PR's review if my explanation here doesn't satisfy you! 🙂

docs/reference/math.md
chiphogg: Yeah, makes sense.  I'll fix it in #769.

